PR #12, opened 2018-04-23
Sharding: copy primary key tables only at the end of the ferry

hkdsun:
This change helps keep the primary key table in the target database consistent for a longer time period.

For example, for us, we have developers who `JOIN ON` a primary key table in their queries to filter out some data but see unexpected results. This change makes that sort of inconvenience less likely

follow up to: https://github.com/Shopify/shopify/pull/153139
related to: https://github.com/Shopify/pods/issues/259
@Shopify/pods

### DIFF
--- a/sharding/sharding.go
+++ b/sharding/sharding.go
@@ -24,16 +24,10 @@ func NewFerry(config *Config) (*ShardingFerry, error) {
 
 	config.DatabaseRewrites = map[string]string{config.SourceDB: config.TargetDB}
 
-	primaryKeyTables := map[string]struct{}{}
-	for _, name := range config.PrimaryKeyTables {
-		primaryKeyTables[name] = struct{}{}
-	}
-
 	config.CopyFilter = &ShardedCopyFilter{
-		ShardingKey:      config.ShardingKey,
-		ShardingValue:    config.ShardingValue,
-		JoinedTables:     config.JoinedTables,
-		PrimaryKeyTables: primaryKeyTables,
+		ShardingKey:   config.ShardingKey,
+		ShardingValue: config.ShardingValue,
+		JoinedTables:  config.JoinedTables,
 	}
 
 	ignored, err := compileRegexps(config.IgnoredTables)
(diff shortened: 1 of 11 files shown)


hkdsun:
Added table verification logic which required further refactor of the iterative verifier.
 
Found a bug in the `Pool` helper on the way where we were not consuming the whole error channel as well.

@fw42 could you have another look?

utils.go
```diff
@@ -115,12 +115,27 @@ loop:
...
 		case workQueue <- i:
 			i++
-		case err = <-errCh:
-			break loop
+		case err = <-errCh: // abort pool if an error was discovered
+			if err != nil {
+				break loop
+			}
+		default:
+			time.Sleep(500 * time.Millisecond)
```
hkdsun: We were busy-looping previously if all goroutines were busy doing work

sharding/sharding.go
```diff
@@ -212,6 +216,59 @@ func (r *ShardingFerry) deltaCopyJoinedTables() error {
...
+		if _, exists := pkTables[table.Name]; exists {
+			if len(table.PKColumns) != 1 {
+				return fmt.Errorf("Multiple PK columns are not supported with the PrimaryKeyTables option")
+			}
+			tables = append(tables, table)
+		}
+	}
+
+	if len(tables) == 0 {
+		r.logger.Warn("found no primary key tables to copy")
```
fw42: I assume this is for cases where the "shops table isn't sharded yet"?

hkdsun: yeah or the PrimaryKeyTables config is not provided. maybe we should return an error here

utils.go
```diff
@@ -115,12 +115,27 @@ loop:
...
 
 	close(workQueue)
 	wg.Wait()
+	close(errCh)
+
+	if err != nil {
+		return results, err
+	}
+
+	for e := range errCh {
```
fw42: If there's multiple errors in the channel, this will make it so that we return (only) the last one. That's intentional?

hkdsun: yeah. unless we return an array of errors which is unconventional. what would you expect?

sharding/sharding.go
fw42: If we return an error then people who's tenant table isn't sharded can't use this, right? I think we should allow them to.

fw42: Or do you mean return an error if a table name is specified but the table isn't found? If so then I agree.

hkdsun: yup latter is what i meant. will push that change

utils.go
fw42: I'd probably expect to see the first instead of the last error, but no big deal.

hkdsun:
Shuhao and Justin are both away for a few days so I'm going to merge this. Feel free to leave comments on this PR and I'll come back to them

PR #15, opened 2018-05-01
Debian compliant debian packaging

shuhaowu:
This PR adds Debian compliant Debian package builds for the binaries of Ghostferry such that we can upload this to a PPA and even possibly get it upstreamed into Ubuntu/Debian. 

This currently only builds for bionic (possibly artful) as xenial doesn't have required Go toolchain from its repository. On Launchpad, it is possible to use the Copy Packages functionality to copy the built binaries from bionic to xenial (and possibly trusty). It does mean that if we build from circle, we need to build it in a bionic environment (which means we're locked using go 1.10 i believe, for the time being).

Not sure if we should merge this or not as we already have some sort of debian building and can just upload that deb file into Github releases instead of doing this dance instead.

A test PPA is setup here: https://launchpad.net/~shuhao/+archive/ubuntu/ghostferry-unofficial

@pushrax I'm curious to see what you think about this approach, and whether or not we should use this approach going forward as we are now OSS.


pushrax:
I think this approach makes sense, I also set it up in our nginx build. A proper changelog available to apt is nice. However, being locked to bionic is a bit annoying, and for now GH releases seem pragmatic until there's some interest in upstreaming.

PR #21, opened 2018-05-16
Migrated from Circle to Travis

shuhaowu:
As Circle 1.0 is shutting down, we're moving to Travis instead.

Some features implemented here:

- Run tests for each commit.
- Automatically build documentations for each `master` commit.
- Automatically build documentations AND deb releases for each tag.


shuhaowu:
Oh yeah I forgot to mention that: I don't have access to Packagecloud and I was wondering if someone who does can help me get that to work after I merge this branch.

pushrax:
For packagecloud we'll need to set up travis secrets, not sure what precedent we have set already for that. Should be OK if restricted to the ghostferry packagecloud repo.

PR #27, opened 2018-05-23
Step 1: Cursor to use PK struct instead of uint64 to iterate tables

shuhaowu:
This is step one of the work to migrate data with generic sortable columns. This implementation switch the Cursor's implementation to use a PK struct instead of uint64 to iterate over the data. The PK struct is currently only supports uint64 although it should allow arbitrary sortable types in the future.

*I would like to have some feedback in this approach before moving forward with the full implementation, detailed below:*

In step two (near future), I would like to convert the remaining code base to utilize the PK struct. This includes the `BuildSelect` method that we use to filter for arbitrary rows and some other locations.

In step three (near future), I would like to add string based primary key support. **At this point, we can migrate tables like `schema_migrations` without having to resort to mysqldump or other out of band methods during the cutover stage.**

After this, we can potentially support non primary key (unique/sortable columns), other data types, multiple column keys, and so on. However, I currently have no plans to implement the support for those myself.

### DIFF
--- a/pk.go
+++ b/pk.go
@@ -1,0 +1,49 @@
+package ghostferry
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/siddontang/go-mysql/schema"
+)
+
+type PK struct {
+	Value interface{}
+}
+
+func (pk PK) Compare(other PK) int {
+	switch v1 := pk.Value.(type) {
+	case uint64:
+		v2 := other.Value.(uint64)
+
+		if v1 < v2 {
+			return -1
+		} else if v1 == v2 {
+			return 0
+		} else {
+			return 1
+		}
+
+	default:
+		panic(fmt.Sprintf("pk type %T are current not supported", v1))
+	}
+}
+
+// Could possibly upstream these functions in the future.
+func MinPK(column *schema.TableColumn) PK {
+	switch column.Type {
+	case schema.TYPE_NUMBER:
+		return PK{Value: uint64(0)}
+	default:
+		panic(fmt.Sprintf("non integer type is currently not supported"))
+	}
+}
+
+func MaxPK(column *schema.TableColumn) PK {
+	switch column.Type {
+	case schema.TYPE_NUMBER:
+		return PK{Value: uint64(math.MaxUint64)}
+	default:
+		panic(fmt.Sprintf("non integer type is currently not supported"))
+	}
+}
--- a/cursor.go
+++ b/cursor.go
@@ -40,7 +40,7 @@ type CursorConfig struct {
 }
 
 // returns a new Cursor with an embedded copy of itself
-func (c *CursorConfig) NewCursor(table *schema.Table, maxPk uint64) *Cursor {
+func (c *CursorConfig) NewCursor(table *schema.Table, maxPk PK) *Cursor {
 	return &Cursor{
 		CursorConfig:  *c,
 		Table:         table,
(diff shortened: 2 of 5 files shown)


pk.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+	default:
+		panic(fmt.Sprintf("pk type %T are current not supported", v1))
+	}
+}
+
+func MinPK(t int) PK {
+	switch t {
+	case schema.TYPE_NUMBER:
```
hkdsun: If this function is going to depend on siddontang's `schema` then it makes more sense to me that the input argument to it be the `TableColumn` struct itself

shuhaowu:
Another alternative would be something similar to gh-ost: I believe the idea is that it will try to fetch the next range of keys to iterate through from the database and not do a comparison within the Go code base. If this next range calculation is delegated to MySQL, we essentially skip the issue of having to serializing it in Go. This _should_ (?) allow us to implement arbitrary types more easily (or even for free?), including compound primary keys.

However, it looks like it'll take more time to implement than I have right now as it would require more dramatic changes in the code base such that these incremental improvements are not possible.

The primary objective for me right now is to support string primary key. This would give us an immediate benefit as we no longer need an out of band method to copy `schema_migrations`.

shuhaowu:
The IterativeVerifier is the other one as it uses pk batches. Maybe it's not as bad as I thought, I could potentially take a look again next week.

kolbitsch-lastline:
did we ever manage to get this into master - I see the conversation just died on minor requests on the review, but the code seems useful to help towards composite primary keys

fjordan:
Support for composite primary keys, and subsequently using a configurable `paginationKey`, was added in https://github.com/Shopify/ghostferry/pull/138

kolbitsch-lastline:
> Support for composite primary keys, and subsequently using a configurable paginationKey, was added in #138

got it, thanks. I was confused, because part of that commit was actually reverted already (I couldn't find some of the helper methods introduced in that merge).

What I'm mostly looking into is supporting **composite** primary keys (even if integers) - there was a bit of misleading conversation on #138 . Probably it's well known, but we cannot support that with the current code, as the batching/pagination may lead to cases where we skip copying rows. For example:

table columns: `id1`, `id2`, `data`
PK: (`id1`, `id2`)
batch size: 3
data:

```
    id1 id2   data
row1: 1, 1, "data1.1"
row2: 2, 1, "data2.1"
row3: 2, 2, "data2.2"
row4: 2, 3, "data2.3"
row5: 3, 1, "data3.1"
```

here:
- the first batch would be rows `1`, `2`, and `3`.
- resuming would happen with PK `id1 > 2`, because the last PK value we copied is `2`
- this would lead to skipping the copy of `row4`, as `row5` is the first row that meets the resume criteria

again, you probably knew this already, just wanted to post here in case, and **please correct me if I'm mistaken**.

IMO there is a very simple approach of skipping copies of "incomplete PK batches" (shorten the batch if we end up crossing a PK ID boundary), but that can lead to empty batches / errors if there are more rows than the batch size. Also, it's ugly and inefficient

Another, much better, way would be to truly support multi-column keys, where then the resume criteria becomes very easy (it's a `>=` for all PK columns except the least significant one, where it remains `>`). Thus, `len(PK) == 1` just becomes corner case of this resume condition.

Given your answer, the above has not been implemented in any unmerged branch, correct? We don't need it in our current use-cases, but I can see it become important in the future, and I'm considering adding support (since I'm knee-deep in `ghostferry` already anyways :-) ).

shuhaowu:
You're correct. The current assumption of Ghostferry's pagination key support is that the pagination key column is an unique, integer column.

PR #29, opened 2018-05-28
Refactor sql.DB creation to DatabaseConfig

shuhaowu:
Previously, the way we create `*sql.DB` are scattered all over the place and they do the same thing. It make sense to refactor this into `DatabaseConfig`. This means I won't have to write it again in order to wait for the replica to catch up to the master.

This is a precursor to the PR for waiting on replica to catch up to the master writer node.

### DIFF
--- a/config.go
+++ b/config.go
@@ -3,11 +3,13 @@ package ghostferry
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"database/sql"
 	"errors"
 	"fmt"
 	"io/ioutil"
 
 	"github.com/go-sql-driver/mysql"
+	"github.com/sirupsen/logrus"
 )
 
 type TLSConfig struct {
@@ -49,7 +51,7 @@ type DatabaseConfig struct {
 	TLS *TLSConfig
 }
 
-func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
+func (c DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
 	cfg := &mysql.Config{
 		User:      c.User,
 		Passwd:    c.Pass,
@@ -80,7 +82,7 @@ func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
 	return cfg, nil
 }
 
-func (c *DatabaseConfig) Validate() error {
+func (c DatabaseConfig) Validate() error {
 	if c.Host == "" {
 		return fmt.Errorf("host is empty")
 	}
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -49,7 +51,7 @@ type DatabaseConfig struct {
 	TLS *TLSConfig
 }
 
-func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
+func (c DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
```
shuhaowu: Nothing used the pointer interface. I figure I might as well make this consistent. Shouldn't cause any issues afaict.

ferry.go
```diff
@@ -369,10 +349,10 @@ func checkConnection(logger *logrus.Entry, dsn string, db *sql.DB) error {
 	hasSSL := cipher != ""
 
 	logger.WithFields(logrus.Fields{
-		"hasSSL":     hasSSL,
-		"ssl_cipher": cipher,
-		"dsn":        dsn,
-	}).Infof("connected to %s", dsn)
+		"hasSSL":      hasSSL,
+		"ssl_cipher":  cipher,
+		"destination": destination,
```
hkdsun: this dsn was pretty useful to have 🤔

config.go
hkdsun: like justin, i think that we'll need this to be a pointer if we're gonna cache connections in it

ferry.go
shuhaowu: It still shows up, just in the line immediately above, which will be tagged with destination=source/target.

ferry.go
```diff
@@ -100,23 +100,13 @@ func (f *Ferry) Initialize() (err error) {
...
-		f.logger.WithError(err).Error("failed to build config for source database")
-		return err
-	}
-
-	sourceDSN := sourceConfig.FormatDSN()
-	maskedSourceDSN := MaskedDSN(sourceConfig)
-
-	f.logger.WithField("dsn", maskedSourceDSN).Info("connecting to the source database")
-	f.SourceDB, err = sql.Open("mysql", sourceDSN)
+	f.SourceDB, err = f.Source.SqlDB(f.logger.WithField("destination", "source"))
```
pushrax: `destination` is a bit confusing for me since it has a similar meaning to `target`

hkdsun: could be `db` or `dbName` or something

shuhaowu: Yeah I didn't have the best name for it, I'll push a commit changing it to dbname.

PR #30, opened 2018-05-28
 WaitUntilReplicaIsCaughtUpToMaster in Ghostferry core

shuhaowu:
**Note: the current implementation is NOT hooked up to sharding/copydb. The present work focuses on the main algorithm. Implementation inside sharding/copydb can come in follow up PRs.**

### Problem

If we are copying from a slave source database, we must ensure that the slave is up to date after making the source read only before flushing and stopping the binlog streamer. Without doing so, a race condition exists to cause data corruption. As an example:

1. User sets source master writer to READONLY. Suppose master binlog position    is now 10 and slave binlog position is 8.
2. Ghostferry request for binlog streamer to stop. The binlog streamer sets the target stop position 8.
3. Ghostferry misses binlog entry 9 and 10, causing data corruption.

### Solution

This commit fixes the problem by making sure the replica is caught up to the master by introducing a optional loop checking this before we flush and stop the binlog streamer. This is accomplished by:

1. User sets source master writer to READONLY.
2. Ghostferry connects to the master writer and reads the binlog position via SHOW MASTER STATUS. This position is saved.
3. Ghostferry connects to the slave and reads the current master binlog position via a custom query. As an example, you could SELECT the master node from the pt-heartbeat table and thus discover its binlog position that has been replicated to the slave.
4. Ghostferry continues to execute this custom query until the position is greater than the target position saved in step 2.

We use a heartbeat table because `SHOW SLAVE STATUS` does not give the correct result if the slave is not replicating directly from the master, as in a case with a complex replication topology. However, it is possible to write some code to support the case of `SHOW SLAVE STATUS` as the interface is general.

### Implementation detail

Currently it is implemented as a class you can optionally initialize in the sharding/copydb/custom Ferry and passed into `ghostferry.Ferry`. For now, the way we get the master binlog position that has been replicated to the slave is via a custom query specified by the consumer of this API. An example is given in the comment/test.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -3,11 +3,14 @@ package ghostferry
 import (
 	"context"
 	"crypto/rand"
+	"database/sql"
 	"encoding/binary"
+	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/siddontang/go-mysql/mysql"
 	"github.com/sirupsen/logrus"
 )
 
@@ -138,4 +141,27 @@ loop:
 		}
 	}
 	return results, err
+}
+
+func ShowMasterStatusBinlogPosition(db *sql.DB) (mysql.Position, error) {
+	row := db.QueryRow("SHOW MASTER STATUS")
+	var file string
+	var position uint32
+	var binlog_do_db, binlog_ignore_db, executed_gtid_set string
+	err := row.Scan(&file, &position, &binlog_do_db, &binlog_ignore_db, &executed_gtid_set)
+	switch {
+	case err == sql.ErrNoRows:
+		return mysql.Position{}, fmt.Errorf("no results from show master status")
+	case err != nil:
+		return mysql.Position{}, err
+	default:
+		if file == "" {
+			return mysql.Position{}, fmt.Errorf("show master status does not show a file")
+		}
+
+		return mysql.Position{
+			Name: file,
+			Pos:  position,
+		}, nil
+	}
 }
--- a/wait_until_replica_is_caught_up_to_master.go
+++ b/wait_until_replica_is_caught_up_to_master.go
@@ -1,0 +1,125 @@
+package ghostferry
+
+import (
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/siddontang/go-mysql/mysql"
+	"github.com/sirupsen/logrus"
+)
+
+type ReplicatedMasterPositionFetcher interface {
+	Current(*sql.DB) (mysql.Position, error)
+}
+
+// Selects the master position that we have replicated until from a heartbeat
+// table of sort.
+type ReplicatedMasterPositionViaCustomQuery struct {
+	// The custom query executing should return a single row with two values:
+	// the string file and the integer position. For pt-heartbeat, this query
+	// would be:
+	//
+	// "SELECT file, position FROM meta.ptheartbeat WHERE server_id = %d" % serverId
+	//
+	// where serverId is the master server id, and meta.ptheartbeat is the table
+	// where pt-heartbeat writes to.
+	//
+	// For pt-heartbeat in particular, you should not use the
+	// relay_master_log_file and exec_master_log_pos of the DB being replicated
+	// as these values are not the master binlog positions.
+	Query string
+}
+
+func (r ReplicatedMasterPositionViaCustomQuery) Current(replicaDB *sql.DB) (mysql.Position, error) {
+	var file string
+	var pos uint32
+	row := replicaDB.QueryRow(r.Query)
+	err := row.Scan(&file, &pos)
+	if err != nil {
+		return mysql.Position{}, nil
+	}
+
+	switch {
+	case err == sql.ErrNoRows:
+		return mysql.Position{}, fmt.Errorf("no results from query")
+	case err != nil:
+		return mysql.Position{}, err
+	default:
+		if file == "" {
+			return mysql.Position{}, fmt.Errorf("query does not return a file")
+		}
+
+		return mysql.Position{
+			Name: file,
+			Pos:  pos,
+		}, nil
+	}
+}
+
+// Only set the MasterDB and ReplicatedMasterPosition options in your code as
+// the others will be overwritten by the ferry.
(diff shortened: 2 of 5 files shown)


shuhaowu:
One thing I'm slightly worried about this setup is that we have no integration test (although there is an unittest). It's also unclear how we can even test it in this codebase with meaningful results.

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -0,0 +1,125 @@
...
+		return mysql.Position{
+			Name: file,
+			Pos:  pos,
+		}, nil
+	}
+}
+
+// Only set the MasterDB and ReplicatedMasterPosition options in your code as
+// the others will be overwritten by the ferry.
+type WaitUntilReplicaIsCaughtUpToMaster struct {
```
hkdsun: maybe name this `SlaveReplicationWaiter` in the spirit of the other vocabulary suggestion (`SlavePosition`)?

utils.go
```diff
@@ -139,3 +142,26 @@ loop:
 	}
 	return results, err
 }
+
+func ShowMasterStatusBinlogPosition(db *sql.DB) (mysql.Position, error) {
+	row := db.QueryRow("SHOW MASTER STATUS")
+	var file string
+	var position uint32
+	var binlog_do_db, binlog_ignore_db, executed_gtid_set string
```
hkdsun: all of these appear to be unused. can we get rid of them?

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return WithRetries(100, 600*time.Millisecond, w.logger, "read master binlog position", func() error {
+		var err error
+		w.targetMasterPos, err = ShowMasterStatusBinlogPosition(w.MasterDB)
+		return err
+	})
+}
+
+func (w *WaitUntilReplicaIsCaughtUpToMaster) IsCaughtUp() (bool, error) {
+	var currentReplicatedMasterPos mysql.Position
```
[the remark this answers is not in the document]
shuhaowu: Hmm.. I'm not sure if `currentReplicaPosition` is the correct name here. The position referred here is the position of the master that the replica has caught up to. `currentReplicaPosition` would imply to mean the binlog position of the replica, which is not the same as the master position due to the implementation of MySQL's binlog position.

utils.go
shuhaowu: No we cannot. `row.Scan` expects the same number of arguments as the columns returned. `SHOW MASTER STATUS` always return these columns and we don't really have an ability to filter them.

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/siddontang/go-mysql/mysql"
+	"github.com/sirupsen/logrus"
+)
+
+type ReplicatedMasterPositionFetcher interface {
```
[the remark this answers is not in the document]
hkdsun: I was going for trading some ambiguity for more simplicity but the verbose terminology is also fair enough

shuhaowu:
I've fixed everything except the verbose terminologies as agreed. Any other comments before I merge this?

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -0,0 +1,107 @@
...
+	for !isCaughtUp {
+		isCaughtUp, err = w.IsCaughtUp()
+		if err != nil {
+			w.logger.WithError(err).Error("failed to get replica binlog coordinates")
+			w.ErrorHandler.Fatal("wait_replica", err)
+			return
+		}
+
+		time.Sleep(600 * time.Millisecond)
+	}
```
pushrax: nitpick: could be

```go
for {
  isCaughtUp, err := w.IsCaughtUp()
  if err != nil {
    ...
  }
  if isCaughtUp {
    break
  }
  time.Sleep(600 * time.Millisecond)
}
```

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		w.logger.WithError(err).Error("failed to get master binlog coordinates")
+		w.ErrorHandler.Fatal("wait_replica", err)
+		return
+	}
+
+	w.logger.Infof("target master position is: %v\n", w.targetMasterPos)
+
+	isCaughtUp := false
+	for !isCaughtUp {
```
pushrax: Should there be some timeout on this?

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -0,0 +1,107 @@
...
+	ReplicatedMasterPositionFetcher ReplicatedMasterPositionFetcher
+
+	ReplicaDB    *sql.DB
+	ErrorHandler ErrorHandler
+
+	logger          *logrus.Entry
+	targetMasterPos mysql.Position
+}
+
+func (w *WaitUntilReplicaIsCaughtUpToMaster) Start() error {
```
pushrax: `MarkTargetMasterPosition`?

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (w *WaitUntilReplicaIsCaughtUpToMaster) Wait() {
+	err := w.Start()
+	if err != nil {
+		w.logger.WithError(err).Error("failed to get master binlog coordinates")
+		w.ErrorHandler.Fatal("wait_replica", err)
+		return
+	}
+
+	w.logger.Infof("target master position is: %v\n", w.targetMasterPos)
```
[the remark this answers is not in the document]
pushrax: Heartbeat in a goroutine?

shuhaowu:
There are still some changes that needs to be made, but will be done in a follow up PR:

- [ ] Make the struct use only 1 method and still have the ability to test it somehow.
- [ ] Configurable timeout to make sure we don't block forever.
- [ ] Don't use the error handler, just return an error.

PR #31, opened 2018-05-29
Deploy packages to Shopify's packagecloud

hkdsun:
We need this back ahead of other work being done as our internal applications depend on it.

@Shopify/pods

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,3 +57,8 @@ deploy:
     file: build/ghostferry-*.deb
     on:
       tags: true
+  - provider: script
+    skip-cleanup: true
+    script: .travisci/deploy-packagecloud.sh
+    on:
+      branch: master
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -57,3 +57,8 @@ deploy:
     file: build/ghostferry-*.deb
     on:
       tags: true
+  - provider: script
+    skip-cleanup: true
+    script: "gem install package_cloud && package_cloud push --url https://packages.shopify.io shopify/ghostferry/ubuntu/trusty build/ghostferry-*.deb"
+    on:
+      all_branches: true
```
shuhaowu: All branches? Shouldn't it be master only?

hkdsun:
I think this is good to go:

image

Changed the branch back to `master` after the test above

PR #33, opened 2018-05-30
Deploy to packagecloud on staging-* branches in addition to master

hkdsun:
@Shopify/pods @shuhaowu


hkdsun:
Seems to be working for the current branch (`staging-test`):

image

PR #35, opened 2018-06-04
Allow CopyDB to run on a read only replica

shuhaowu:
This refactored the sanity check logic for the replication config into `ghostferry.Ferry` so that both sharding and copydb can benefit from it. This PR also adds the necessary configuration to copydb to allow it to run off of a read only replica.

I've also updated the local docker-compose setup to three nodes so people can test it easily with the example config.


sharding/sharding.go
```diff
@@ -149,15 +152,6 @@ func (r *ShardingFerry) Run() {
...
 
 	var err error
-
-	if r.config.RunFerryFromReplica {
-		err := r.initializeWaitUntilReplicaIsCaughtUpToMasterConnection()
-		if err != nil {
-			r.logger.WithField("error", err).Errorf("could not open connection to master replica")
-			r.Ferry.ErrorHandler.Fatal("sharding", err)
-		}
-	}
```
hkdsun: The reason this was here is that we pt-kill idle connections on the db. So, if we open the connection too soon it might get killed by the server by the time we need it.

shuhaowu: I thought the Go driver reconnect automatically?

shuhaowu:
Can I get a review on this PR soon?

shuhaowu:
I'm merging this for now as agreed upon for now so we can unblock further work.

PR #36, opened 2018-06-06
Iterative Verifier: keep on re-verifying until store is small enough

hkdsun:
This PR iterates on our previous heuristic for improving cutover phase verification performance for the iterative verifier. Rather than eagerly re-verifying only once to cut down the `reverifyStore` size, keep re-verifying until the overall size of the store does not improve between iterations – with a reasonable static `maxIterations` and a configurable `MaxDowntime` duration which will determine if the Ferry should proceed any further if the cutover phase verification is predicted to surpass that value.

The PR also adds some helpful metrics for having more visibility into long-running Ferries.

@Shopify/pods

### DIFF
--- a/iterative_verifier.go
+++ b/iterative_verifier.go
@@ -75,6 +75,7 @@ func (r *ReverifyStore) Add(entry ReverifyEntry) {
 		r.MapStore[tableId][entry.Pk] = struct{}{}
 		r.RowCount++
 		if r.RowCount%r.EmitLogPerRowCount == 0 {
+			metrics.Gauge("iterative_verifier_store_rows", float64(r.RowCount), []MetricTag{}, 1.0)
 			logrus.WithFields(logrus.Fields{
 				"tag":  "reverify_store",
 				"rows": r.RowCount,
@@ -132,11 +133,12 @@ type IterativeVerifier struct {
 	SourceDB         *sql.DB
 	TargetDB         *sql.DB
 
-	Tables           []*schema.Table
-	IgnoredTables    []string
-	DatabaseRewrites map[string]string
-	TableRewrites    map[string]string
-	Concurrency      int
+	Tables              []*schema.Table
+	IgnoredTables       []string
+	DatabaseRewrites    map[string]string
+	TableRewrites       map[string]string
+	Concurrency         int
+	MaxExpectedDowntime time.Duration
 
 	reverifyStore *ReverifyStore
 	logger        *logrus.Entry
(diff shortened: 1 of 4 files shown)


iterative_verifier.go
```diff
@@ -239,10 +241,37 @@ func (v *IterativeVerifier) VerifyBeforeCutover() error {
...
+
+		_, err := v.verifyStore("reverification_before_cutover", []MetricTag{{"iteration", string(iteration)}})
+		if err != nil {
+			return err
+		}
+
+		after := v.reverifyStore.RowCount
+		timeToVerify = time.Now().Sub(start)
+
+		if after >= before {
```
sirupsen: Might it be worth quitting if you're either _above_, or _below_ by e.g. 2-3%? Too complicated?

sirupsen: I like this heuristic much more than time, by the way. So much simpler.

iterative_verifier.go
```diff
@@ -239,10 +241,37 @@ func (v *IterativeVerifier) VerifyBeforeCutover() error {
...
+			break
+		}
+	}
+
+	if v.MaxDowntime != 0 && timeToVerify > v.MaxDowntime {
+		return fmt.Errorf("Unable to verify data within max downtime duration (took %s)", timeToVerify)
```
fw42: I find it confusing to call this "downtime". Doesn't this run before the cutover? So the shop isn't down yet and reverifying it more often doesn't increase the downtime.

hkdsun: The rationale here is that if we ran our heuristic to completion, then we expect the cutover phase verification to take at least `timeToVerify` duration. This is why I chose `MaxDowntime`. Do you have any suggestions for more clarity?

fw42: MaxReverificationTime?

fw42: MaxExpectedDowntime?

iterative_verifier.go
hkdsun: I'm skeptical about whether the 2-3% breakout is worth the additional math (not that it's too complicated)

iterative_verifier.go
```diff
@@ -239,10 +241,37 @@ func (v *IterativeVerifier) VerifyBeforeCutover() error {
...
+
+		_, err := v.verifyStore("reverification_before_cutover", []MetricTag{{"iteration", string(iteration)}})
+		if err != nil {
+			return err
+		}
+
+		after := v.reverifyStore.RowCount
+		timeToVerify = time.Now().Sub(start)
+
+		if after <= 1000 || after >= before {
```
pushrax: A log with timeToVerify, before, after, and iteration would be good here.

PR #39, opened 2018-06-12
sharding: fix duration config 

hkdsun:
Follow up to https://github.com/Shopify/ghostferry/pull/36

For some reason I had thought using `30[s|d|h]` notation is the canonical way of unmarshalling a time.Duration. Guess not 😬 

@pushrax @fw42


hkdsun:
fixed

PR #42, opened 2018-06-15
Add missing ; to tutorial

arthurnn:
Spotted some missing semi-colons that makes the create db .sql script invalid


sirupsen:
How's it going Arthur! I found a magic mouse the other day...

image

PR #43, opened 2018-06-26
TLA+ for reconciliation process after resuming

shuhaowu:
When we interrupt Ghostferry, we except to resume later. The two important things are: catching up the changes done to the source database by reading the binlogs and memorizing and restarting from a known PK position.

The changes to the source database while Ghostferry is down is performed by the Ferry process, as I haven't figured out a way generate a modified SourceTable directly during the Init. During the process that the source changes are initialized by the Ferry, the application is not writing to the database to simplify the model. Once the reconciliation starts, the modeled application begins writing to the database.

This is still kind of a WIP as I'm not 100% confident that it is correct. The model with the reconciliation is also not fast to verify as it took ~50min on my laptop (albeit within a VM), checking ~660M states (~190M unique states). This run is subject to an action constraint where the BinlogSize is restricted to 5 (with 3 already filled due to the changes that occurred while Ghostferry is theoretically down, so realistically 2 changes after starting Ghostferry). If the BinlogSize is restricted to 4, the amount of time reduce to ~5min.


tlaplus/ghostferry.tla
```diff
@@ -386,51 +494,90 @@ PossibleRecords == Records \cup {NoRecordHere}
  ***************************************************************************)
 \* BEGIN TRANSLATION
```
shuhaowu: Everything below this is just automatically generated by TLA+.

shuhaowu:
One thing that's still need to be implemented is check if the starting PK/binlog position is less than the actual positions that have been copied, if the algorithm will still work. I believe it will, but having it here would make sure of that fact.

This would make verifying that interrupting easier: we just need to ensure that we copied at least as much as the final saved location, which is an easy invariant to write.

PR #44, opened 2018-07-09
Add verification of snappy-compressed data

fjordan:
This is an initial commit of adding verification of snappy compressed data. This approach adds a `CompressionVerifier` that will allow `Snappy` and other compression algorithms to be added to decompress data to be fingerprinted and verified. A component is added to the configuration asking the user to identify the compression algorithms used, if any,  in corresponding tables and columns. 

The process to decompress the data is done during the call to the `IterativeVerifier`'s `GetHashes` method [here](https://github.com/Shopify/ghostferry/pull/44/files#diff-f02ffb9491865803e5c8fb0ecf6ef3caR604). Is there a better approach?

### Questions
Q: Why do we need this? 
A: Certain compression algorithms may create different hashes of data depending on the version, hardware, or simply because they are not deterministic (Snappy being one of them). Because of this, we cannot blindly rely on the `md5` hash of the compressed data in the tables to verify equality. We must first decompress the data and then we can fingerprint for equality.

Q: Does the `CompressionVerifier` warrant its own test suite? Or should the functionality be included in the existing `IterativeVerifier`'s test suite?
A: Nope. Not right now.

### To be completed
- [x] Validation of approach
- [x] Tests with fixture data
- [x] Metrics
- [x] Logging

@Shopify/pods


compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (e UnsupportedCompressionError) Error() string {
+	return "Compression algorithm: " + e.algorithm +
+		" not supported on table: " + e.table +
+		" for column: " + e.column
+}
+
+// CompressionVerifier provides the same functionality as the iterative verifier but also
+// provides support for manually verifying the payload of compressed columns that may
+// have different hashes for the same data
```
fw42: Super nitpick: I think this is a tiny bit confusing. The "root problem" is that there are multiple compressed versions of the same data. The fact that there are also (as a result of this) multiple hashes for the same (decompressed) data is a side-effect of that, not the actual problem (and it's confusing since hash functions themselves can't have this property, it's the combination with compression functions that causes this problem).

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+type CompressionVerifier struct {
+	logger *logrus.Entry
+
+	// supportedAlgorithms provide O(1) lookup to check if a configured algorithm is supported
```
fw42: `s/provide/provides/`?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+	supportedAlgorithms map[string]struct{}
+
+	// tableCmpressions is represented as table[column][compression_algorithm]
```
fw42: `s/Cmp/Comp/`

fw42: Nitpick: Took me a minute to understand what you mean by `table[column][compression_algorithm]`. Is that a common notation? How about something like `tableName -> columnName -> compressionAlgorithmName`?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+	tableCompressions map[string]map[string]string
```
fw42: nitpick: You call this `TableColumnCompression` below, I like that name better 😄

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// NewCompressionVerifier first checks the map for supported compression algorithms before
+// initializing and returning the initialized instance.
+func NewCompressionVerifier(tableCompressions map[string]map[string]string) (*CompressionVerifier, error) {
+	supportedAlgorithms := make(map[string]struct{})
```
fw42: nitpick: Could `supportedAlgorithms` maybe be a constant instead of a variable on every single `CompressionVerifier` "instance"? After all this can't possibly change (without recompiling the binary).

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+
+	return nil
+}
+
+// GetHashes compares the source data with the target data to ensure the integrity of the
+// data being copied.
+//
+// The CompressionVerifier method checks if the existing table contains compressed data
```
fw42: nitpick: Saying "The CompressionVerifier method checks ..." is a tiny bit confusing since CompressionVerifier is not a method.

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+		OrderBy(quotedPK).
+		ToSql()
+
+	if err != nil {
+		return nil, err
+	}
+
+	// This query must be a prepared query. If it is not, querying will use
+	// MySQL's plain text interface, which will scan all values into []uint8
+	// if we give it []interface{}.
```
fw42: Instead of (or in addition to) having this comment, can we have a test that fails if someone changes this from prepared to non-prepared? (i.e. a test that captures the bug that would appear if this were not a prepared statement)

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+	rows, err := c.getRows(db, schema, table, pkColumn, columns, pks)
+	if err != nil {
+		return nil, err
+	}
+
+	// Decompress applicable columns and hash the resulting column values for comparison
+	return c.decompressAndHash(table, columns, rows, tableCompression)
+}
+
+func (c *CompressionVerifier) getRows(db *sql.DB, schema, table, pkColumn string, columns []schema.TableColumn, pks []uint64) (*sql.Rows, error) {
```
fw42: This function seems very generic and totally unrelated to compression. Feels like maybe this should be on some other struct and not on `CompressionVerifier`? Do we have this code elsewhere already from which we could reuse it?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+	rows, err := stmt.Query(args...)
+	if err != nil {
+		return nil, err
+	}
+
+	return rows, nil
+}
+
+func (c *CompressionVerifier) decompressAndHash(table string, columns []schema.TableColumn, rows *sql.Rows, tableCompression map[string]string) (map[uint64][]byte, error) {
+	defer rows.Close()
```
fw42: Nitpick: I feel like this line should probably be in `GetHashes`, not in `decompressAndHash`.

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+	resultSet := make(map[uint64][]byte)
+	for rows.Next() {
+		rowData, err := ScanGenericRow(rows, 2)
```
fw42: This hardcoded `2` seems odd. Shouldn't this say something like `len(columns)`? Can we add a test for a table that has more than one compressed column?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+
+		pk, err := rowData.GetUint64(0)
+		if err != nil {
+			return nil, err
+		}
+
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decodedRowData contains a map of all
+		// columns and associated decoded values by the index of the column
+		var decodedRowData map[int][]byte
```
fw42: nitpick: `uint`? column index can't be negative, can it?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+		for idx, column := range columns {
+			// Check if column is configured as compressed and decompress if necessary
+			if algorithm, ok := tableCompression[column.Name]; ok {
+				decodedColData, err := c.Decode(table, column.Name, algorithm, rowData[idx].([]byte))
+				if err != nil {
+					return nil, err
+				}
+				decodedRowData[idx] = decodedColData
+			} else {
+				// Do not compress the column, add it to the decodedRowData to be fingerprinted
```
fw42: `s/compress/decompress/`

iterative_verifier.go
```diff
@@ -601,6 +602,11 @@ func compareHashes(source, target map[uint64][]byte) []uint64 {
 }
 
 func (v *IterativeVerifier) GetHashes(db *sql.DB, schema, table, pkColumn string, columns []schema.TableColumn, pks []uint64) (map[uint64][]byte, error) {
+	if tableCompression, ok := v.CompressionVerifier.tableCompressions[table]; ok {
```
fw42: nitpick: I find it confusing that there are two things that are both called `tableCompression` but one is a `map[string]map[string]string` while the other is a `map[string]string`. Can we use different terminology? Maybe one is called `tableCompressions` and the other (the smaller one) is called `columnCompressions`?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decodedRowData contains a map of all
+		// columns and associated decoded values by the index of the column
+		var decodedRowData map[int][]byte
+		for idx, column := range columns {
+			// Check if column is configured as compressed and decompress if necessary
+			if algorithm, ok := tableCompression[column.Name]; ok {
+				decodedColData, err := c.Decode(table, column.Name, algorithm, rowData[idx].([]byte))
```
fw42: nitpick: maybe `decodedCellData`, since this is only one cell (not the entire column)?

compression_verifier.go
fw42: nitpick: Can we call this `decompressedRowData`? No need to introduce more terminology than necessary (decode vs. decompress, I think it's less confusing if we are consistent, otherwise the next person might ask themselves what the difference is while you are actually using them interchangeably).

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+
+		resultSet[pk] = decodedRowHash
+	}
+
+	return resultSet, nil
+}
+
+// Decode will apply the configured decompression algorithm to the configured columns data
+func (c *CompressionVerifier) Decode(table, column, algorithm string, compressed []byte) ([]byte, error) {
```
fw42: Same here, for clarity's sake, can we call it `Decompress` instead of `Decode`?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decodedRowData contains a map of all
+		// columns and associated decoded values by the index of the column
+		var decodedRowData map[int][]byte
+		for idx, column := range columns {
+			// Check if column is configured as compressed and decompress if necessary
+			if algorithm, ok := tableCompression[column.Name]; ok {
+				decodedColData, err := c.Decode(table, column.Name, algorithm, rowData[idx].([]byte))
+				if err != nil {
+					return nil, err
```
fw42: How do you feel about ignoring errors here? Imagine we have corrupted data in a compressed column, i.e. data that can't be decompressed (for example because it was truncated). With your code here, we will never ever be able to copy that data. But (at this layer) we actually don't care about the data being decompressable, we only care about it being equal to the data in the other database. So as long as both the source and the target are the same (even if both are corrupted, as long as they are corrupted in the exact same way), we should not treat that as an error (if you agree, can you add a test for that?). I'd suggest to treat decompression errors as "this column wasn't meant to be decompressed anyway" and just store it in `decodedRowData` as raw compressed data (same as the `else` branch below) and then give it a second chance by comparing using the regular md5 approach.

Not sure if this actually ever happens in practice, but (if I remember correctly) our current Ruby implementation of this algorithm doesn't have that problem since it only decompresses if the fingerprints are not equal already.

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// Decode will apply the configured decompression algorithm to the configured columns data
+func (c *CompressionVerifier) Decode(table, column, algorithm string, compressed []byte) ([]byte, error) {
+	var decoded []byte
+	switch algorithm {
+	case CompressionSnappy:
+		return snappy.Decode(decoded, compressed)
+	default:
+		return nil, UnsupportedCompressionError{
```
fw42: Can this ever happen? Don't you check for this in the initializer already?

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+			table:     table,
+			column:    column,
+			algorithm: algorithm,
+		}
+	}
+
+}
+
+// HashRow will fingerprint the non-primary columns of the row to be verify data equality
```
fw42: `s/to be/to/`

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+	// supportedAlgorithms provide O(1) lookup to check if a configured algorithm is supported
+	supportedAlgorithms map[string]struct{}
+
+	// tableCmpressions is represented as table[column][compression_algorithm]
+	tableCompressions map[string]map[string]string
+}
+
+// NewCompressionVerifier first checks the map for supported compression algorithms before
+// initializing and returning the initialized instance.
+func NewCompressionVerifier(tableCompressions map[string]map[string]string) (*CompressionVerifier, error) {
```
fw42: This method isn't used anywhere (only in the tests). That seems a bit odd. Am I missing something?

fw42:
I'd like Forrest to have some feedback ready to act on as soon as he gets back. Not super urgent but if anyone has some time to take a look, I think that would help make sure that Forrest isn't still blocked on this when he returns from vacation.

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/siddontang/go-mysql/schema"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	// CompressionSnappy is used to identify Snappy (https://google.github.io/snappy/) compressed column data
+	CompressionSnappy = "SNAPPY"
+)
+
+// UnsupportedCompressionError is used to identify a
```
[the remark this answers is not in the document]
fjordan: Fixed!

compression_verifier.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (e UnsupportedCompressionError) Error() string {
+	return "Compression algorithm: " + e.algorithm +
+		" not supported on table: " + e.table +
+		" for column: " + e.column
+}
+
+// CompressionVerifier provides the same functionality as the iterative verifier but also
+// provides support for manually verifying the payload of compressed columns that may
```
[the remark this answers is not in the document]
fjordan: Good feedback. Changed to just "verifying" as that makes more sense. 👍

compression_verifier.go
fjordan: Nice call out. Updated the comment to be a bit more clear and not confuse the two. What do you think?

compression_verifier.go
fjordan: oops 🤕

compression_verifier.go
fjordan: Yeah the name `tableColumnCompressions` is a bit more clear and helps the reader understand the data structure. Updated 😄

compression_verifier.go
fjordan: This should have said, `GetHashes`. Good catch!

compression_verifier.go
fjordan: Go doesn't allow map constants, but we could scope this at the package level and put it in `init` (although it wouldn't be exported so it would be local to this file). I usually try to avoid that because it [adds a bit of magic and is bad](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html), though I did consider it. Keeping it in the constructor does scope it to the instantiated `CompressionVerifier`, and only one instance should ever be needed.

compression_verifier.go
fjordan: This was copied from code in the `IterativeVerifier` 🍝. I'll look into catching this in a test (good idea).

compression_verifier.go
fjordan: Yeah, the `2` is a bug and needs to be addressed. Will definitely be fixed and verified in tests

compression_verifier.go
fjordan: Good call out on the `uint` and naming inconsistency. It should be standardized.

hkdsun:
> I think defaulting to MD5 and only decompressing on mismatches might be a big performance win compared to the approach in this PR (always decompress).

In my opinion, there won't be a huge performance win overall since this configuration is a very special case that I imagine will be use infrequently by the average user. 

What you suggested, in my opinion, will make the verification logic more complex to follow and since we can avoid it without a huge performance hit, I prefer the current approach

fw42:
> there won't be a huge performance win overall since this configuration is a very special case that I imagine will be use infrequently by the average user.

Not sure what you mean by "average user", but Shopify's use-case will run into this for every single shop, potentially tens of thousands of times, depending on how many rows the table has.

hkdsun:
My justification was that majority of tables will not go through this codepath - therefore the overall ferry performance will not be affected by these changes. So even if this is used for every ferry run, for every shop, all the other tables (which are much larger) dwarf the time spent in this codepath.

compression_verifier.go
```diff
@@ -0,0 +1,225 @@
...
+// may have different hashes for the same data by first decompressing the compressed
+// data before fingerprinting
+type CompressionVerifier struct {
+	logger *logrus.Entry
+
+	// supportedAlgorithms provide O(1) lookup to check if a configured algorithm is supported
+	supportedAlgorithms map[string]struct{}
+
+	// tableColumnCompressions is represented as table -> column -> compression_algorithm
+	tableColumnCompressions map[string]map[string]string
```
hkdsun: I think using a simple type annotation here would greatly improve readability. Something like:

```
type ColumnCompressionConfig map[string]string // map of columns to decompressors. e.g. { "body": "SNAPPY" }
```

you'd then have:

```
tableColumnCompressions map[string]ColumnCompressionConfig
```

which is a hash of `table -> ColumnCompressionConfig` which is readily comprehensible

compression_verifier.go
fjordan: I think what threw me off was seeing the snappy libraries use of `Decode` versus `Decompress`. I'll change it everywhere to use `Decompress` instead but obviously can't change the snappy library's API.

compression_verifier.go
fjordan: Does calling it `Col`/`Column` make you think we're decompressing the entire column of the table and not the column bound by the row's PK/idx? I haven't seen any other libs refer to this as a `cell`, but if it's confusing we can certainly change it.

compression_verifier.go
fjordan: Great call out. If there is a decompression error then surely it will cause problems in other places so it should definitely be logged (and possible make some noise?). 

We can continue execution and yes just confirm equality, however, if we're only going down this path when the fingerprints of the SQL don't match (as you point out in your other comment), and now the data cannot be decompressed, then we won't be able to confirm the equality of the data, right? 🤔

compression_verifier.go
fjordan: Changed, but yeah it was definitely the `Snappy` library that threw me off and caused me to start naming things using `decode` instead of just `decompress`.

compression_verifier.go
fjordan: It could happen if the method is called outside of the workflow of the `CompressionVerifier` and in an ad hoc sense. Should we not export this method?

compression_verifier.go
fw42: @hkdsun, how do you feel about this comment? i.e. ALWAYS decompressing makes it impossible to move data that is corrupted (while that was previously possible).

compression_verifier.go
fjordan: fixed!

compression_verifier.go
fjordan: fixed

compression_verifier.go
fjordan: Added two types, one being exported (`TableColumnCompressionConfig`). Do you think this is more clear?

compression_verifier.go
fjordan: It may or may not be used depending on the user's use case. If it were to be used, this function is provided as a convenience and safeguard to ensure the configured compression algorithms are supported.

compression_verifier.go
fjordan: @fw42 just confirmed with @hkdsun that moves are aborted if the data cannot be decompressed and the compressed hashes do not match. Planning to implement this here. What are your thoughts?

fw42: Yes, that's the current behaviour in the Ruby implementation.

compression_verifier.go
fjordan: Added https://github.com/Shopify/ghostferry/issues/52

fjordan:
This PR now only uses the `CompressionVerifier` on mismatch and a compression configuration check. Do we have any fixture data we can share with a mismatched hash that can be used in the test?

iterative_verifier.go
```diff
@@ -573,7 +574,27 @@ func (v *IterativeVerifier) compareFingerprints(pks []uint64, table *schema.Tabl
 		return nil, targetErr
 	}
 
-	return compareHashes(sourceHashes, targetHashes), nil
+	mismatches := compareHashes(sourceHashes, targetHashes)
+	if len(mismatches) > 0 && v.CompressionVerifier != nil {
```
fw42: I feel like this block of code here could be it's own function to make things more readable.
```go
mismatches := compareHashes(sourceHashes, targetHashes)

if mismatchInCompressedColumns(mismatches) {
  return compareCompressedColumns(...)
} else {
  return mismatches, nil
}
```
wdyt?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	if err := compressionVerifier.verifyConfiguredCompression(tableColumnCompressions); err != nil {
+		return nil, err
+	}
+
+	return compressionVerifier, nil
+}
+
+func (c *CompressionVerifier) verifyConfiguredCompression(tableColumnCompressions TableColumnCompressionConfig) error {
```
hkdsun: I think it's convention in the project to keep exported methods on top of the file and push all unexported methods down

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+
+		pk, err := rowData.GetUint64(0)
+		if err != nil {
+			return nil, err
+		}
+
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decompressedRowData contains a map of all
+		// columns and associated decompressed values by the index of the column
+		decompressedRowData := make(map[uint64][]byte)
```
hkdsun: why does this need to be a map? the map's keys are never used

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+	return nil
+}
+
+// GetCompressedHashes compares the source data with the target data to ensure the integrity of the
+// data being copied.
+//
+// The GetCompressedHashes method checks if the existing table contains compressed data
+// and will apply the decompression algorithm to the applicable columns if necessary.
+// After the columns are decompressed, the hashes of the data are used to verify equality
+func (c *CompressionVerifier) GetCompressedHashes(db *sql.DB, schema, table, pkColumn string, columns []schema.TableColumn, pks []uint64, tableCompression map[string]string) (map[uint64][]byte, error) {
```
hkdsun: `tableCompression` should be `columnCompressionConfig` instead of `map[string]string`, right?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+// may have different hashes for the same data by first decompressing the compressed
+// data before fingerprinting
+type CompressionVerifier struct {
+	logger *logrus.Entry
+
+	// supportedAlgorithms provide O(1) lookup to check if a configured algorithm is supported
+	supportedAlgorithms map[string]struct{}
+
+	// tableColumnCompressions is represented as table -> column -> compression-type
+	tableColumnCompressions TableColumnCompressionConfig
```
hkdsun: I don't understand why `tableColumnCompressions` is a field on `CompressionVerifier`. It's never accessed from within the methods here. It's in the constructor but that's just to populate the field – not consume it

The only other time I see it being accessed is from within the `IterativeVerifier`. Is that a sign that it should not be here? Perhaps it should just be a config on the `IterativeVerifier` directly?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decompressedRowData contains a map of all
+		// columns and associated decompressed values by the index of the column
+		decompressedRowData := make(map[uint64][]byte)
+		for idx, column := range columns {
+			if column.Name == pkColumn {
+				continue
+			}
```
hkdsun: Why are we excluding the primary key column from hashing/comparison?

I was under the impression that what we're trying to do is `SELECT pk_col, *` and then massaging the result to a `map[pk]rowData` where `rowData` is all the row columns after decompression and which must contain the pk column as well

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+			// Check if column is configured as compressed and decompress if necessary
+			if algorithm, ok := tableCompression[column.Name]; ok {
+				decompressedColData, err := c.Decompress(table, column.Name, algorithm, rowData[idx].([]byte))
```
hkdsun: I think `rowData[idx]` doesn't always map one-to-one to the ordering of the `columns` slice. Do you agree?

---

Imagine the following example table:

```
| col_1 | col_2 | pk_col | col_3 |
```

In this scenario, with the rowSelector that we've defined below, you'd have:

```
rowData = [pk_col, col_1, col_2, col_3]
```

whereas

```
columns = [col_1, col_2, pk_col, col_3]
```

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+				if err != nil {
```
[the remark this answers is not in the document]
hkdsun: If the fingerprints already match, we never get here due to the changes in `iterative_verifier.go`

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+// data being copied.
+//
+// The GetCompressedHashes method checks if the existing table contains compressed data
+// and will apply the decompression algorithm to the applicable columns if necessary.
+// After the columns are decompressed, the hashes of the data are used to verify equality
+func (c *CompressionVerifier) GetCompressedHashes(db *sql.DB, schema, table, pkColumn string, columns []schema.TableColumn, pks []uint64, tableCompression map[string]string) (map[uint64][]byte, error) {
+	c.logger.Info("decompressing table data before verification")
+
+	// Extract the raw rows using SQL to be decompressed
+	rows, err := c.getRows(db, schema, table, pkColumn, columns, pks)
```
hkdsun: I think a lot of confusion would be cleared up if the `getRows` function did the hard work of massaging `SELECT pk_col, *` to `map[pk]rowData` where `rowData`'s indexes map one-to-one with the `columns` slice.

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// GetCompressedHashes compares the source data with the target data to ensure the integrity of the
+// data being copied.
+//
+// The GetCompressedHashes method checks if the existing table contains compressed data
+// and will apply the decompression algorithm to the applicable columns if necessary.
+// After the columns are decompressed, the hashes of the data are used to verify equality
+func (c *CompressionVerifier) GetCompressedHashes(db *sql.DB, schema, table, pkColumn string, columns []schema.TableColumn, pks []uint64, tableCompression map[string]string) (map[uint64][]byte, error) {
+	c.logger.Info("decompressing table data before verification")
```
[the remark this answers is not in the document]
fjordan: I was following convention of other info statements I found in the iterative verifier that seemed to just notify the user of various steps in the verification process (https://github.com/Shopify/ghostferry/blob/master/iterative_verifier.go#L199, https://github.com/Shopify/ghostferry/blob/master/iterative_verifier.go#L208, 
 https://github.com/Shopify/ghostferry/blob/master/iterative_verifier.go#L219, etc.)

We can certainly add more info to this message. That makes sense given not all tables/columns should be decompressed.

compression_verifier.go
fjordan: I mistakenly thought the `IterativeVerifier` wasn't including the primary key for some reason in the fingerprint because it is selected in addition to the `row_fingerprint`. Looking at it again I see that it is selected separately in addition to it being included.

I'll make adjustments to include the PK in the compression verifier's row hash 👍

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+			return nil, err
+		}
+
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decompressedRowData contains a map of all
+		// columns and associated decompressed values by the index of the column
+		decompressedRowData := make(map[uint64][]byte)
+		for idx, column := range columns {
+			if column.Name == pkColumn {
+				continue
```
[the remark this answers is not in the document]
fjordan: Mistake on my understanding of how the `IterativeVerifier` was generating the `row_fingerprint` (also see comment below: https://github.com/Shopify/ghostferry/pull/44#discussion_r207946018). Will update 👍

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+func rowSelector(columns []schema.TableColumn, pkColumn string) sq.SelectBuilder {
+	quotedPK := quoteField(pkColumn)
+
+	columnStrs := make([]string, len(columns))
+	for idx, column := range columns {
+		if column.Name == pkColumn {
+			continue
+		}
+
+		quotedCol := normalizeAndQuoteColumn(column)
```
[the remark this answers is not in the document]
fjordan: Looks like we can remove it :)

compression_verifier.go
fjordan: It's currently part of the `CompressionVerifier` because it's the configuration for it. Given the `CompressionVerifier` is also a component of the `IterativeVerifier` it's essentially part of the `IterativeVerifier` already, but contained within the `CompressionVerifier` as the `CompressionVerifier` contains the logic and info for all things compressed. It's also not accessed within the `CompressionVerifier`, but referenced at runtime and then passed to the `GetCompressedHashes` method as that's how I understood your earlier comment about the direct logic flow: https://github.com/Shopify/ghostferry/pull/44#pullrequestreview-140863219. 

Does that make sense? If you feel we should still move it let me know so we can discuss because I want this to make sense 👍

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+		return nil, err
+	}
+
+	return compressionVerifier, nil
+}
+
+func (c *CompressionVerifier) verifyConfiguredCompression(tableColumnCompressions TableColumnCompressionConfig) error {
+	for table, columns := range tableColumnCompressions {
+		for column, algorithm := range columns {
+			// Ensure algorithm is uppercased
```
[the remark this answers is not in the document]
fjordan: removed 👍

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+			algorithm = strings.ToUpper(algorithm)
+			tableColumnCompressions[table][column] = algorithm
```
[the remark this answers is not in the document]
fjordan: That's interesting. What if a user configures their compression as `snappy` or `Snappy` while this method looks for `SNAPPY`? Would you rather it throw an error and tell the user to fix their capitalization? Or should we just handle it here for them?

fjordan:
@fw42 thanks for the comments and review! 😄  I've just pushed up a test that I believe satisfies your concerns. I was working on it last week (as we were discussing getting this fixture data for the tests over Slack) but ran into some encoding issues with the snappy compressed data that I didn't resolve until earlier this morning. I'm working to address the rest of your questions/comments/concerns. 

@hkdsun thanks 😄 ! We can certainly integrate this into the sharding package and/or `copydb`. It is already integrated into the `IterativeVerifier` if the user has configured the compression mappings. Also working to get your comments and feedback addressed. 

Thanks so much guys looking forward to getting this merged in and used once it's all ready to ship 🚢

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+
+	return nil
+}
+
+// GetCompressedHashes compares the source data with the target data to ensure the integrity of the
+// data being copied.
+//
+// The GetCompressedHashes method checks if the existing table contains compressed data
+// and will apply the decompression algorithm to the applicable columns if necessary.
+// After the columns are decompressed, the hashes of the data are used to verify equality
```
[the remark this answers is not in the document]
fjordan: @fw42 I made a change to only generate a client-side fingerprint on mismatch. See here: https://github.com/Shopify/ghostferry/pull/44/files#diff-f02ffb9491865803e5c8fb0ecf6ef3caR578

iterative_verifier.go
fjordan: That does look simpler, however, we would need to introduce some compression-specific logic to the `IterativeVerifier` to prevent duplication of code or flowing back and forth between the `CompressionVerifier` and `IterativeVerifier`. That would clean up this block of execution inside the conditional, but would blur the line of responsibility between these two components. We could do something like:
```go
if  v.compressionMismatch(mismatches) {
    sourceHashes, targetHashes, err = GetCompressedHashes(...)
    if err != nil {
        return nil, err
    }
    return compareHashes(sourceHashes, targetHashes), nil
}

func (v *IterativeVerifier) compressionMismatch(mismatches) {
    if len(mismatches) > 0 && v.CompressionVerifier != nil { 
        return true
    }
    return false
} 
```

and rename the current `GetCompressedHashes` to `GetCompressedTableHashes` and wrap the former to provide a clean interface.

Thoughts?

glide.yaml
```diff
@@ -0,0 +1,24 @@
...
+  subpackages:
+  - assert
+  - require
+  - suite
+- package: github.com/gorilla/mux
+  version: ^1.4.0
+- package: github.com/go-sql-driver/mysql
+  version: ^1.3.0
+- package: github.com/Shopify/go-dogstatsd
+- package: github.com/golang/snappy
```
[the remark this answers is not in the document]
fjordan: Oops 🤕 Will add!

config.go
```diff
@@ -153,12 +153,26 @@ type Config struct {
...
 	// the target database. See DatabaseRewrite.
 	//
 	// Optional: defaults to empty map/no rewrites
 	TableRewrites map[string]string
 
+	// Map of the table and column identifying the compression type
+	// (if any) of the column. This is used during verification  to ensure
+	// the data was successfully copied as it must be manually verified.
+	//
+	// Note that the VerifierType must be set to the IterativeVerifier
```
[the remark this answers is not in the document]
fjordan: Updated 👍

config.go
```diff
@@ -153,12 +153,26 @@ type Config struct {
...
 
-	// Map the table name on the source dataabase to a different name on
+	// Map the table name on the source database to a different name on
 	// the target database. See DatabaseRewrite.
 	//
 	// Optional: defaults to empty map/no rewrites
 	TableRewrites map[string]string
 
+	// Map of the table and column identifying the compression type
+	// (if any) of the column. This is used during verification  to ensure
```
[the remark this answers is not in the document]
fjordan: 👀

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+				}
+				decompressedRowData[uint64(idx)] = decompressedColData
+			} else {
+				// Do not decompress the column, add it to the decompressedRowData to be fingerprinted
+				decompressedRowData[uint64(idx)] = rowData[idx].([]byte)
+			}
+		}
+
+		// Hash the data of the row to be added to the result set
+		decompressedRowHash, err := c.HashRow(decompressedRowData)
```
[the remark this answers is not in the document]
fjordan: 🤔 Would it really be faster? It takes the m * n iterations down to just m iterations (one value to compare versus a value for each column) plus the cost to calculate the hash. Unless you're referring to just comparing the compressed column and allowing the `IterativeVerifier` to hash/fingerprint the other columns, which adds some complexity. What do you think?

compression_verifier.go
fjordan: 🙆 it should be a slice!

compression_verifier.go
fjordan: Yep! Surprised the compiler didn't catch this 🤔

compression_verifier.go
fjordan: How does the convention work between struct methods and functions? Is it:
1. Exported struct methods
2. Non-exported struct methods
3. Exported functions? 
4. Non-exported functions

Or are 3 and 1 swapped?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+	quotedPK := quoteField(pkColumn)
+
+	columnStrs := make([]string, len(columns))
+	for idx, column := range columns {
+		if column.Name == pkColumn {
+			continue
+		}
+
+		quotedCol := normalizeAndQuoteColumn(column)
+		columnStrs[idx] = fmt.Sprintf("COALESCE(%s, 'NULL')", quotedCol)
```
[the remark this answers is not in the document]
fjordan: This line just writes `NULL` instead of having an empty value for the column. Turns out we can remove it and just use `NULL`s. Why is this needed for the `IterativeVerifier`?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+	TableColumnCompressionConfig map[string]columnCompressionConfig
+)
+
+// UnsupportedCompressionError is used to identify errors resulting
+// from attempting to decompress unsupported algorithms
+type UnsupportedCompressionError struct {
+	table     string
+	column    string
+	algorithm string
+}
```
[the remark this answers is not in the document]
fjordan: We aren't handling it now, but once we integrate into `copydb` or `sharding` we may. It allows the type of error to be checked against if we do want any logical conditions based on the error. If there will never be a need we should just remove it, but I was thinking we may want it.

compression_verifier.go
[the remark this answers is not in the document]
fw42: Basically because you can't `MD5(NULL)` in MySQL, if I remember correctly.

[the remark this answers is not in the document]
fjordan: Hmm that doesn't appear to be the case.

```mysql
mysql> select md5(null)\G
*************************** 1. row ***************************
md5(null): NULL
1 row in set (0.00 sec)

mysql>
```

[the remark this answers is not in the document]
fw42: Hm then it was probably the `CONCAT` that needed this

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// tableColumnCompressions is represented as table -> column -> compression-type
+	tableColumnCompressions TableColumnCompressionConfig
+}
+
+// NewCompressionVerifier first checks the map for supported compression algorithms before
+// initializing and returning the initialized instance.
+func NewCompressionVerifier(tableColumnCompressions TableColumnCompressionConfig) (*CompressionVerifier, error) {
+	supportedAlgorithms := make(map[string]struct{})
+	supportedAlgorithms[CompressionSnappy] = struct{}{}
```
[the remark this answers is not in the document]
fjordan: Do we create multiple `IterativeVerifier`s or do we reuse a single one? If just a single instance then this will only be instantiated once. @fw42 mentioned this same thing, and I considered it, but didn't want to make a package-level global var. We can obviously if we intentionally want to do that, but it's [generally frowned upon](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html).

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+		" for column: " + e.column
+}
+
+// CompressionVerifier provides support for verifying the payload of compressed columns that
+// may have different hashes for the same data by first decompressing the compressed
+// data before fingerprinting
+type CompressionVerifier struct {
+	logger *logrus.Entry
+
+	// supportedAlgorithms provide O(1) lookup to check if a configured algorithm is supported
```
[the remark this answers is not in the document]
fjordan: 🔪 removed

compression_verifier.go
hkdsun: The convention is probably loosely followed but i'd also add: 

0. Exported struct constructors
1. ...

compression_verifier.go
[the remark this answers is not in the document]
hkdsun: I think this is fine.

I was originally thinking we could always switch on the upper-cased string (i.e. upper-case it every time it's being used):

```
	switch strings.ToUpper(algorithm) {
	case CompressionSnappy:
           ...
```

compression_verifier.go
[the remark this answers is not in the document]
hkdsun: Sure, I'm fine with keeping it since it's not a huge amount of complexity. 

Though, as a general principle, I like to avoid complexity if it's not being used immediately (and I don't see us handling this error any differently than any other verifier/initialization error when integrating into sharding package) – while keeping the future maintenance/changes in mind of course.

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	hash := md5.New()
+	var rowFingerprint []byte
+	for _, colData := range decompressedRowData {
+		rowFingerprint = append(rowFingerprint, colData...)
+	}
+
+	hash.Write(rowFingerprint)
+	return []byte(hex.EncodeToString(hash.Sum(nil))), nil
```
hkdsun: Note to other reviewers: this `Sum()` method is not the same as `md5.Sum()`

We're actually working with a `hash.Hash` implementation. See the example: https://golang.org/pkg/crypto/md5/#example_New

iterative_verifier.go
```diff
@@ -329,6 +297,79 @@ func (v *IterativeVerifier) Result() (VerificationResultAndStatus, error) {
 	return v.verificationResultAndStatus, v.verificationErr
 }
 
+func (v *IterativeVerifier) GetHashes(db *sql.DB, schema, table, pkColumn string, columns []schema.TableColumn, pks []uint64) (map[uint64][]byte, error) {
```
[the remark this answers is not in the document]
fjordan: Yes, I can undo and just let another PR re-order them.

iterative_verifier.go
```diff
@@ -573,6 +614,29 @@ func (v *IterativeVerifier) compareFingerprints(pks []uint64, table *schema.Tabl
 		return nil, targetErr
 	}
 
+	mismatches := compareHashes(sourceHashes, targetHashes)
+	if len(mismatches) < 0 {
```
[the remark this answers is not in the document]
fjordan: 👀 good catch

iterative_verifier.go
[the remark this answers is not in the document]
hkdsun: It's fine, just double checking

test/iterative_verifier_test.go
```diff
@@ -82,6 +105,28 @@ func (t *IterativeVerifierTestSuite) TestVerifyOncePass() {
...
+
+	result, err := t.verifier.VerifyOnce()
+	t.Require().NotNil(result)
+	t.Require().Nil(err)
+	t.Require().True(result.DataCorrect)
+	t.Require().Equal("", result.Message)
+}
+
+func (t *IterativeVerifierTestSuite) TestVerifyCompressedMismatchOncePass() {
+	t.InsertCompressedRowInDb(43, testhelpers.TestCompressedData3, t.Ferry.SourceDB)
```
hkdsun: Let's assert our test's assumption right before this with a helpful comment (after all it's the whole point of the PR):

```
         // Two fixtures that have different compressed values but have equal decompressed values
	t.Require().NotEqual(testhelpers.TestCompressedData3, testhelpers.TestCompressedData4)
```

test/iterative_verifier_test.go
```diff
@@ -95,6 +140,19 @@ func (t *IterativeVerifierTestSuite) TestBeforeCutoverFailuresFailAgainDuringCut
...
+	t.InsertCompressedRowInDb(42, testhelpers.TestCompressedData2, t.Ferry.TargetDB)
+
+	err := t.verifier.VerifyBeforeCutover()
+	t.Require().Nil(err)
+
+	result, err := t.verifier.VerifyDuringCutover()
+	t.Require().Nil(err)
+	t.Require().False(result.DataCorrect)
+	t.Require().Equal(fmt.Sprintf("verification failed on table: %s.%s for pks: %s", "gftest", testhelpers.TestCompressedTable1Name, "42"), result.Message)
+}
```
hkdsun: Can we possibly have a symmetrical test for the Data3 and Data4 case?

fjordan: Yep!

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+	rows, err := getRows(db, schema, table, pkColumn, columns, pks)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	// Decompress applicable columns and hash the resulting column values for comparison
+	resultSet := make(map[uint64][]byte)
+	for rows.Next() {
+		rowData, err := ScanByteRow(rows, len(columns)+1)
```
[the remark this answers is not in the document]
fw42: Why the `+1`? Is that because `columns` doesn't include the primary key column?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		pk, err := strconv.ParseUint(string(rowData[0]), 10, 64)
```
fw42: why do we have to parse the `pk` out of `rowData`? Don't we already have all the pks in `pks`?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decompressedRowData contains a map of all
+		// the non-compressed columns and associated decompressed values by the
+		// index of the column
+		decompressedRowData := [][]byte{}
+		for idx, column := range columns {
+			if algorithm, ok := tableCompression[column.Name]; ok {
+				// rowData contains the result of "SELECT pkColumn, * FROM ...", so idx+1 to get each column
+				decompressedColData, err := c.Decompress(table, column.Name, algorithm, rowData[idx+1])
```
fw42: same question here. why the `+1`? Is that because `rowData` contains the primary key column but `columns` doesn't?

fw42: nvm I just saw the comment 🤦‍♂️

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+		pk, err := strconv.ParseUint(string(rowData[0]), 10, 64)
+		if err != nil {
+			return nil, err
+		}
+
+		// Decompress the applicable columns and then hash them together
+		// to create a fingerprint. decompressedRowData contains a map of all
+		// the non-compressed columns and associated decompressed values by the
+		// index of the column
+		decompressedRowData := [][]byte{}
```
fw42: Seems like you already know exactly how long this array will be. Isn't it more idiomatic to `make` it here than rather than to `append` below? Seems more memory-efficient (fewer allocations). But probably no big deal and I honestly don't know which one is preferable. Just curious.

compression_verifier.go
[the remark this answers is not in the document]
fw42: wfm

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+		// Hash the data of the row to be added to the result set
+		decompressedRowHash, err := c.HashRow(decompressedRowData)
+		if err != nil {
+			return nil, err
+		}
+
+		resultSet[pk] = decompressedRowHash
+	}
+
+	metrics.Gauge("compression_verifier_decompress_rows", float64(len(resultSet)), []MetricTag{}, 1.0)
```
fw42: Maybe add a tag for the table name?

compression_verifier.go
```diff
@@ -0,0 +1,238 @@
...
+		return nil, errors.New("Row data to fingerprint must not be empty")
+	}
+
+	hash := md5.New()
+	var rowFingerprint []byte
+	for _, colData := range decompressedRowData {
+		rowFingerprint = append(rowFingerprint, colData...)
+	}
+
+	hash.Write(rowFingerprint)
```
fw42: `hash.Write` can return an error. We probably want to check that here?

config.go
```diff
@@ -153,12 +153,26 @@ type Config struct {
...
+	//
+	// Note that the IterativeVerifier must be used and the
+	// CompressionVerifiers for the configuration below will be instantiated
+	// to handle the decompression before verification
+	//
+	// Currently supported compression algorithms are:
+	//	1. Snappy (https://google.github.io/snappy/) as "SNAPPY"
+	//
+	// Optional: defaults to empty map/no compression
+	TableColumnCompression map[string]map[string]string
```
fw42: Is this actually used anywhere? I don't see it. Or is that a follow-up PR?

iterative_verifier.go
```diff
@@ -573,6 +614,29 @@ func (v *IterativeVerifier) compareFingerprints(pks []uint64, table *schema.Tabl
 		return nil, targetErr
 	}
 
+	mismatches := compareHashes(sourceHashes, targetHashes)
+	if len(mismatches) <= 0 {
```
fw42: Just curious, can this actually ever be negative? Why not `== 0`?

iterative_verifier.go
```diff
@@ -573,6 +614,29 @@ func (v *IterativeVerifier) compareFingerprints(pks []uint64, table *schema.Tabl
...
+		return mismatches, nil
+	}
+
+	if v.CompressionVerifier != nil && v.CompressionVerifier.IsCompressedTable(table.Name) {
```
fw42: Nitpick: Why not `if len(mismatches) > 0 && ...` here and then you can get rid of line 618?

test/data_iterator_test.go
```diff
@@ -55,62 +55,95 @@ func (this *DataIteratorTestSuite) SetupTest() {
...
 	this.di.Initialize()
 	this.di.AddBatchListener(func(ev *ghostferry.RowBatch) error {
-		this.receivedRows = append(this.receivedRows, ev.Values()...)
+		this.receivedRows[ev.TableSchema().Name] = append(this.receivedRows[ev.TableSchema().Name], ev.Values()...)
 		return nil
 	})
 }
 
 func (this *DataIteratorTestSuite) TestNoEventsForEmptyTable() {
 	_, err := this.Ferry.SourceDB.Query(fmt.Sprintf("DELETE FROM `%s`.`%s`", testhelpers.TestSchemaName, testhelpers.TestTable1Name))
```
fw42: This line needs it's own `this.Require().Nil(err)`, otherwise it's useless

test/iterative_verifier_test.go
```diff
@@ -82,6 +105,30 @@ func (t *IterativeVerifierTestSuite) TestVerifyOncePass() {
...
+	t.InsertCompressedRowInDb(42, testhelpers.TestCompressedData1, t.Ferry.TargetDB)
+
+	result, err := t.verifier.VerifyOnce()
+	t.Require().NotNil(result)
+	t.Require().Nil(err)
+	t.Require().True(result.DataCorrect)
+	t.Require().Equal("", result.Message)
+}
+
+func (t *IterativeVerifierTestSuite) TestVerifyCompressedMismatchOncePass() {
```
fw42: nitpick but `Mismatch` seems a bit ambiguous, how about something like `SameDecompressedDataButDifferentHash` to be more explicit?

test/iterative_verifier_test.go
```diff
@@ -222,6 +297,25 @@ func (t *IterativeVerifierTestSuite) InsertRowInDb(id int, data string, db *sql.
 	t.Require().Nil(err)
 }
 
+func (t *IterativeVerifierTestSuite) InsertCompressedRowInDb(id int, data string, db *sql.DB) {
+	t.SetColumnType(testhelpers.TestSchemaName, testhelpers.TestCompressedTable1Name, testhelpers.TestCompressedColumn1Name, "MEDIUMBLOB", db)
```
fw42: This seems weird. Why do we change the schema during the test (rather than to just always have it be a MEDIUMBLOB and set it during test setup or whenever we create the db)?

testhelpers/unit_test_suite.go
```diff
@@ -29,10 +30,39 @@ func SetupTest() {
...
+)
+
+func init() {
+	filePaths := []string{
+		FixturePath("urls1.snappy"),
+		FixturePath("urls2.snappy"),
+	}
+
+	decompressed := [][]byte{}
+	for _, path := range filePaths {
```
fw42: Nitpick, but this whole loop thing seems a bit unnecessary to me.. I'd create some kind of `LoadFixtureFromFile` helper function and then just do
```golang
func init() {
  TestCompressedData3 = LoadFixtureFromFile("urls1.snappy")
  TestCompressedData4 = LoadFixtureFromFile("urls2.snappy")
}
```

compression_verifier.go
fjordan: We don't actually know the number of rows until iterating through all of them. That would be more efficient though if we did.

testhelpers/unit_test_suite.go
fjordan: Will update. I like this more 👍

test/iterative_verifier_test.go
fjordan: We change to `MEDIUMBLOB` because of the snappy-compressed data. `TEXT` expects utf8 data and throws an error when we try to insert the compressed data. I didn't want to change any other tests but the ones added in this PR. If we know later it won't negatively impact the other tests then I'm sure we can change it and just use `MEDIUMBLOB` everywhere.

config.go
fjordan: yep @hkdsun will be following up with a PR to integrate into sharding and use this

PR #45, opened 2018-07-23
Updated siddontang/mysql and switched from Glide -> Dep

### LINKED ISSUE #23: Binlog replication uses timestamp values from the machine-local timezone as oppose to UTC
I Test Copying , Found  This  Problem ,
Table Cloum Type  Is:
```
| sf_modify_time | timestamp           | NO   |     | CURRENT_TIMESTAMP | on update CURRENT_TIMESTAMP |
````
I'll try to reproduce this and record it here.

### PR DESCRIPTION
**Pull request by shuhaowu**
As far as I can tell, Glide is deprecated and Dep is adopted as the official package management tool for Golang.

However, this switch is made by a practical reason: I want to update siddontang/mysql separately from all the other dependencies for now to prepare to fix #23 

Given that the vendor directories makes up the bulk of the commit here, the commits to actually look at are:

- b34dc02: actual change from glide.yml to Gopkg.toml.
- 1158d20: updated the siddontang/mysql dependency.

I also reviewed siddontang/mysql's changes and a summary of changes relevant to us is presented below:

- https://github.com/siddontang/go-mysql/commit/eecba186bbef1be3171588cd76a40ef0b68d53d9
  - Added MaxReconnectAttempts to BinlogSyncer, which allows for retries to occur when the binlogsyncer fails to sync. We might be able to use this in our code somehow.
- https://github.com/siddontang/go-mysql/commit/919af6c46b676b3c5ea0856eafbff42660a6b0ce
  - Canal detects schema changes in the binlog streamer. We might be able to draw some inference from this, although it's more questionable.
- https://github.com/siddontang/go-mysql/commit/2b6ddb04ded1419f74aa5e26d8f8adc2fec1a5a1
  - Tangentially related: This adds a method to dump all remaining event in the binlog streamer. Don't think it will be useful as we stop at a target position.
- https://github.com/siddontang/go-mysql/commit/a972a0f0c62a4552f0674d035f8642b3c6e1ce22
  - Switched from logrus to a different logger. This might cause issues with logging/error reporting?
- https://github.com/siddontang/go-mysql/commit/e59113b0edf2d547ec35e85213ad6ccde1848de9
  - Bug fix of parsing 0000-00-00 00:00:00 in replication events. Previously it would have converted that value to something like 0001-XX-XX XX:XX:XX.
- https://github.com/siddontang/go-mysql/commit/29d34f19266fe38678ecfcf7ad4449c33b2bd26d
  - Added option to checksum binlogs via the binlog_checksum mysql entry.

### CODE REVIEW

**Comment by hkdsun** [comment]
> Switched from logrus to a different logger. This might cause issues with logging/error reporting?

https://github.com/birkirb/loggers-mapper-logrus/ seems to suggest a way of plugging logrus back into the logger interface. though, i dont know how much value we were previously getting from the  logs produced by siddontang. 

Here's some log snippets of how it looks now from the tests – you can see there's some inconsistency now:

```
time="2018-07-23T16:03:14Z" level=info msg="connecting to database" dbname=target dsn="root:<masked>@tcp(127.0.0.1:29292)/?collation=utf8mb4_unicode_ci&multiStatements=true&charset=utf8mb4&time_zone=%27%2B00%3A00%27&sql_mode=%27STRICT_ALL_TABLES%2CNO_BACKSLASH_ESCAPES%27" tag=ferry
time="2018-07-23T16:03:14Z" level=info msg="connected to target" dbname=target hasSSL=false ssl_cipher= tag=ferry
time="2018-07-23T16:03:14Z" level=info msg="ferry initialized" tag=ferry
2018/07/23 16:03:15 INFO  create BinlogSyncer with config {91919  127.0.0.1 29291 root    false false <nil> false true 0 0s 0s 0}
time="2018-07-23T16:03:15Z" level=info msg="reading current binlog position" tag=binlog_streamer
time="2018-07-23T16:03:15Z" level=info msg="found binlog position, starting synchronization" file=mysql-bin.000003 pos=624493 tag=binlog_streamer
2018/07/23 16:03:15 INFO  begin to sync binlog from position (mysql-bin.000003, 624493)
2018/07/23 16:03:15 INFO  register slave for master server 127.0.0.1:29291
2018/07/23 16:03:15 INFO  rotate to (mysql-bin.000003, 624493)
time="2018-07-23T16:03:15Z" level=debug msg="loading tables from database" database=gftest tag=table_schema_cache
time="2018-07-23T16:03:15Z" level=debug msg="fetching table schema" database=gftest table=table1 tag=table_schema_cache
time="2018-07-23T16:03:15Z" level=debug msg="caching table schema" database=gftest table=table1 tag=table_schema_cache
time="2018-07-23T16:03:15Z" level=info msg="table schemas cached" tables="[gftest.table1]" tag=table_schema_cache
time="2018-07-23T16:03:15Z" level=info msg="starting ferry run" tag=ferry
time="2018-07-23T16:03:15Z" level=info msg="starting data iterator run" tablesCount=1 tag=data_iterator
.... [all logrus logs]
time="2018-07-23T16:03:17Z" level=info msg="requesting binlog streamer to stop" tag=binlog_streamer
time="2018-07-23T16:03:17Z" level=info msg="current stop binlog position was recorded" tag=binlog_streamer target_position="(mysql-bin.000003, 819002)"
time="2018-07-23T16:03:17Z" level=debug msg="received event *ghostferry.BinlogUpdateEvent at 2018-07-23 16:03:17 +0000 UTC" database=gftest table=table1 tag=binlog_streamer
time="2018-07-23T16:03:17Z" level=info msg="exiting binlog streamer" tag=binlog_streamer
2018/07/23 16:03:17 INFO  syncer is closing...
2018/07/23 16:03:17 ERROR  connection was bad
2018/07/23 16:03:17 ERROR close sync with err: Sync was closed
2018/07/23 16:03:17 INFO  syncer is closed
time="2018-07-23T16:03:17Z" level=info msg="starting verification during cutover" tag=iterative_verifier
time="2018-07-23T16:03:17Z" level=debug msg=reverifying batches=2 tag=iterative_verifier
time="2018-07-23T16:03:17Z" level=debug msg="received pk batch to reverify" len(pks)=200 table=gftest.table1 tag=iterative_verifier
time="2018-07-23T16:03:17Z" level=debug msg="received pk batch to reverify" len(pks)=64 table=gftest.table1 tag=iterative_verifier
time="2018-07-23T16:03:17Z" level=error msg="failed reverification: verification failed on table: gftest.table1 for pks: 42" tag=iterative_verifier
time="2018-07-23T16:03:17Z" level=error msg="failed reverification: verification failed on table: gftest.table1 for pks: 43" tag=iterative_verifier
time="2018-07-23T16:03:17Z" level=info msg="cutover verification complete" tag=iterative_verifier
--- PASS: TestVerificationFailsUpdatedRow (2.89s)
```


Maybe we can log an issue for this and move on

File: glide.yaml [hunk 1]
```diff
@@ -1,23 +0,0 @@
-package: github.com/Shopify/ghostferry
-import:
-- package: github.com/Masterminds/squirrel
```
hkdsun: looks like we're missing squirrel in Gopkg.toml.. not sure how tests are passing, maybe because we have an indirect dependency?

Gopkg.toml
```diff
@@ -0,0 +1,54 @@
...
+
+[[constraint]]
+  name = "github.com/sirupsen/logrus"
+  version = "1.0.0"
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
+  version = "1.1.4"
+
+[prune]
```
hkdsun: I'm getting a warning about this for some reason:

```
hormoz@RMT-HormozKheradmand:[~/d/go/src/github.com/shuhaowu/ghostferry] (1)
 -> dep ensure                                                                                                                         11:32:29 (glide-to-dep)
👩‍💻  activated go:1.9
dep: WARNING: Unknown field in manifest: prune
```

hkdsun: had to update `dep` for this. kcco

shuhaowu:
> `met?: glide install`

Is this even required at all? I was under the impression since we have everything in vendor/, go build/test/etc would just work. Dep/glide is only needed to update the dependencies.

When I setup a new Ghostferry dev environment, I've never ran that command (or use dev to do so).

hkdsun:
It's not strictly required as you mention but it's a workflow familiar to Shopify developers.

For example, in a Rails project our developers are used to editing a Gemfile and then relying on `dev up` to do the right thing for updating dependencies.

Keeping the dependency update step in the dev config will keep our project familiar to our internal developers.

The change required is simply `s/glide install/dep ensure`

shuhaowu:
Oh what I'm saying is that dep ensure/glide install should always be a no-op as the vendored dependencies should always be the correct versions (assuming that one used dep ensure -add to add dependencies and etc). Thus we don't really need the step. 

However, since having it doesn't hurt due to it being a no-op, I'll change the file instead of removing that section.

hkdsun:
Yeah what I was trying to say is that a lot of our developers default to using `dev up` instead of `dep ensure`/`bundle install` to update dependencies (after they just edited the dependency config file)

shuhaowu:
Added squirrel into Gopkg.toml as well as fixed the dev.yml file. The squirrel addition caused it to update to the latest commit, which doesn't seem to change much compared to before:

https://github.com/Masterminds/squirrel/compare/2a85d1d526545de922a487d4841e61d6ee690c35...b127ed9be03443fe3c0877e391130e3dd3f3107a

Also dep updated to 0.5, which has a slightly different .lock format.

glide.yaml
```diff
@@ -1,23 +0,0 @@
...
-- package: github.com/sirupsen/logrus
-  version: ^1.0.0
-- package: github.com/siddontang/go-mysql
```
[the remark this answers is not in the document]
shuhaowu: Tests are passing likely because the dependency is in the lock file. This has been fixed along with any other missing dependencies.

shuhaowu:
I just reinitialized the Gopkg.toml file from the code base as opposed to using the glide migration tool. This should fix all the dependency issues. I noticed that we have a bunch of outdated dependencies right now. After merging this PR, I'll take a look at upgrading them to ensure we can follow upstream to a reasonable degree.

shuhaowu:
I updated go-mysql again as there has been a few changes on master. Notably, there's siddontang/go-mysql@29d34f1, which allows the binlog streamer to checksum the binlog events from the mysql master via the MySQL binlog_checksum config for corruption. This could be a good add to the code base given that it's effectively free and we do this with regular MySQL replication already in production.

I'm also going to merge this for now so I can get unblocked for now. I checked through the yaml/toml files and dev/package management is _correct_ this time around. If we have any other concerns, we can address this in a follow up PR.

Issue #47, opened 2018-07-30
ShardingErrorHandler can be merged with PanicErrorHandler


shuhaowu:
All ShardingErrorHandler does is send a webhook. We can definitely merge this into the default but optional behaviour of core Ghostferry.

Also should make it so the error callback can be called independently of panic, so it can be invoked during the Initialize/Start phase of Ghostferry.

PR #48, opened 2018-08-01
Check if target database is writable before starting ferry

hkdsun:
This allows a ferry to fail early in case the provided target database connection is not writeable. We previously had this check in our application but it makes more sense to perform this check at the Ghostferry level.

@Shopify/pods @shuhaowu

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -134,6 +134,15 @@ func (f *Ferry) Initialize() (err error) {
 		return err
 	}
 
+	isReplica, err := f.checkDbIsAReplica(f.TargetDB)
+	if err != nil {
+		f.logger.WithError(err).Error("cannot check if target db is writable")
+		return err
+	}
+	if isReplica {
+		return fmt.Errorf("@@read_only must be OFF on target db")
+	}
+
 	if f.WaitUntilReplicaIsCaughtUpToMaster != nil {
 		f.WaitUntilReplicaIsCaughtUpToMaster.ReplicaDB = f.SourceDB
 
@@ -432,6 +441,10 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
 
 	row = db.QueryRow("SHOW VARIABLES LIKE 'binlog_row_image'")
 	err = row.Scan(&name, &value)
+	if err != nil {
+		return err
+	}
+
 	if strings.ToUpper(value) != "FULL" {
 		return fmt.Errorf("binlog_row_image must be FULL, not %s", value)
 	}
--- a/sharding/sharding.go
+++ b/sharding/sharding.go
@@ -86,7 +86,12 @@ func (r *ShardingFerry) Initialize() error {
 		}
 	}
 
-	return r.Ferry.Initialize()
+	err := r.Ferry.Initialize()
+	if err != nil {
+		r.Ferry.ErrorHandler.Fatal("ferry.initialize", err)
+		return err
+	}
+	return nil
 }
 
 func (r *ShardingFerry) newIterativeVerifier() (*ghostferry.IterativeVerifier, error) {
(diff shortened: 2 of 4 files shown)


fw42:
We also need to check the source, not just the target. And not just at the beginning of the move but also before entering the cutover (that's currently our only safety net for "was the pod failed over during the shop move?").

hkdsun:
To be honest, it doesn't really make sense to check the source db in Ghostferry. We should solve that differently

> And not just at the beginning of the move but also before entering the cutover

In the case of the target, we'll probably fail the ferry due to write errors if the target becomes read-only at any point during the ferry run

sharding/sharding.go
```diff
@@ -78,7 +78,12 @@ func NewFerry(config *Config) (*ShardingFerry, error) {
 }
 
 func (r *ShardingFerry) Initialize() error {
-	return r.Ferry.Initialize()
+	err := r.Ferry.Initialize()
+	if err != nil {
+		r.Ferry.ErrorHandler.Fatal("ferry.initialize", err)
```
shuhaowu: Is it necessary to call the error handler and panic here? Until now, I believe we only use the error handler during the Run() phase of ghostferry, where it is not possible to directly return an error.

If the idea is to give an error callback, we expose that API and/or hook up the logrus error reporting as a part of #47.

ferry.go
```diff
@@ -383,9 +389,29 @@ func checkConnectionForBinlogFormat(db *sql.DB) error {
...
+	}
+
 	if strings.ToUpper(value) != "FULL" {
 		return fmt.Errorf("binlog_row_image must be FULL, not %s", value)
 	}
 
 	return nil
 }
+
+func checkConnectionIsWritable(db *sql.DB) error {
```
shuhaowu: As a part of #35, I made all these methods under the scope of Ferry as they are (1) not used anywhere else and also are (2)  private methods such that nothing external can use it. Can this be made consistent as well?

hkdsun:
> To be honest, it doesn't really make sense to check the source db in Ghostferry. We should solve that differently

Thinking about this more and talking about it with @pushrax, maybe it does make sense. 

Currently, we run our ferries from readonly replica and we require the user to provide the replica's master. Before we `WaitUntilReplicaIsCaughtUpToMaster`, it makes sense to ensure it's the active writer, which is exactly what we want to ensure.

sharding/sharding.go
hkdsun: The idea is exactly giving our application the callback. We still don't have #47 solved, so we'll have to live with this for a bit

test/ferry_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	ferry *ghostferry.Ferry
+}
+
+func (t *FerryTestSuite) SetupTest() {
+	t.GhostferryUnitTestSuite.SetupTest()
+}
+
+func (t *FerryTestSuite) TestReadOnlyDatabaseFailsInitialization() {
+	_, err := t.Ferry.TargetDB.Exec("SET GLOBAL read_only = ON")
```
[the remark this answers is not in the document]
shuhaowu: As a note, this already happens elsewhere in tests quite frequently.

test/ferry_test.go
```diff
@@ -0,0 +1,39 @@
...
+	t.Require().Nil(err)
+
+	ferry := testhelpers.NewTestFerry().Ferry // make new ferry that re-uses the same targetDB as t.Ferry
```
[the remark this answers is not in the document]
hkdsun: just so we're not re-`Initialize()`ing an already initialized ferry

test/ferry_test.go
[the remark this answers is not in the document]
hkdsun: added a teardown

ferry.go
hkdsun: sure, done

PR #50, opened 2018-08-02
TIMESTAMP Timezone fix

### LINKED ISSUE #23: Binlog replication uses timestamp values from the machine-local timezone as oppose to UTC

shuhaowu:
https://github.com/siddontang/go-mysql/pull/295 has been merged and thus we have a way to fix the timezone issue causing binlogs to not properly sync over. 

However, the deeper question now is whether or not setting this to UTC will solve the issue. Should we also try to set the mysql session timezone to UTC while executing queries on the target and the source? Or maybe we have to do something else as well?

Fixes #23

### DIFF
--- a/test/types_integration_test.go
+++ b/test/types_integration_test.go
@@ -20,7 +20,7 @@ func addTypesToTable(db *sql.DB, dbName, tableName string) {
 		"ADD date_col DATE," +
 		"ADD time_col TIME," +
 		"ADD dt_col DATETIME," +
-		"ADD ts_col TIMESTAMP," + // TODO broken on master
+		"ADD ts_col TIMESTAMP," +
 		"ADD varchar_col VARCHAR(128)," +
 		"ADD enum_col ENUM('foo', 'bar')," +
 		"ADD set_col SET('foo', 'bar', 'baz')," +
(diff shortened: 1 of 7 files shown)


test/types_integration_test.go
```diff
@@ -20,7 +20,7 @@ func addTypesToTable(db *sql.DB, dbName, tableName string) {
 		"ADD date_col DATE," +
 		"ADD time_col TIME," +
 		"ADD dt_col DATETIME," +
-		"ADD ts_col TIMESTAMP," + // TODO broken on master
```
hkdsun: What was this TODO for? Can we add a test for the broken behaviour? I think that was Justin's intention?

shuhaowu: This was the test that is broken due to this bug if you have a non UTC timezone on your computer/testing setup.

hkdsun: Which test? you're just removing a comment here. What am I missing?

hkdsun: In other words: the `setupMultiTypeTable` method below is not inserting anything into this column because doing so would have caused the failure you're trying to fix here. But now that we don't have the bug, we should be able to re-insert the value

shuhaowu: Huh interesting, I recall when debugging distinctly seeing a TZ issue. The test case `TestCopyDataWithManyTypes` also fails without this patch (without modification to setupMultiTypeTable) on my machine with EDT set. 

I'll check this more carefully next week.

pushrax: Need to add the column here https://github.com/Shopify/ghostferry/blob/a37ec446d7b065760d52819903c4b808c8c9e3f3/test/types_integration_test.go#L53

shuhaowu: So it appears this test was ACTUALLY broken in master (well as it never passed on my machine) because even though we don't INSERT the ts_col explicitly, it's done implicitly. If we run SHOW CREATE TABLE table1, we can see:

```
  `ts_col` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```

I'll add the ts_col into that query regardless, but that explains why I see the failed test despite never specifying that column.

shuhaowu:
> The already set the connection to UTC (specifically +00:00). The issue was entirely a parsing one afaik.

Cool. So the issue was essentially that go-mysql is using `time.Unix` to convert the timestamp from the BinlogStreamer into a local time struct, which when stringed during AsSQLQuery gives us a wrong value.

This then is a complete fix given that when interacting with MySQL, we're using UTC everywhere.

I'll merge this once the test passes on CI.

PR #51, opened 2018-08-02
Abort ferry if master db becomes read-only

hkdsun:
Currently, we allow running our ferries from readonly replica and we require the user to provide the replica's master connection details. Before we `WaitUntilReplicaIsCaughtUpToMaster`, it makes sense to ensure the connection we have is the currently active writer and avoid data loss.

In our current implementation, you can have a scenario where the master writer has been failed over to another db and our perceived master is lagging behind by a few minutes (compared to the actual master) leading to data loss because we don't wait long enough to catch up to the _actual_ master but rather to another replica.

More context in #48

@Shopify/pods @shuhaowu

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -134,6 +134,15 @@ func (f *Ferry) Initialize() (err error) {
 		return err
 	}
 
+	isReplica, err := f.checkDbIsAReplica(f.TargetDB)
+	if err != nil {
+		f.logger.WithError(err).Error("cannot check if target db is writable")
+		return err
+	}
+	if isReplica {
+		return fmt.Errorf("@@read_only must be OFF on target db")
+	}
+
 	if f.WaitUntilReplicaIsCaughtUpToMaster != nil {
 		f.WaitUntilReplicaIsCaughtUpToMaster.ReplicaDB = f.SourceDB
 
@@ -432,6 +441,10 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
 
 	row = db.QueryRow("SHOW VARIABLES LIKE 'binlog_row_image'")
 	err = row.Scan(&name, &value)
+	if err != nil {
+		return err
+	}
+
 	if strings.ToUpper(value) != "FULL" {
 		return fmt.Errorf("binlog_row_image must be FULL, not %s", value)
 	}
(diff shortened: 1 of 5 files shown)


ferry.go
```diff
@@ -380,7 +380,22 @@ func (f *Ferry) WaitUntilBinlogStreamerCatchesUp() {
...
-		err := f.WaitUntilReplicaIsCaughtUpToMaster.Wait()
+		isReplica, err := f.checkDbIsAReplica(f.WaitUntilReplicaIsCaughtUpToMaster.MasterDB)
+		if err != nil {
+			f.ErrorHandler.Fatal("wait_replica", err)
+		}
+
+		if isReplica {
+			err = errors.New("source master is no longer a master writer")
+			msg := "aborting the ferry since the source master is no longer the master writer " +
+				"this means that the perceived master can be lagged compared to the actual master " +
```
fw42: `s/lagged/lagging/`?

ferry.go
```diff
@@ -380,7 +380,22 @@ func (f *Ferry) WaitUntilBinlogStreamerCatchesUp() {
...
+				"and lead to missed writes. aborting ferry"
+
+			f.logger.WithError(err).Error(msg)
```
fw42: Doesn't `ErrorHandler.Fatal` already log the error? wouldn't this log here cause it to get logged twice?

ferry.go
```diff
@@ -380,7 +380,22 @@ func (f *Ferry) WaitUntilBinlogStreamerCatchesUp() {
 // You will know that the BinlogStreamer finished when .Run() returns.
 func (f *Ferry) FlushBinlogAndStopStreaming() {
```
fw42: I assume this function is called only once (before the cutover)?

hkdsun: yes 👍

ferry.go
```diff
@@ -134,6 +134,15 @@ func (f *Ferry) Initialize() (err error) {
 		return err
 	}
 
+	isReplica, err := f.checkDbIsAReplica(f.TargetDB)
```
shuhaowu: All this method does is check if the DB is READ_ONLY, right? Is this a reliable measure of whether or not the DB is a replica? Should we look at something like SLAVE STATUS?

hkdsun: SLAVE STATUS is actually not a great way. For example, Shopify runs a master-master configuration which means masters from different location are slaving off each other (and hence SLAVE STATUS will show the slave thread on both instances)

shuhaowu: Not for now, but perhaps it is a good idea to defer this task to WaitUntilReplica struct such that people can implement arbitrary ways to check this.

hkdsun: yeah that seems desirable for non-shopify users. I'll open an issue and we'll take care of it when there's interest

PR #56, opened 2018-08-08
Merge ShardingErrorHandler with PanicErrorHandler

### LINKED ISSUE #47: ShardingErrorHandler can be merged with PanicErrorHandler

bbuchalter:
Closes #47.

* HTTPCallback moved from sharding to ghostferry
* HTTPCallback is optional in PanicErrorHandler
* HTTPCallback will only be called once in PanicErrorHandler (was called multiple times if multiple errors occurred simultaneously)

Regarding test coverage changes, ShardingErrorHandler embeds a test ErrorHandler to avoid a panic in tests.
PanicErrorHandler does not allow embedding of additional error handlers, therefore
if invoked in a test, a panic would occur.

We *want* a panic to occur, as that is the intended behavior of PanicErrorHandler,
so we allow the panic in the test, catch it with a defer, and add assertions.

For all other callback tests, we use the test error handler, as before.

We merged the assertions about the HTTP part and the panic part of the callback into a single
test to avoid catching panics in more than one test.

# TODO
- [x] Remove `return` after `ErrorHandler.Fatal` calls, because `ErrorHandler` is no longer a goroutine
- [x] Create an interface to be able to invoke the HTTPCallback, without the panic. This is to [avoid `ErrorHandler.Fatal` in the initialize phase](https://github.com/Shopify/ghostferry/blob/master/sharding/sharding.go#L91)


shuhaowu:
The TODOs on this issue is not super important. I can take care of them after merging this PR, which would be nice for the interruption work.

Normally I would review a PR like this and can merge it as it is relatively straightforward. However, I helped write this PR in a pair-programming session and it does touch some ghostferry-sharding tests in perhaps non-intuitive ways, so I would like to request at least one more review on this.

sharding/test/callbacks_test.go
```diff
@@ -71,12 +66,34 @@ func (t *CallbacksTestSuite) TestFailsRunOnPanicError() {
...
+	defer func() {
+		if r := recover(); r != nil {
+			panicReceived = true
+			t.Require().Equal(r, "fatal error detected, see logs for details")
+		}
+	}()
 	t.Ferry.Ferry.ErrorHandler.Fatal("test_error", errors.New("test error"))
 
 	t.Require().True(callbackReceived)
+	t.Require().True(panicReceived)
```
hkdsun: I don't think any of lines 93+ are executed. Try this example (https://play.golang.org/p/SPKXST8TKvw):

```
package main

import "fmt"

func main() {
    defer func() {
        if r := recover(); r != nil {
            fmt.Println("Recovered in main. Panic was:", r)
        }
    }()

    fmt.Println("Calling f.")

    f()

    fmt.Println("I'M NEVER EXECUTED!!!!")
}

func f() {
    panic("test")   
}
```

shuhaowu:
Two points to address: (1) why are we merging the sharding and panic error handler and (2) how can we avoid a panic during the Initialize phase of Ghostferry.

For (1):

Right now the code base has two error handler: one of which only logs to stderr and the other reports to HTTP in addition to the previous. Since both of these are fairly standard operations (assuming that most people who will write a custom Ghostferry application will likely have HTTP based reporting), why not just have both of them available in the standard error handler instead of having two structs, where one of the structs embeds the other one. The two struct with its inheritance feels more complicated than it needs to be.

For (2):

Really, the ErrorHandler is just an interface that the API consumer has to implement in order for Ghostferry to _report_ an unrecoverable error. The ErrorHandler implementation should also deal with that error in some sane way. So far, `panic` and serialize the state (albeit not yet working), is all we do. 

Right now, both of these behaviour is hidden under the same API: `.Fatal()`. Since we want to _report_ some errors during the Initialize/Start phase of Ghostferry, we are also using this method as there's no alternative. What I'm proposing is we implement a `.ReportError()` method, which only reports the error to stderr and possibly a HTTP. We then implement `.Fatal()` to call `.ReportError()`, before dumping the state and panicking. This PR hasn't touched on this part, but it should be just moving some code around.

sharding/test/callbacks_test.go
shuhaowu: True. I think I can fix this test if I implement the `ReportError` mechanism.

shuhaowu:
I added the ReportError API to ErrorHandler, which allows us to report errors without calling to `panic`, which addresses the issue where we were panicking during the Initialize phase of sharding/Ferry.

binlog_streamer.go
```diff
@@ -128,7 +128,6 @@ func (s *BinlogStreamer) Run() {
 
 		if err != nil {
 			s.ErrorHandler.Fatal("binlog_streamer", err)
-			return
```
hkdsun: Why remove these returns? I know we'll panic anyway but it seems more correct to return anyway

shuhaowu: In the beginning, ErrorHandler was a goroutine and Fatal send the error down a channel. Since be4f15da531200c8cfa4b1f2d4686469e9a0d2ee, we no longer have a goroutine to panic. panic should occur immediately and block the goroutine from executing the next line anyway, making the return unnecessary.

We already stopped doing return for most of these calls. It's just a couple of the really old line that didn't get cleaned up.

PR #57, opened 2018-08-09
Integration of CompressionVerifier into sharding package

hkdsun:
This integrates the work done in #44 into the sharding package for use in the shop mover.

I also realized that after performing delta-copy of joined tables we never verify the copied rows.

@Shopify/pods

### DIFF
--- a/sharding/sharding.go
+++ b/sharding/sharding.go
@@ -109,6 +109,15 @@ func (r *ShardingFerry) newIterativeVerifier() (*ghostferry.IterativeVerifier, e
 		}
 	}
 
+	var compressionVerifier *ghostferry.CompressionVerifier
+	if r.config.TableColumnCompression != nil {
+		var err error
+		compressionVerifier, err = ghostferry.NewCompressionVerifier(r.config.TableColumnCompression)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return &ghostferry.IterativeVerifier{
 		CursorConfig: &ghostferry.CursorConfig{
 			DB:          r.Ferry.SourceDB,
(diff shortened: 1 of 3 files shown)


sharding/sharding.go
```diff
@@ -189,12 +199,14 @@ func (r *ShardingFerry) Run() {
 	r.Ferry.FlushBinlogAndStopStreaming()
 	copyWG.Wait()
 
+	// Joined tables cannot be easily monitored for changes in the binlog stream
```
fjordan: Curious as to why joined tables create this problem. Is there a related issue?

sharding/sharding.go
```diff
@@ -189,12 +199,14 @@ func (r *ShardingFerry) Run() {
...
+	// so we copy and verify them once more during the cutover phase
```
fw42: This wording is a bit confusing. the "once more" is about "verify", not "copy", right? we don't actually copy them twice, but that's what this makes it sound like.

sharding/sharding.go
```diff
@@ -240,7 +252,31 @@ func (r *ShardingFerry) deltaCopyJoinedTables() error {
...
+
+	err = verifier.Initialize()
+	if err != nil {
+		return err
+	}
+
+	verificationResult, err := verifier.VerifyOnce()
+	if err != nil {
+		return err
+	} else if !verificationResult.DataCorrect {
```
fw42: nitpick: Can you make this just an `if` instead of an `else if` (because of the `return`, that's equivalent here, but I find it a tiny bit easier to understand).

sharding/sharding.go
hkdsun: Joined tables usually don't have the sharding key stored on them (because a single row can be shared by multiple tenants), hence discovering applicable (i.e. belonging to the tenant) DML events produced by these tables is impossible. 

Note: you could technically do the join every time you see a DML event and _find out_ if it's applicable but this could really slow down the replication process for Ghostferry

sharding/sharding.go
hkdsun: We do actually. Before the cutover phase, the DataIterator batch copies the table. The "delta-copy" here ensures that we've copied over any new rows that have appeared since.

Issue #53, opened 2018-08-03
Provide a generic way of checking whether a database is the active writer


hkdsun:
Currently, we rely on the `@@read_only` variable to check whether our connection to the database is indeed to an active writer or if it's actually a read replica. This is done specifically before we use the `WaitUntilReplicaIsCaughtUpToMaster` struct to enter the cutover phase.

Perhaps not everybody's failover strategy is compatible with this behaviour. We could provide a configuration option on the `WaitUntilReplicaIsCaughtUpToMaster` struct so that the user can provide us with a query that checks this condition.

PR #58, opened 2018-08-10
sharding: abort if target db becomes readonly before cutover

hkdsun:
Follow up to https://github.com/Shopify/ghostferry/pull/51

The rationale for this one is that it's too risky to go into the cutover phase if the target db is read-only anyway. It replaces a call that we previously did on the shop mover side.

@Shopify/pods

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -169,3 +169,10 @@ func NewMysqlPosition(file string, position uint32, err error) (mysql.Position, 
 		}, nil
 	}
 }
+
+func CheckDbIsAReplica(db *sql.DB) (bool, error) {
+	row := db.QueryRow("SELECT @@read_only")
+	var isReadOnly bool
+	err := row.Scan(&isReadOnly)
+	return isReadOnly, err
+}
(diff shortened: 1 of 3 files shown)


utils.go
```diff
@@ -169,3 +169,10 @@ func NewMysqlPosition(file string, position uint32, err error) (mysql.Position,
 		}, nil
 	}
 }
+
+func CheckDbIsAReplica(db *sql.DB) (bool, error) {
+	row := db.QueryRow("SELECT @@read_only")
```
fjordan: Have we thought about using `SHOW SLAVE STATUS` here instead of just relying on the `@@read_only` variable? It seems it is technically possible to run an async slave without setting this flag as MySQL doesn't force you to set `read_only = ON`. What do you think?

hkdsun: Shopify runs a master-master configuration which means masters from different location are slaving off each other (and hence SLAVE STATUS will show the slave thread on both instances). We're tracking making this configurable in https://github.com/Shopify/ghostferry/issues/53

PR #60, opened 2018-08-13
Refactored DataIteratorState to StateTracker for interruptibility

shuhaowu:
*The code review might look big, but most of it is me moving the DataIteratorState into its own file called StateTracker.*

Instead of having the DataIterator own the state of the Ferry run, Ferry should own the state as it involves both the DataIterator and the BinlogWriter. This means effectively, the DataIterator and the BinlogWriter "notifies" the StateTracker about their progress. 

This PR moves DataIteratorState to a StateTracker managed by the Ferry. However, this centralization does make using the DataIterator/BinlogStreamer independently of the Ferry slightly more difficult. I think this trade off is okay as it makes the state tracking easy to reason about (it's all in one component!). If we need to, we can refactor a little bit to provide some helper methods to use those underlying components.
 
The current StateTracker does not track the verifier progress. This will be done in the future in a similar way. We also can't resume with the dump state. That code will come after the initial effort for interrupt/resume during the normal Ghostferry run.

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -35,7 +35,7 @@ type BinlogStreamer struct {
 	eventListeners []func([]DMLEvent) error
 }
 
-func (s *BinlogStreamer) Initialize() (err error) {
+func (s *BinlogStreamer) Initialize() error {
 	s.logger = logrus.WithField("tag", "binlog_streamer")
 	s.stopRequested = false
 	return nil
--- a/binlog_writer.go
+++ b/binlog_writer.go
@@ -2,6 +2,7 @@ package ghostferry
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 
 	"github.com/siddontang/go-mysql/schema"
@@ -18,12 +19,17 @@ type BinlogWriter struct {
 	WriteRetries int
 
 	ErrorHandler ErrorHandler
+	StateTracker *StateTracker
 
 	binlogEventBuffer chan DMLEvent
 	logger            *logrus.Entry
 }
 
 func (b *BinlogWriter) Initialize() error {
+	if b.StateTracker == nil {
+		return errors.New("must specify StateTracker")
+	}
+
 	b.logger = logrus.WithField("tag", "binlog_writer")
 	b.binlogEventBuffer = make(chan DMLEvent, b.BatchSize)
 	return nil
@@ -110,7 +116,7 @@ func (b *BinlogWriter) writeEvents(events []DMLEvent) error {
 		return fmt.Errorf("exec query (%d bytes): %v", len(query), err)
 	}
 
-	// lastWrittenBinlogPosition = events[len(events)-1].BinlogPosition()
+	b.StateTracker.UpdateLastWrittenBinlogPosition(events[len(events)-1].BinlogPosition())
 
 	return nil
 }
--- a/data_iterator.go
+++ b/data_iterator.go
@@ -91,26 +91,6 @@ func (d *DataIterator) Run() {
 						}
 					}
 
-					// The way we save the LastSuccessfulPK is probably incorrect if we
-					// want to ensure that when we crash, we have a "correct" view of
-					// the LastSuccessfulPK.
-					// However, it's uncertain if it is even theoretically possible to
-					// save the "correct" value.
-					// TODO: investigate this if we want to ensure that on error, we have
-					//       the "correct" last successful PK and other values.
-					// TODO: it is also perhaps possible to save the Cursor objects
-					// directly as opposed to saving a state, but that is left to
-					// the future.
-					lastRow := batch.Values()[len(batch.Values())-1]
-					pkpos, err := lastRow.GetUint64(batch.PkIndex())
-					if err != nil {
-						logger.WithError(err).Error("failed to convert pk to uint64")
-						return err
-					}
-
-					logger.Debugf("updated last successful PK to %d", pkpos)
-					d.StateTracker.UpdateLastSuccessfulPK(table.String(), pkpos)
-
 					return nil
 				})
 
--- a/data_iterator.go
+++ b/data_iterator.go
@@ -1,142 +1,14 @@
 package ghostferry
 
 import (
-	"container/ring"
 	"database/sql"
+	"errors"
+	"fmt"
 	"sync"
-	"time"
 
 	"github.com/siddontang/go-mysql/schema"
 	"github.com/sirupsen/logrus"
 )
-
-type PKPositionLog struct {
-	Position uint64
-	At       time.Time
-}
-
-type DataIteratorState struct {
-	// We need to lock these maps as Go does not support concurrent access to
-	// maps.
-	// Maybe able to convert to sync.Map in Go1.9
-	targetPrimaryKeys         map[string]uint64
-	lastSuccessfulPrimaryKeys map[string]uint64
-	completedTables           map[string]bool
-	copySpeedLog              *ring.Ring
-
-	targetPkMutex     *sync.RWMutex
-	successfulPkMutex *sync.RWMutex
-	tablesMutex       *sync.RWMutex
-}
-
-func newDataIteratorState(tableIteratorCount int) *DataIteratorState {
-	// We want to make sure that the ring buffer is not filled with
-	// only the timestamp from the last iteration.
-	//
-	// Having it some multiple times of the number of table iterators
-	// _should_ allow for this.
-	speedLog := ring.New(tableIteratorCount * 5)
-	speedLog.Value = PKPositionLog{
-		Position: 0,
-		At:       time.Now(),
-	}
-
-	return &DataIteratorState{
-		targetPrimaryKeys:         make(map[string]uint64),
-		lastSuccessfulPrimaryKeys: make(map[string]uint64),
-		completedTables:           make(map[string]bool),
-		copySpeedLog:              speedLog,
-		targetPkMutex:             &sync.RWMutex{},
-		successfulPkMutex:         &sync.RWMutex{},
-		tablesMutex:               &sync.RWMutex{},
-	}
-}
-
-func (this *DataIteratorState) UpdateTargetPK(table string, pk uint64) {
-	this.targetPkMutex.Lock()
-	defer this.targetPkMutex.Unlock()
-
-	this.targetPrimaryKeys[table] = pk
-}
-
-func (this *DataIteratorState) UpdateLastSuccessfulPK(table string, pk uint64) {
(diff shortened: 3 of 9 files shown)


hkdsun:
Wow, exciting times!!! Thanks for starting this right away 🙂 

----

> this centralization does make using the DataIterator/BinlogStreamer independently of the Ferry slightly more difficult

It shouldn't have to. Why can't we do something like this?

```
   if b.StateTracker != nil {
     b.StateTracker.UpdateLastStreamedBinlogPosition(...)
   }
```

----

>Also added in this PR is to listen to the SIGTERM/SIGINT given to the process, which will trigger a state dump to stdout. This also requires suppressing logging from go-mysql so it doesn't pollute the state dump. This is okay as everything logged from go-mysql we already log via logrus into stderr.

It would be great if the state dump action is configurable right off the bat. I can't think of use-cases for dumping to stdout other than having logs, playing around, etc. 

Usually we'd want to: (1) dump to local disk like in copydb, (2) serialize to a datastore like in the shop mover, or (3) save it to a local variable and feed it to a new ferry to test resume behaviour

binlog_streamer.go
```diff
@@ -169,6 +177,8 @@ func (s *BinlogStreamer) Run() {
 		default:
 			s.updateLastStreamedPosAndTime(ev)
 		}
+
+		s.StateTracker.UpdateLastStreamedBinlogPosition(s.lastStreamedBinlogPosition)
```
hkdsun: putting this call here is confusing because of the access to `s.lastStreamedBinlogPosition` which is updated somewhere completely different. why not put it in the `updateLastStreamedPosAndTime` method?

also, why don't we want to update state when we hit a `RotateEvent`?

shuhaowu:
>    if b.StateTracker != nil {
>     b.StateTracker.UpdateLastStreamedBinlogPosition(...)
>   }

This seems like a good idea, although it would break things like the ControlServer in non-intuitive ways if an API consumer were to skip the StateTracker and use the other Ghostferry provided components with it. I'm tempted to leave it as is, until we have a concrete use case for making it optional (this might happen soon as I'm evaluating refactoring the verifiers to the Ferry level).

==========

> It would be great if the state dump action is configurable right off the bat. I can't think of use-cases for dumping to stdout other than having logs, playing around, etc.
>
> Usually we'd want to: (1) dump to local disk like in copydb, (2) serialize to a datastore like in the shop mover, or (3) save it to a local variable and feed it to a new ferry to test resume behaviour

The state dump action is completely dependent on the API consumer for Ghostferry. Right now, copydb implements state dumping via stdout as this has maximum flexibility (redirect it to a file, a pipe, a parent process, whatever). In the general case, the state can be serialized into a JSON serializable struct at any point during the run (via `StateTracker.Serialize()` for this PR, to be integrated with Ferry in the next PR). This means ghostferry-sharding, or a custom application, can take this struct and do whatever it wants.

As a historical note: It was also no a coincident that we were logging to stderr all this time. Even the earliest commits had this design of stderr = log, stdout = state dump. The choice of logrus logging to stderr was a conscious decision, which happened to be the default of the library. In fact, we previously were already state dumping into stdout, it was just not a resumable state dump.

error_handler.go
```diff
@@ -28,15 +28,15 @@ func (this *PanicErrorHandler) Fatal(from string, err error) {
...
 
-	state := make(map[string]interface{})
-	state["LastSuccessfulBinlogPos"] = this.Ferry.BinlogStreamer.GetLastStreamedBinlogPosition()
-	state["LastSuccessfulPrimaryKeys"] = this.Ferry.DataIterator.CurrentState.LastSuccessfulPrimaryKeys()
-	state["CompletedTables"] = this.Ferry.DataIterator.CurrentState.CompletedTables()
+	serializedState := &SerializableState{
+		GhostferryVersion:         VersionString,
+		LastKnownTableSchemaCache: this.Ferry.Tables,
+	}
+	this.Ferry.StateTracker.Serialize(serializedState)
```
hkdsun: this is not very gopher, could we pass the `Tables` and `VersionString` as arguments and get back a `SerializableState` instead?

shuhaowu:
> Any thoughts on how we're going to handle signals during cutover?
> 
> I'm thinking disallow interruption?

Yes. This would be the way to go.

> I wasn't contesting that choice, I was just saying we're going to have to provide a generic way of fetching the state for interruption.

Yeah this PR definitely allow you to do that. The only thing this PR does is to make the choice of using stdout within copydb.

> listen to the SIGTERM/SIGINT given to the process, which will trigger a state dump to stdout.
> 
> This also doesn't seem nice to do as a library, should we instead leave this up to the caller?

This is a great idea :+1:. Let' me implement that.

hkdsun:
@shuhaowu, are you still working on this? Could you ping when it's ready for another round of review?

shuhaowu:
Yeah I want to finish the error handler stuff first now given how significant of a change it will be. I'll ping again once it's ready for review.

binlog_streamer.go
```diff
@@ -87,6 +94,7 @@ func (s *BinlogStreamer) ConnectBinlogStreamerToMysql() error {
 		s.logger.WithError(err).Error("failed to read current binlog position")
 		return err
 	}
+	s.StateTracker.UpdateLastStreamedBinlogPosition(s.lastStreamedBinlogPosition)
```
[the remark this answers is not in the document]
shuhaowu: This is removed as we are now tracking the binlog position in the BinlogWriter for correctness.

binlog_streamer.go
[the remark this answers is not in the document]
shuhaowu: Same: moved to BinlogWriter, after the `db.Exec` call to ensure correctness.

shuhaowu:
With #66 merged, this PR is now ready for review. 

Note that there have been some talks about handling SIGINT/TERM, I've removed that code from this PR to keep things simple as it fits well with my PR enabling interrupt/resume coming next.

status.go
```diff
@@ -79,9 +79,17 @@ func FetchStatus(f *Ferry, v Verifier) *Status {
 
 	// Getting all table statuses
 	status.TableStatuses = make([]*TableStatus, 0, len(f.Tables))
-	completedTables := f.DataIterator.CurrentState.CompletedTables()
-	targetPKs := f.DataIterator.CurrentState.TargetPrimaryKeys()
-	lastSuccessfulPKs := f.DataIterator.CurrentState.LastSuccessfulPrimaryKeys()
+
+	serializedState := f.StateTracker.Serialize(nil)
```
hkdsun: `nil` argument 😢

binlog_writer.go
```diff
@@ -110,7 +116,7 @@ func (b *BinlogWriter) writeEvents(events []DMLEvent) error {
 		return fmt.Errorf("exec query (%d bytes): %v", len(query), err)
 	}
 
-	// lastWrittenBinlogPosition = events[len(events)-1].BinlogPosition()
+	b.StateTracker.UpdateLastWrittenBinlogPosition(events[len(events)-1].BinlogPosition())
```
hkdsun: is there a test that would fail if we removed this line?

how about one that would fail if it was executed at the wrong time?

data_iterator.go
```diff
@@ -225,7 +109,7 @@ func (d *DataIterator) Run() {
 					}
 
 					logger.Debugf("updated last successful PK to %d", pkpos)
-					d.CurrentState.UpdateLastSuccessfulPK(table.String(), pkpos)
+					d.StateTracker.UpdateLastSuccessfulPK(table.String(), pkpos)
```
hkdsun: copy pasting the other comment 😛 

is there a test that would fail if we removed this line?

how about one that would fail if it was executed at the wrong time?

binlog_streamer.go
```diff
@@ -35,7 +35,7 @@ type BinlogStreamer struct {
 	eventListeners []func([]DMLEvent) error
 }
 
-func (s *BinlogStreamer) Initialize() (err error) {
+func (s *BinlogStreamer) Initialize() error {
```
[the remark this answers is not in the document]
shuhaowu: This is to stay consistent with the other APIs for DataIterator/Ferry/etc.

binlog_writer.go
shuhaowu: I thought about this for quite a bit and I'm not sure if there is a way to positively assert this line: we can't know what the binlog position is when we execute a statement easily. To do so we need to tail the binlogs in the tests, which is challenging. The best way I can think of testing this is blackbox testing: if we resume a dumped state, does it correctly copy the table with writes happen simultaneously. This should be enough that the dumped coordinates are "correct".

data_iterator.go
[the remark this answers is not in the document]
shuhaowu: Yes. This is a fantastic idea.

binlog_writer.go
shuhaowu: Actually I think there's a way to assert the dumped binlog position from the ruby test, but I'll have to spend sometime looking at it. Either way, I don't think I can do it from this PR.

data_iterator.go
shuhaowu: This is more easily testable, but I think black box testing via the ruby integration tests should be enough. 

As stated earlier, I also use the ruby integration test to assert the dumped data iterator coordinates, but requires additional investigation and can't be done in this PR as the ruby test framework is not merged yet.

shuhaowu:
I updated this branch based on the comments and it is ready for another review.

state_tracker.go
```diff
@@ -0,0 +1,151 @@
...
+	copySpeedLog *ring.Ring
+}
+
+func (s *StateTracker) UpdateLastSuccessfulPK(table string, pk uint64) {
+	s.tableMutex.Lock()
+	defer s.tableMutex.Unlock()
+
+	s.lastSuccessfulPrimaryKeys[table] = pk
+
+	deltaPK := pk - s.lastSuccessfulPrimaryKeys[table]
```
hkdsun: wouldn't this always be 0 now?

data_iterator.go
```diff
@@ -236,7 +100,7 @@ func (d *DataIterator) Run() {
 				}
 
 				logger.Debug("table iteration completed")
-				d.CurrentState.MarkTableAsCompleted(table.String())
+				d.StateTracker.MarkTableAsCompleted(table.String())
```
hkdsun: I love how all call sites to the critical StateTracker methods make sense now. I think we did good 🙂

state_tracker.go
shuhaowu: Thanks for the catch!

shuhaowu:
@hkdsun Thank you so much for the detailed review!

PR #61, opened 2018-08-14
iterative verifier: allow ignored columns

hkdsun:
**Problem**
Sometimes, it's desirable to intentionally allow ignoring certain columns like `created_at` during verification. This happens most often in tables that share data between tenants.

**But the sharding package only copies tables that have a tenant_id, how can they share data?**
We allow two other types of tables: joined tables and primary key tables. This is exactly where we need to use this feature

**Solution**
It's a fairly simple change to exclude verification of some columns without impacting their copying

@Shopify/pods

### DIFF
--- a/iterative_verifier.go
+++ b/iterative_verifier.go
@@ -137,6 +137,7 @@ type IterativeVerifier struct {
 
 	Tables              []*schema.Table
 	IgnoredTables       []string
+	IgnoredColumns      map[string]map[string]struct{}
 	DatabaseRewrites    map[string]string
 	TableRewrites       map[string]string
 	Concurrency         int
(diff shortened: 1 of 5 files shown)


iterative_verifier.go
```diff
@@ -137,6 +137,7 @@ type IterativeVerifier struct {
 
 	Tables              []*schema.Table
 	IgnoredTables       []string
+	IgnoredColumns      map[string][]string
```
fw42: If I remember correctly, we identify columns via `#{table_name}.#{column_name}` in other places (so this could just be a `[]string` instead of a `map[string][]string`). Curious why you chose this approach instead?

iterative_verifier.go
```diff
@@ -572,6 +573,27 @@ func (v *IterativeVerifier) tableIsIgnored(table *schema.Table) bool {
 	return false
 }
 
+func (v *IterativeVerifier) verifiedColumns(table *schema.Table) []schema.TableColumn {
```
fw42: nitpick: The name makes it sound like the columns are already verified. How about `columnsToVerify`?

iterative_verifier.go
hkdsun: Where did you find `#{table_name}.#{column_name}`?

The approach here makes skipping tables that have no ignored columns (the vast majority)  O(1). In the alternative approach you suggested you'd have to pair up each table and its columns and iterate through the configured list to check if there are matches

iterative_verifier.go
```diff
@@ -572,6 +573,27 @@ func (v *IterativeVerifier) tableIsIgnored(table *schema.Table) bool {
 	return false
 }
 
+func (v *IterativeVerifier) columnsToVerify(table *schema.Table) []schema.TableColumn {
+	ignoredCols, containsIgnoredColumns := v.IgnoredColumns[table.Name]
+	if !containsIgnoredColumns {
+		return table.Columns
+	}
+
+	ignoredColsSet := make(map[string]struct{})
```
fjordan: Would it be possible to structure the configuration of the ignored columns to remove lines 582-585? Or, instead of having to iterate and create the set using a map each time would it be better to just do it at startup instead of this phase in the runtime?

PR #62, opened 2018-08-20
context.WithContext everywhere to allow us to stop goroutines

shuhaowu:
The point of this PR is to have the ability to stop goroutines during a run without having to use panic. This will allow us to test the interrupt/resume code. A couple of additional small things are fixed and they're done in different commits, each with their own detailed message. 

One notable commit is the second to last one, where we moved the onFinishedIterations code, which simply waits until `AutomaticCutover` is true so things like copydb could have a manual cutover phase, into Run directly. This is necessary to be able to cancel the main Ferry as well as makes the code cleaner to read (as all the logic goes into one place).

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -115,7 +115,8 @@ func (s *BinlogStreamer) Run() {
 		var timedOut bool
 
 		err := WithRetries(5, 0, s.logger, "get binlog event", func() (er error) {
-			ctx, _ := context.WithTimeout(context.Background(), 500*time.Millisecond)
+			ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
+			defer cancel()
 			ev, er = s.binlogStreamer.GetEvent(ctx)
 
 			if er == context.DeadlineExceeded {
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -32,7 +32,7 @@ type BinlogStreamer struct {
 	stopRequested bool
 
 	logger         *logrus.Entry
-	eventListeners []func(context.Context, []DMLEvent) error
+	eventListeners []func([]DMLEvent) error
 }
 
 func (s *BinlogStreamer) Initialize() (err error) {
@@ -176,7 +176,7 @@ func (s *BinlogStreamer) Run(ctx context.Context) {
 	}
 }
 
-func (s *BinlogStreamer) AddEventListener(listener func(context.Context, []DMLEvent) error) {
+func (s *BinlogStreamer) AddEventListener(listener func([]DMLEvent) error) {
 	s.eventListeners = append(s.eventListeners, listener)
 }
 
@@ -188,7 +188,7 @@ func (s *BinlogStreamer) IsAlmostCaughtUp() bool {
 	return time.Now().Sub(s.lastProcessedEventTime) < caughtUpThreshold
 }
 
-func (s *BinlogStreamer) FlushAndStop(ctx context.Context) {
+func (s *BinlogStreamer) FlushAndStop(ctx context.Context) error {
 	if ctx.Err() == nil {
 		s.logger.Info("requesting binlog streamer to stop")
 	}
@@ -204,16 +204,18 @@ func (s *BinlogStreamer) FlushAndStop(ctx context.Context) {
 	})
 
 	if err == context.Canceled {
-		return
+		return err
 	}
 
 	if err != nil {
-		s.ErrorHandler.Fatal("binlog_streamer", err)
+		s.ErrorHandler.ReportError("binlog_streamer", err)
+		return err
 	}
 
 	s.logger.WithField("target_position", s.targetBinlogPosition).Info("current stop binlog position was recorded")
 
 	s.stopRequested = true
+	return nil
 }
 
 func (s *BinlogStreamer) updateLastStreamedPosAndTime(ev *replication.BinlogEvent) {
@@ -275,7 +277,7 @@ func (s *BinlogStreamer) handleRowsEvent(ctx context.Context, ev *replication.Bi
 	}
 
 	for _, listener := range s.eventListeners {
-		err := listener(ctx, events)
+		err := listener(events)
 		if err != nil {
 			return err
 		}
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -32,7 +32,7 @@ type BinlogStreamer struct {
 	stopRequested bool
 
 	logger         *logrus.Entry
-	eventListeners []func([]DMLEvent) error
+	eventListeners []func(context.Context, []DMLEvent) error
 }
 
 func (s *BinlogStreamer) Initialize() (err error) {
@@ -102,7 +102,7 @@ func (s *BinlogStreamer) ConnectBinlogStreamerToMysql() error {
 	return nil
 }
 
-func (s *BinlogStreamer) Run() {
+func (s *BinlogStreamer) Run(ctx context.Context) {
 	defer func() {
 		s.logger.Info("exiting binlog streamer")
 		s.binlogSyncer.Close()
@@ -114,10 +114,10 @@ func (s *BinlogStreamer) Run() {
 		var ev *replication.BinlogEvent
 		var timedOut bool
 
-		err := WithRetries(5, 0, s.logger, "get binlog event", func() (er error) {
-			ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
-			defer cancel()
-			ev, er = s.binlogStreamer.GetEvent(ctx)
+		err := WithRetriesContext(ctx, 5, 0, s.logger, "get binlog event", func() (er error) {
+			eventCtx, eventCancel := context.WithTimeout(ctx, 500*time.Millisecond)
+			defer eventCancel()
+			ev, er = s.binlogStreamer.GetEvent(eventCtx)
 
 			if er == context.DeadlineExceeded {
 				timedOut = true
@@ -126,6 +126,11 @@ func (s *BinlogStreamer) Run() {
 
 			return er
 		})
+
+		if err == context.Canceled {
+			s.logger.Info("cancellation received, terminating goroutine abruptly")
+			return
+		}
 
 		if err != nil {
 			s.ErrorHandler.Fatal("binlog_streamer", err)
@@ -147,7 +152,7 @@ func (s *BinlogStreamer) Run() {
 				"file": s.lastStreamedBinlogPosition.Name,
 			}).Info("rotated binlog file")
 		case *replication.RowsEvent:
-			err = s.handleRowsEvent(ev)
+			err = s.handleRowsEvent(ctx, ev)
 			if err != nil {
 				s.logger.WithError(err).Error("failed to handle rows event")
 				s.ErrorHandler.Fatal("binlog_streamer", err)
@@ -171,7 +176,7 @@ func (s *BinlogStreamer) Run() {
 	}
 }
 
-func (s *BinlogStreamer) AddEventListener(listener func([]DMLEvent) error) {
+func (s *BinlogStreamer) AddEventListener(listener func(context.Context, []DMLEvent) error) {
 	s.eventListeners = append(s.eventListeners, listener)
 }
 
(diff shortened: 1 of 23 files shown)


shuhaowu:
I just looked at the code again and I realized there are a lot more places where we need context. Examples: wait until replica has caught up to master, wait for row copy to complete, getting the target binlog position in order to flush the binlog position, etc etc etc.

I'm really wondering if this is the best way to do this as it adds a tonne of complexity... Thoughts?

hkdsun:
I think the concept of cancellation needs to be solved for us in one way or another. The solution in Golang is to use the context package. That's the whole reason for its existence:

> Package context defines the Context type, which carries deadlines, cancelation signals, and other request-scoped values across API boundaries and between processes.
[(source)](https://golang.org/pkg/context/)

Interestingly enough, pretty much every function call involving a "request" at Google needs a context:

> At Google, we require that Go programmers pass a Context parameter as the first argument to every function on the call path between incoming and outgoing requests
[(source)](https://blog.golang.org/context)

If you're really finding this adds a lot of complexity, we could give the Tomb package a good try: https://godoc.org/gopkg.in/tomb.v2

I read some interesting articles on topic in the [ExperienceReports#context](https://github.com/golang/go/wiki/ExperienceReports#context) page and my conclusion is that in this PR we are doing everything that the package was designed for and _not_ doing everything that the community is wary of.

Anyway, I think what makes the most sense is to let the caller of Ghostferry cancel/interrupt the Ferry at any time in one way or another, it happens to be passing around context in Go

shuhaowu:
Yeah I've looked around and didn't find much beyond tomb myself, which still requires manual handling.

I think we'll have to just use context as is. I'll just need to rework this PR from another angle (start from the methods on the bottom as oppose to starting from the Ferry.Run level) to get a better view of the problem and ensure it's added to all relevant locations.

hkdsun:
Are there any static analysis tool to help us here?

shuhaowu:
Not to the extent we want as far as I can tell. The only thing that currently exists is go vet, which only checks whether or not the cancel method is called for every context.

shuhaowu:
I just rewrote this PR to actually account for all of the places where a goroutine can get stuck if we invoked cancel via something like signals. I had to read the code from CopydbFerry.Run and inspect each method call (and their methods calls, and so on) until I hit a method where cancellation doesn't make sense anymore (either no more IO operations, SQL operations, or metric emits). Then I had to add context.Context to all the functions examined. It's also possible that I missed something, but I tried my best to follow every method call all the way down.

I'm not even sure how anyone would review this PR without going through the entire codebase, which took a couple of hours to do. I tried to structure the commits so they're one segment at a time, but even then it's pretty hard to follow.

I haven't yet hooked up the context.Context to sharding.Ferry. They're marked as context.TODO for now. I'll likely have to copy the signal handling code from copydb to sharding, but I want to get some feedback first.

Here's a run of copydb being cancelled, the following log lines should show up:

```
^CWARN[0002] shutdown signal received                      signal=interrupt
INFO[0002] cancellation received, terminating goroutine abruptly  tag=binlog_writer
INFO[0002] entering cutover phase, notifying caller that row copy is complete  tag=ferry
INFO[0002] cancellation received, terminating goroutine abruptly  tag=binlog_streamer
INFO[0002] exiting binlog streamer                       tag=binlog_streamer
[2018/08/21 17:41:48] [info] binlogsyncer.go:163 syncer is closing...
ERRO[0002] error on ListenAndServe                       error="http: Server closed"
[2018/08/21 17:41:49] [error] binlogsyncer.go:623 connection was bad
[2018/08/21 17:41:49] [error] binlogstreamer.go:57 close sync with err: Sync was closed
[2018/08/21 17:41:49] [info] binlogsyncer.go:178 syncer is closed
INFO[0002] ghostferry run is complete, shutting down auxiliary services  tag=ferry
WARN[0002] ghostferry run was canceled, certain log lines may be inaccurate  tag=ferry
INFO[0002] copydb run was canceled via signals  
```

binlog_streamer.go
```diff
@@ -114,9 +114,10 @@ func (s *BinlogStreamer) Run() {
 		var ev *replication.BinlogEvent
 		var timedOut bool
 
-		err := WithRetries(5, 0, s.logger, "get binlog event", func() (er error) {
-			ctx, _ := context.WithTimeout(context.Background(), 500*time.Millisecond)
-			ev, er = s.binlogStreamer.GetEvent(ctx)
+		err := WithRetriesContext(ctx, 5, 0, s.logger, "get binlog event", func() (er error) {
+			eventCtx, eventCancel := context.WithTimeout(ctx, 500*time.Millisecond)
+			defer eventCancel()
+			ev, er = s.binlogStreamer.GetEvent(eventCtx)
```
fjordan: nit: can we rename `er` to `err` and just define it as a `var` underneath `timeout`?

binlog_writer.go
```diff
@@ -69,16 +86,20 @@ func (b *BinlogWriter) Stop() {
 	close(b.binlogEventBuffer)
 }
 
-func (b *BinlogWriter) BufferBinlogEvents(events []DMLEvent) error {
+func (b *BinlogWriter) BufferBinlogEvents(ctx context.Context, events []DMLEvent) error {
 	for _, event := range events {
-		b.binlogEventBuffer <- event
+		select {
```
fjordan: 👍

ferry.go
```diff
@@ -356,19 +377,26 @@ func (f *Ferry) RunStandaloneDataCopy(tables []*schema.Table) error {
 	dataIterator.AddBatchListener(f.BatchWriter.WriteRowBatch)
 	f.logger.WithField("tables", tables).Info("starting standalone table copy")
 
-	dataIterator.Run()
+	dataIterator.Run(context.TODO())
```
fjordan: why is this a `context.TODO`? Do we not have access to the parent context here?

ferry.go
```diff
@@ -378,10 +406,17 @@ func (f *Ferry) WaitUntilBinlogStreamerCatchesUp() {
...
-func (f *Ferry) FlushBinlogAndStopStreaming() {
+func (f *Ferry) FlushBinlogAndStopStreaming(ctx context.Context) {
 	if f.WaitUntilReplicaIsCaughtUpToMaster != nil {
 		isReplica, err := CheckDbIsAReplica(f.WaitUntilReplicaIsCaughtUpToMaster.MasterDB)
 		if err != nil {
+			// TODO: this method is usually called in the main thread. panicking
+			// here is different from the style where only Run panics (as it runs
+			// in a separate thread).
+			//
+			// Perhaps it's also a better idea to return errors in all situations
```
fjordan: I think wherever we can provide additional insights to the user in the form of an error is best if we can also ensure we're not going to corrupt anything or cause data loss. `panic`ing is definitely a way to ensure we immediately abort, but with the additions of pause/resume we'll need more control 👍

binlog_streamer.go
```diff
@@ -146,7 +152,7 @@ func (s *BinlogStreamer) Run() {
 				"file": s.lastStreamedBinlogPosition.Name,
 			}).Info("rotated binlog file")
 		case *replication.RowsEvent:
-			err = s.handleRowsEvent(ev)
+			err = s.handleRowsEvent(ctx, ev)
```
hkdsun: This `err` needs to check for context.Canceled

hkdsun: Do we really need to pass the context to the listener? I want reiterate that our cancellation policy is fairly relaxed

binlog_streamer.go
```diff
@@ -182,22 +188,29 @@ func (s *BinlogStreamer) IsAlmostCaughtUp() bool {
 	return time.Now().Sub(s.lastProcessedEventTime) < caughtUpThreshold
 }
 
-func (s *BinlogStreamer) FlushAndStop() {
-	s.logger.Info("requesting binlog streamer to stop")
+func (s *BinlogStreamer) FlushAndStop(ctx context.Context) {
+	if ctx.Err() == nil {
+		s.logger.Info("requesting binlog streamer to stop")
+	}
```
hkdsun: What's this conditional for?

binlog_writer.go
```diff
@@ -49,14 +57,23 @@ func (b *BinlogWriter) Run() {
...
-		err := WithRetries(b.WriteRetries, 0, b.logger, "write events to target", func() error {
-			return b.writeEvents(batch)
+		err := WithRetriesContext(ctx, b.WriteRetries, 0, b.logger, "write events to target", func() error {
+			return b.writeEvents(ctx, batch)
 		})
+
+		if err == context.Canceled {
+			b.logger.Info("cancellation received, terminating goroutine abruptly")
+			return
+		}
```
hkdsun: Hmmmm.. this seems very sketchy. I don't believe the binlog writer should take a context. But rather it should shut down naturally due to its buffer channel being closed by the Ferry/BinlogStreamer.  Otherwise, it will lose writes that we assumed would complete 

https://github.com/Shopify/ghostferry/blob/5e461ead1b6104d83fc736ece2b426e354a7cbe4/ferry.go#L279

https://github.com/Shopify/ghostferry/blob/5e461ead1b6104d83fc736ece2b426e354a7cbe4/ferry.go#L341-L343

data_iterator.go
```diff
@@ -160,7 +161,11 @@ func (d *DataIterator) Initialize() error {
 	return nil
 }
 
-func (d *DataIterator) Run() {
+func (d *DataIterator) Run(ctx context.Context) {
+	if ctx.Err() != nil {
+		return
+	}
```
hkdsun: Can you justify this too? I'm not sure I get why we need it

data_iterator.go
```diff
@@ -185,22 +190,28 @@ func (d *DataIterator) Run() {
...
 
 			for {
-				table, ok := <-tablesQueue
+				var table *schema.Table
+				var ok bool
+				select {
+				case table, ok = <-tablesQueue:
+				case <-ctx.Done():
+					logrus.WithField("tag", "table_iterator").Info("cancellation received, terminating goroutine abruptly")
+				}
```
hkdsun: We don't need this here. The producer of the `tablesQueue` channel can close the channel and terminate this loop.

data_iterator.go
```diff
@@ -185,22 +190,28 @@ func (d *DataIterator) Run() {
...
-				err := cursor.Each(func(batch *RowBatch) error {
+				err := cursor.Each(ctx, func(batch *RowBatch) error {
 					metrics.Count("RowEvent", int64(batch.Size()), []MetricTag{
 						MetricTag{"table", table.Name},
 						MetricTag{"source", "table"},
 					}, 1.0)
 
 					for _, listener := range d.batchListeners {
-						err := listener(batch)
+						err := listener(ctx, batch)
```
hkdsun: Again, I'm not sure why the listener needs a context. There's no rush to terminate _immediately_.

We want to finish up writing our batch, record the last successful PK, and _then_ check the termination signal.

i.e. the only places that the DataIterator should check for interruption are:
(1) After a batch has been written and its bookkeeping has been updated
(2) The producer of the `tablesQueue` channel

ferry.go
```diff
@@ -26,6 +26,7 @@ const (
 	StateWaitingForCutover = "wait-for-cutover"
 	StateCutover           = "cutover"
 	StateDone              = "done"
+	StateCanceled          = "canceled"
```
hkdsun: Can we call it `Interrupted`/`Interruption`? Canceled makes it sound like the user changed their mind

ferry.go
```diff
@@ -310,35 +312,54 @@ func (f *Ferry) Run() {
...
+	f.OverallState = StateWaitingForCutover
+	f.waitUntilAutomaticCutoverIsTrue(ctx)
+
+	f.logger.Info("entering cutover phase, notifying caller that row copy is complete")
+	f.OverallState = StateCutover
+	f.notifyRowCopyComplete(ctx)
+
+	binlogWg.Wait()
+
+	f.logger.Info("ghostferry run is complete, shutting down auxiliary services")
```
hkdsun: I wouldn't say "complete". An interrupted run is not complete.

Also "shutting down auxiliary services" is an implementation detail that i cannot see any user caring about

I would just say "ghostferry is shutting down".

ferry.go
```diff
@@ -356,19 +377,26 @@ func (f *Ferry) RunStandaloneDataCopy(tables []*schema.Table) error {
...
+	}
 }
 
-func (f *Ferry) WaitUntilBinlogStreamerCatchesUp() {
+func (f *Ferry) WaitUntilBinlogStreamerCatchesUp(ctx context.Context) {
 	for !f.BinlogStreamer.IsAlmostCaughtUp() {
-		time.Sleep(500 * time.Millisecond)
+		select {
+		case <-ctx.Done():
+			return
```
hkdsun: This seems super dangerous. We should at the very least return an `error` here, otherwise the caller would assume that the binlog streamer has caught up and they can proceed to cutover 💩

ferry.go
```diff
@@ -395,30 +430,30 @@ func (f *Ferry) FlushBinlogAndStopStreaming() {
...
 		}
 
-		err = f.WaitUntilReplicaIsCaughtUpToMaster.Wait()
 		if err != nil {
 			f.ErrorHandler.Fatal("wait_replica", err)
 		}
 	}
 
-	f.BinlogStreamer.FlushAndStop()
+	f.BinlogStreamer.FlushAndStop(ctx)
```
hkdsun: Again, this seems super dangerous. The method should either return an error (that can be `context.Canceled`or nil) or just not allow interruption at this stage.

ferry.go
```diff
@@ -395,30 +430,30 @@ func (f *Ferry) FlushBinlogAndStopStreaming() {
...
-
+func (f *Ferry) waitUntilAutomaticCutoverIsTrue(ctx context.Context) {
 	for !f.AutomaticCutover {
-		time.Sleep(1 * time.Second)
-		f.logger.Debug("waiting for AutomaticCutover to become true before signaling for row copy complete")
+		select {
+		case <-time.After(1 * time.Second):
+			f.logger.Debug("waiting for AutomaticCutover to become true before signaling for row copy complete")
+		case <-ctx.Done():
+			return
```
hkdsun: Ditto

binlog_writer.go
shuhaowu: Good catch here. I overlooked the fact that we're buffering the binlog events to write here.

binlog_streamer.go
shuhaowu: One of the issues of not passing this down to the listeners is because we have listeners that have retries that could proceed to block for a long period of time after a cancel has been received.

This kind of blocking might become a concern during testing and is the ultimate reason why I put the cancellation all over the place. Perhaps we can limit our test designs such that those WithRetries calls to interact with the database do not fail.

binlog_writer.go
shuhaowu: Actually this is an issue with the panic based interruption as well. We have no idea as is, if a binlog has been written to the database or not in the binlog streamer.

ferry.go
shuhaowu: There's currently not capabilities to stop the process of the delta copier, which uses this method (which does not take in a context for the time being).

PR #65, opened 2018-09-18
Updated Ghostferry TLA model with BinlogWriter

shuhaowu:
This is an important component of Ghostferry and thus deserves to be modeled. It also has implications when modeling Ghostferry with interrupt + resume.

In the previous TLA+ model, the binlogs are written directly to the target database as they are read by the BinlogStreamer. In the present model, they are send into a queue. This queue is read by a separate process (BinlogWriter) and written to the target database atomically one at a time. No batch writing (multistatement transaction that we have implemented) is modeled as I believe only the order is important, not how many binlog entries are written in a single atomic operation. 

Checking this new specification on the standard model yields no invariant violations. I also tested that the specification is correct by intentionally writing the incorrect data and ensuring that the model checking finds some invariant violations, which it does.

The large diff is due to artifacts from the PlusCal => TLA+ translation and comments. The actual PlusCal source changes are about **15++ 5--**

### DIFF
--- a/tlaplus/ghostferry.tla
+++ b/tlaplus/ghostferry.tla
@@ -30,15 +30,17 @@
  (the application/dba/whatever) will perform INSERT/UPDATE/DELETEs on the source
  database concurrently with respect to Ghostferry.
 
- This whole process is modeled as four separate processes:
+ This whole process is modeled as five separate processes:
  1. TableIterator: Performs the job of the DataIterator, but only on one table
                    (see Data Model for justification).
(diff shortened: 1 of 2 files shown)


tlaplus/ghostferry.tla
```diff
@@ -386,24 +409,27 @@ PossibleRecords == Records \cup {NoRecordHere}
  ***************************************************************************)
 \* BEGIN TRANSLATION
```
shuhaowu: Everything below this line is generated.

PR #66, opened 2018-09-18
Pass the current BinlogPosition to DMLEvent so BinlogWriter can set lastWrittenBinlogPosition

shuhaowu:
Since the BinlogWriter is a separate goroutine to the BinlogStreamer, the position that it wrote to is important if we want to interrupt and resume Ghostferry (interrupt saves BinlogWriter position, and resume starts the BinlogStreamer from the BinlogWriter position).

To allow for this, we need to pass the binlog position from the binlog streamer, via the DMLEvent, to the binlog writer. This patch adds BinlogPosition() to DMLEvent to make this possible.

### DIFF
--- a/binlog_writer.go
+++ b/binlog_writer.go
@@ -109,5 +109,8 @@ func (b *BinlogWriter) writeEvents(events []DMLEvent) error {
 	if err != nil {
 		return fmt.Errorf("exec query (%d bytes): %v", len(query), err)
 	}
+
+	// lastWrittenBinlogPosition = events[len(events)-1].BinlogPosition()
+
 	return nil
 }
--- a/test/binlog_streamer_test.go
+++ b/test/binlog_streamer_test.go
@@ -2,6 +2,8 @@ package test
 
 import (
 	"database/sql"
+	"strings"
+	"sync"
 	"testing"
 
 	"github.com/Shopify/ghostferry"
@@ -15,6 +17,7 @@ type BinlogStreamerTestSuite struct {
 
 	config         *ghostferry.Config
 	binlogStreamer *ghostferry.BinlogStreamer
+	sourceDb       *sql.DB
 }
 
 func (this *BinlogStreamerTestSuite) SetupTest() {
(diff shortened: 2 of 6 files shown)


binlog_writer.go
```diff
@@ -109,5 +109,8 @@ func (b *BinlogWriter) writeEvents(events []DMLEvent) error {
 	if err != nil {
 		return fmt.Errorf("exec query (%d bytes): %v", len(query), err)
 	}
+
+	// lastWrittenBinlogPosition = events[len(events)-1].BinlogPosition()
```
shuhaowu: This is temporary. Once this merges, I'll change this line in #60 such that we send this position to the StateTracker.

test/binlog_streamer_test.go
```diff
@@ -70,6 +83,35 @@ func (this *BinlogStreamerTestSuite) TestConnectErrorsOutIfErrorInServerIdGenera
...
+func (this *BinlogStreamerTestSuite) TestBinlogStreamerSetsBinlogPositionOnDMLEvent() {
+	err := this.binlogStreamer.ConnectBinlogStreamerToMysql()
+	this.Require().Nil(err)
+
+	eventAsserted := false
+
+	this.binlogStreamer.AddEventListener(func(evs []ghostferry.DMLEvent) error {
+		eventAsserted = true
+		this.Require().Equal(1, len(evs))
+		this.Require().NotEqual("", evs[0].BinlogPosition().Name)
```
hkdsun: If you wanna make this more solid `SELECT @@log_bin_basename` should give you the basename that you can then `assertMatch`

test/binlog_streamer_test.go
```diff
@@ -70,6 +83,35 @@ func (this *BinlogStreamerTestSuite) TestConnectErrorsOutIfErrorInServerIdGenera
...
+		this.Require().NotEqual(0, evs[0].BinlogPosition().Pos)
```
hkdsun: can you assert(pos > 0) instead?

hkdsun:
thanks for making this easier to review ❤️

it's a huge help when we're not primarily focused on this!

test/binlog_streamer_test.go
shuhaowu: We actually set `--log-bin=mysql-bin` in the docker-compose config so I'll just match against mysql-bin directly. Thanks for the suggestion.

PR #67, opened 2018-09-25
Interrupt Ghostferry via signals and resume from SerializedState

shuhaowu:
This PR implements the ability to interrupt Ghostferry via SIGTERM/SIGINT. These signals causes the error handler to dump the current state to stdout and panic to terminate the process.

Resume is also implemented: give a SerializedState as the StateToResumeFrom on the Config object and DataIterator will resume via the LastSuccessfulPrimaryKeys, BinlogStreamer will resume via LastWrittenBinlogPosition, and Ferry uses the table schema cache from the SerializedState.

Testing of this will occur in a PR after this, along with the ruby integration test framework, which I'm still iterating on. The code here stands on its own, however.

**Note: this PR does not implement the reconciliation process. This will be in a subsequent PR.** Interestingly, the implementation is currently already "correct" if no schema change occured, as we resume streaming from LastWrittenBinlogPosition. This will of course be changed during the reconciliation process.

In addition, I also did some quick refactoring:

- Created some public methods to create  components like DataIterator, BinlogStream, etc (Ferry.NewDataIterator, Ferry.NewDataIteratorWithoutStateTracker, etc)
- Made StateTracker optional: this is needed as we will reuse some of these components in places where we don't want them to update the state tracker (think: reconciliation process).

I tried my best to make things incremental, so each commit should stand on their own. It might be best to review in that order.

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -15,8 +15,9 @@ import (
 const caughtUpThreshold = 10 * time.Second
 
 type BinlogStreamer struct {
-	Db           *sql.DB
-	Config       *Config
+	DB           *sql.DB
+	DBConfig     DatabaseConfig
+	MyServerId   uint32
 	ErrorHandler ErrorHandler
 	Filter       CopyFilter
 
@@ -35,25 +36,25 @@ type BinlogStreamer struct {
 	eventListeners []func([]DMLEvent) error
 }
 
-func (s *BinlogStreamer) Initialize() error {
-	s.logger = logrus.WithField("tag", "binlog_streamer")
-	s.stopRequested = false
-	return nil
+func (s *BinlogStreamer) ensureLogger() {
+	if s.logger == nil {
+		s.logger = logrus.WithField("tag", "binlog_streamer")
+	}
 }
 
 func (s *BinlogStreamer) createBinlogSyncer() error {
 	var err error
 	var tlsConfig *tls.Config
 
-	if s.Config.Source.TLS != nil {
-		tlsConfig, err = s.Config.Source.TLS.BuildConfig()
+	if s.DBConfig.TLS != nil {
+		tlsConfig, err = s.DBConfig.TLS.BuildConfig()
 		if err != nil {
 			return err
 		}
 	}
 
-	if s.Config.MyServerId == 0 {
-		s.Config.MyServerId, err = s.generateNewServerId()
+	if s.MyServerId == 0 {
+		s.MyServerId, err = s.generateNewServerId()
 		if err != nil {
 			s.logger.WithError(err).Error("could not generate unique server_id")
 			return err
@@ -61,11 +62,11 @@ func (s *BinlogStreamer) createBinlogSyncer() error {
 	}
 
 	syncerConfig := replication.BinlogSyncerConfig{
-		ServerID:                s.Config.MyServerId,
-		Host:                    s.Config.Source.Host,
-		Port:                    s.Config.Source.Port,
-		User:                    s.Config.Source.User,
-		Password:                s.Config.Source.Pass,
+		ServerID:                s.MyServerId,
+		Host:                    s.DBConfig.Host,
+		Port:                    s.DBConfig.Port,
+		User:                    s.DBConfig.User,
+		Password:                s.DBConfig.Pass,
 		TLSConfig:               tlsConfig,
 		UseDecimal:              true,
 		TimestampStringLocation: time.UTC,
@@ -76,7 +77,9 @@ func (s *BinlogStreamer) createBinlogSyncer() error {
 }
 
 func (s *BinlogStreamer) ConnectBinlogStreamerToMysql() error {
-	currentPosition, err := ShowMasterStatusBinlogPosition(s.Db)
+	s.ensureLogger()
+
+	currentPosition, err := ShowMasterStatusBinlogPosition(s.DB)
 	if err != nil {
 		s.logger.WithError(err).Error("failed to read current binlog position")
 		return err
@@ -86,6 +89,8 @@ func (s *BinlogStreamer) ConnectBinlogStreamerToMysql() error {
 }
 
 func (s *BinlogStreamer) ConnectBinlogStreamerToMysqlFrom(startFromBinlogPosition mysql.Position) error {
+	s.ensureLogger()
+
 	err := s.createBinlogSyncer()
 	if err != nil {
 		return err
@@ -108,6 +113,8 @@ func (s *BinlogStreamer) ConnectBinlogStreamerToMysqlFrom(startFromBinlogPositio
 }
 
 func (s *BinlogStreamer) Run() {
+	s.ensureLogger()
+
 	defer func() {
 		s.logger.Info("exiting binlog streamer")
 		s.binlogSyncer.Close()
@@ -197,7 +204,7 @@ func (s *BinlogStreamer) FlushAndStop() {
 	// passed the initial target position.
 	err := WithRetries(100, 600*time.Millisecond, s.logger, "read current binlog position", func() error {
 		var err error
-		s.targetBinlogPosition, err = ShowMasterStatusBinlogPosition(s.Db)
+		s.targetBinlogPosition, err = ShowMasterStatusBinlogPosition(s.DB)
 		return err
 	})
 
@@ -295,7 +302,7 @@ func (s *BinlogStreamer) generateNewServerId() (uint32, error) {
 	for {
 		id = randomServerId()
 
-		exists, err := idExistsOnServer(id, s.Db)
+		exists, err := idExistsOnServer(id, s.DB)
 		if err != nil {
 			return 0, err
 		}
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -76,22 +76,27 @@ func (s *BinlogStreamer) createBinlogSyncer() error {
 }
 
 func (s *BinlogStreamer) ConnectBinlogStreamerToMysql() error {
-	err := s.createBinlogSyncer()
-	if err != nil {
-		return err
-	}
-
-	s.logger.Info("reading current binlog position")
-	s.lastStreamedBinlogPosition, err = ShowMasterStatusBinlogPosition(s.Db)
+	currentPosition, err := ShowMasterStatusBinlogPosition(s.Db)
 	if err != nil {
 		s.logger.WithError(err).Error("failed to read current binlog position")
 		return err
 	}
 
+	return s.ConnectBinlogStreamerToMysqlFrom(currentPosition)
+}
+
+func (s *BinlogStreamer) ConnectBinlogStreamerToMysqlFrom(startFromBinlogPosition mysql.Position) error {
+	err := s.createBinlogSyncer()
+	if err != nil {
+		return err
+	}
+
+	s.lastStreamedBinlogPosition = startFromBinlogPosition
+
 	s.logger.WithFields(logrus.Fields{
 		"file": s.lastStreamedBinlogPosition.Name,
 		"pos":  s.lastStreamedBinlogPosition.Pos,
-	}).Info("found binlog position, starting synchronization")
+	}).Info("starting binlog streaming")
 
 	s.binlogStreamer, err = s.binlogSyncer.StartSync(s.lastStreamedBinlogPosition)
 	if err != nil {
--- a/config.go
+++ b/config.go
@@ -237,6 +237,11 @@ type Config struct {
 	// Optional: defaults to false
 	AutomaticCutover bool
 
+	// This specifies whether or not Ferry.Run will handle SIGINT and SIGTERM
+	// by dumping the current state to stdout.
+	// This state can be used to resume Ghostferry.
+	DumpStateToStdoutOnSignal bool
+
 	// Config for the ControlServer
 	ServerBindAddr string
 	WebBasedir     string
--- a/config.go
+++ b/config.go
@@ -250,14 +250,6 @@ type Config struct {
 	// If this is null, a new Ghostferry run will be started. Otherwise, the
 	// reconciliation process will start and Ghostferry will resume after that.
 	StateToResumeFrom *SerializableState
-
-	// If GhostferryVersion within StateToResumeFrom mismatch with the actual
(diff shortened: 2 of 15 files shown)


ferry.go
```diff
@@ -372,14 +421,24 @@ func (f *Ferry) RunStandaloneDataCopy(tables []*schema.Table) error {
...
 
+	// HACK: This process is not interruptible, so we can override the previous
+	// StateTracker for this instance only as newDataIterator uses the Ferry's
+	// instance of StateTracker.
+	// BUG: if the PanicErrorHandler fires while running the standalone copy, we
+	// will get an error dump even though we should not get one, which could be
+	// misleading.
+	// TODO: refactor this to make it better, possibly have to refactor
+	// newDataIterator.
+	dataIterator.StateTracker = NewStateTracker(f.DataIterationConcurrency * 10)
```
shuhaowu: I don't like this line, but to do it properly would involve having to refactor newDataIterator and this method in general. This is a yakshave I don't want to undertake right now.

When we do refactor this, we should make the DataIterator/BinlogStreamer/BinlogWriter/BatchWriter easier to use as they're very difficult to initialize independently of Ferry (which is why we even have newDataIterator...).

sharding/testhelpers/unit_test_suite.go
```diff
@@ -71,6 +71,8 @@ func (t *ShardingUnitTestSuite) SetupTest() {
 	t.setupShardingFerry()
 	t.dropTestDbs()
 
+	testhelpers.SetupTest()
```
shuhaowu: I added this so we can get some debug log, which proved to be useful when I was trying to debug this implementation. I could remove it, tho.

binlog_streamer.go
```diff
@@ -35,37 +36,37 @@ type BinlogStreamer struct {
 	eventListeners []func([]DMLEvent) error
 }
 
-func (s *BinlogStreamer) Initialize() error {
```
fw42: What's the reasoning for removing the `Initialize` method?

config.go
```diff
@@ -237,9 +237,27 @@ type Config struct {
...
+	// reconciliation process will start and Ghostferry will resume after that.
+	StateToResumeFrom *SerializableState
+
+	// If GhostferryVersion within StateToResumeFrom mismatch with the actual
+	// Ghostferry version, the config validation will by default fail with an
+	// error. This boolean allows one to override that behaviour.
+	//
+	// Caution: there is no guarentee that setting this variable to true will
+	// be safe.
+	AllowResumeWithMismatchedGhostferryVersion bool
```
fw42: do we really need this?

cursor.go
```diff
@@ -40,22 +40,24 @@ type CursorConfig struct {
...
+		MaxPrimaryKey:            maxPk,
+		RowLock:                  true,
+		lastSuccessfulPrimaryKey: startPk,
 	}
 }
 
 // returns a new Cursor with an embedded copy of itself
-func (c *CursorConfig) NewCursorWithoutRowLock(table *schema.Table, maxPk uint64) *Cursor {
+func (c *CursorConfig) NewCursorWithoutRowLock(table *schema.Table, startPk, maxPk uint64) *Cursor {
 	return &Cursor{
```
fw42:
```golang
cursor := c.NewCursor(table, startPk, maxPk)
cursor.RowLock = false
return cursor
```
seems simpler maybe and more in line with what you are doing below with the `WithoutStateTracker` method?

cursor.go
```diff
@@ -72,7 +74,6 @@ type Cursor struct {
 }
 
 func (c *Cursor) Each(f func(*RowBatch) error) error {
-	c.lastSuccessfulPrimaryKey = 0
```
fw42: what was this here for to begin with?

data_iterator.go
```diff
@@ -25,20 +25,18 @@ type DataIterator struct {
 	logger         *logrus.Entry
 }
 
-func (d *DataIterator) Initialize() error {
+func (d *DataIterator) Run() {
 	d.logger = logrus.WithField("tag", "data_iterator")
 	d.targetPKs = &sync.Map{}
 
+	// If a state tracker is not provided, then the caller doesn't care about
+	// tracking state. However, some methods are still useful so we use initialize
```
fw42: `s/use//`?

ferry.go
```diff
@@ -93,8 +97,71 @@ func (f *Ferry) newDataIterator() (*DataIterator, error) {
...
+	return &BinlogStreamer{
+		DB:           f.SourceDB,
+		DBConfig:     f.Source,
+		MyServerId:   f.Config.MyServerId,
+		ErrorHandler: f.ErrorHandler,
+		Filter:       f.CopyFilter,
+	}
+}
+
+// Even tho this function is identical to NewBinlogStreamer, it is here
```
fw42: `s/tho/though/`

ferry.go
```diff
@@ -107,7 +174,14 @@ func (f *Ferry) Initialize() (err error) {
 
 	f.logger.Infof("hello world from %s", VersionString)
 
-	// Connect to the database
+	// Suppress siddontang/go-mysql logging as we already log the equivalents.
+	// It also by defaults logs to stdout, which is different from Ghostferry
+	// logging, which all goes to stderr. stdout in Ghostferry is reserved for
+	// dumping states due to an abort.
```
fw42: This seems a bit fragile. What if one of the other dependencies that we use logs to stdout? Or maybe a future one that we want to use does that and doesn't allow the user to disable it? What do we do then?

ferry.go
```diff
@@ -205,56 +283,16 @@ func (f *Ferry) Initialize() (err error) {
...
-	f.BatchWriter = &BatchWriter{
-		DB:           f.TargetDB,
-		StateTracker: f.StateTracker,
-
-		DatabaseRewrites: f.Config.DatabaseRewrites,
-		TableRewrites:    f.Config.TableRewrites,
-
-		WriteRetries: f.Config.DBWriteRetries,
+	if f.StateToResumeFrom == nil {
+		f.StateTracker = NewStateTracker(f.DataIterationConcurrency * 10)
```
fw42: What's this magical number 10 here?

state_tracker.go
```diff
@@ -44,6 +45,23 @@ func (s *StateTracker) UpdateLastSuccessfulPK(table string, pk uint64) {
 	s.updateSpeedLog(deltaPK)
 }
 
+func (s *StateTracker) LastSuccessfulPK(table string) uint64 {
+	s.tableMutex.RLock()
+	defer s.tableMutex.RUnlock()
+
+	_, found := s.completedTables[table]
+	if found {
+		return math.MaxUint64
```
fw42: Hmm this feels a bit like a hack to me. Is there no cleaner way to do this?

ferry.go
```diff
@@ -324,6 +371,26 @@ func (f *Ferry) Run() {
...
+
+			c := make(chan os.Signal, 1)
+			signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+
+			select {
+			case <-ctx.Done():
+				f.logger.Debug("shutting down signal monitoring goroutine")
+				return
+			case s := <-c:
+				f.ErrorHandler.Fatal("user", fmt.Errorf("signal received: %v", s.String()))
```
fw42: Hm where is this being handled? Where's the actual code that does the dumping to stdout?

shuhaowu: The ErrorHandler has always dumped the state inside the Fatal method. This code change just dumps it when a signal is received.

ferry.go
shuhaowu: There's a ring inside the StateTracker that tracks the speed at which we are copying data. The argument here specifies the ring's size. It should be a multiple of the number of goroutines that could be writing to the ring. 10 is just chosen as a conservative estimate.

binlog_streamer.go
shuhaowu: It's a legacy of the original Ghostferry design. This method doesn't do much and just complicates the API. Anything it does can be done directly in the `Run` method, or within the `Ferry.NewBinlogStreamer` methods.

cursor.go
shuhaowu: It was to ensure that subsequent calls to `.Each` would start from scratch. It turns out we don't actually use this method in this way.

ferry.go
shuhaowu: My current thinking on this is that stdout is the most flexible method to dump the state with the least amount of code within Ghostferry itself. 

If in the future we encounter issues with stdout, we can definitely make Ghostferry write to a file/pipe/whatever. We should deal with it when it comes up, especially since we don't quite have a vision on how to consume this state dump just yet.

state_tracker.go
shuhaowu: I could return an error here, but that feels somewhat untruthful. Alternatively, I could also return the last known successful primary key. This is somewhat misleading as the last known successful pk might be out of date and the table has grown since then. If a table is marked as completed by Ghostferry, it should not be processed again as it is assumed that any growth is copied over by the binlogstreamer. I thus decided to return MaxUint64 here.

It's not that big of a deal as there is only currently on consumer of this method in the DataIterator. The DataIterator already checks if a table is completed and this condition should never be hit in real code atm.

hkdsun: You could also return a boolean indicating completion of a table. WDYT?

ferry.go
```diff
@@ -278,7 +316,12 @@ func (f *Ferry) Start() error {
 	// miss some records that are inserted between the time the
 	// DataIterator determines the range of IDs to copy and the time that
 	// the starting binlog coordinates are determined.
-	err := f.BinlogStreamer.ConnectBinlogStreamerToMysql()
+	var err error
+	if f.StateToResumeFrom != nil {
```
hkdsun: i dont know how i feel about all these conditionals on `f.StateToResumeFrom == nil`. can we somehow boil it down to a single statement at a "higher" level?

config.go
```diff
@@ -255,6 +265,10 @@ func (c *Config) ValidateConfig() error {
 		return fmt.Errorf("Table filter function must be provided")
 	}
 
+	if c.StateToResumeFrom != nil && c.StateToResumeFrom.GhostferryVersion != VersionString {
+		return fmt.Errorf("StateToResumeFrom version mismatch: resume = %s, current = %s", c.StateToResumeFrom.GhostferryVersion, VersionString)
+	}
```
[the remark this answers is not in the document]
shuhaowu: Defensive for now as there's no guarantee that we won't change things. If we want to force resume, we can just set the version passed to Ghostferry to the same one as the Ghostferry instance that's launched.

data_iterator.go
```diff
@@ -76,7 +74,15 @@ func (d *DataIterator) Run() {
 					return
 				}
 
-				cursor := d.CursorConfig.NewCursor(table, targetPKInterface.(uint64))
+				startPk := d.StateTracker.LastSuccessfulPK(table.String())
```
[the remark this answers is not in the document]
shuhaowu: It's `db.table`. There are some issues if either your schema name or your table name includes a `. However, this problem is somewhat wide spread.

ferry.go
```diff
@@ -93,8 +97,71 @@ func (f *Ferry) newDataIterator() (*DataIterator, error) {
...
+
+func (f *Ferry) NewBinlogStreamer() *BinlogStreamer {
+	return &BinlogStreamer{
+		DB:           f.SourceDB,
+		DBConfig:     f.Source,
+		MyServerId:   f.Config.MyServerId,
+		ErrorHandler: f.ErrorHandler,
+		Filter:       f.CopyFilter,
+	}
+}
```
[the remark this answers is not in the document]
shuhaowu: I'm doing this because nothing external to Ferry is creating its own DataIterator/BinlogStreamer instances. I tried to have the alternative, but ended up initializing DataIterator/BinlogStreamer (for the delta copier and the reconciler) by copying all the attributes needed directly from Ferry. 

I figured it may be better to turn the Ferry into a factory. I feel like this line of thinking is more natural anyway as we have a global config object passed to Ferry, and Ferry is already the God object managing all the states.

ferry.go
```diff
@@ -93,8 +97,71 @@ func (f *Ferry) newDataIterator() (*DataIterator, error) {
...
+
+// Even though this function is identical to NewBinlogStreamer, it is here
+// for consistency so it will lead to less confusion.
```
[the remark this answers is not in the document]
shuhaowu: I'll remove this.

ferry.go
shuhaowu: Yeah we can. I'll have to cherry-pick some changes from a downstream branch that I have going on. It's not perfect but it's a bit better.

binlog_streamer.go
```diff
@@ -76,22 +77,31 @@ func (s *BinlogStreamer) createBinlogSyncer() error {
...
+func (s *BinlogStreamer) ConnectBinlogStreamerToMysqlFrom(startFromBinlogPosition mysql.Position) error {
+	s.ensureLogger()
+
+	err := s.createBinlogSyncer()
 	if err != nil {
-		s.logger.WithError(err).Error("failed to read current binlog position")
 		return err
 	}
 
+	s.lastStreamedBinlogPosition = startFromBinlogPosition
```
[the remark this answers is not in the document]
shuhaowu: Having it set there would be kind of redundant as this is an internal tracking variable for the main BinlogStreamer loop. Basically, it would have to look like:

```
ConnectBinlogStreamerToMySqlFrom():
  s.startFromBinlogPosition = startFromBinlogPosition

Run():
  s.lastStreamedBinlogPosition = s.startFromBinlogPosition
  for ... s.lastStreamedBinlogPosition.Compare(s.targetBinlogPosition) < 0:
     ....
```

Without setting lastStreamedBinlogPosition before we start that loop, the conditional would be more complex.

Note that this variable has nothing to do with the StateTracker binlog position, which is set in BinlogWriter and once in Ferry during initial connection.

ferry.go
shuhaowu: Hmm I can't easily cherry pick it back as I ended up moved the table loading logic from `Start` to `Initialize` and used the `StateTracker` instead of `StateToResumeFrom` directly in `Start` after adding in the Reconciler.

I prefer if we can leave this code as is, as the PRs coming up will change this section and hopefully will make this logic less confusing.

PR #68, opened 2018-10-23
Record binlog position in the StateTracker on BinlogStreamer connect

shuhaowu:
There's a race condition where if Ghostferry starts with some rows being copied but no binlog events have been streamed. If this is the case, then the binlog position recorded in the state tracker will be incorrect (as it will be zero). 

This PR fixes this issue.


ferry.go
```diff
@@ -278,11 +278,16 @@ func (f *Ferry) Start() error {
...
-	err := f.BinlogStreamer.ConnectBinlogStreamerToMysql()
+	pos, err := f.BinlogStreamer.ConnectBinlogStreamerToMysql()
 	if err != nil {
 		return err
 	}
 
+	// If we don't set this now, there is a race condition where Ghostferry
+	// is terminated with some rows copied but no binlog events are written.
+	// This guarentees that we are able to restart from a valid location.
+	f.StateTracker.UpdateLastWrittenBinlogPosition(pos)
```
hkdsun: hmm this doesn't feel right to me.. would a magic number like `-1` be more appropriate? 🤔

shuhaowu: This should be correct correct: Suppose if we start a run on a large database that has no traffic during the run. This means the binlog streamer will pick up no events during the run. If we abort while copying, the dumped binlog position will be 0 as it is uninitialized. Then when we restart, we have no idea where the last tailed binlog position and thus cannot "correctly" restart. If we restart at the position once Ghostferry comes up,  some writes could have occurred on the database when Ghostferry is down and we would not know about it.

Instead, we should record the position that Ghostferry last know about, which is the starting position in this case where no writes on the database occurred.

hkdsun: but what if there was a write at `pos` that the BinlogWriter didn't get a chance to successfully write at the time of interruption?

shuhaowu: It's OK for us to be out of date: if we miss an update to the StateTracker because we get interrupted, starting at an older coordinate doesn't have any adverse effects. When we resume, we'll just resume off this coordinate as oppose to not being able resume at all (as without this PR, no coordinates will be recorded if no writes occured).

shuhaowu: The other concern is if the position saved here will be picked up again once we resume. Upon resume, the binlog streamer will get the event corresponds to position `>=` saved position. This is not easily testable in this PR and will need to be done in a ruby interrupt/resume integration test, if possible at all.

hkdsun: I'm sorry I don't understand your explanation in the second comment. My comment was about ghostferry having written an event to the target database, not the source mysql having committed the event.

Why is it hard to write this unit test?

```
serializedState :=  {"pos": "1"}
bs := BinlogStreamerFromSerializedState(serializeState)
firstEventPos := bs.nextEvent().pos()
assertEqual(1, firstEventPos) // 
```

shuhaowu: Basically what I'm saying is that when we start Ghostferry without saved states, we'll get some position P when SHOW MASTER STATUS is called. Ghostferry will process all events where position > P.

The same position P will be saved to disk when we abort immediately or if no events are written to the source. When we restart, Ghostferry will process all events where position > P, identical to the normal case.

If Ghostferry wrote a binlog event > P before it had a chance to increment the last written binlog position before it was shutdown, P would be saved into the state dump. Then when Ghostferry restarts, it will process where pos > P. The already written event will be processed again. We already determined this to be fine as it is a requirement for interrupt/resume.

Issue #79, opened 2018-11-16
binlog_streamer get panic: ev.Header.Timestamp = 0


tclh123:
```
DEBU[8766] found 200 rows                                args="[1089172653]" sql="SELECT `id` FROM `db1`.`table1` WHERE `id` > ? ORDER BY `id` LIMIT 200" table=db1.table1 tag=cursor
DEBU[8766] found 200 rows                                args="[1089172906]" sql="SELECT `id` FROM `db1`.`table1` WHERE `id` > ? ORDER BY `id` LIMIT 200" table=db1.table1 tag=cursor
PANI[8766] logpos: 28594 0 *replication.GenericEvent     tag=binlog_streamer
INFO[8766] exiting binlog streamer                       tag=binlog_streamer
[2018/11/16 15:26:51] [info] binlogsyncer.go:163 syncer is closing...
DEBU[8766] found 200 rows                                args="[1089173124]" sql="SELECT `id` FROM `db1`.`table1` WHERE `id` > ? ORDER BY `id` LIMIT 200" table=db1.table1 tag=cursor
DEBU[8766] found 200 rows                                args="[1089173356]" sql="SELECT `id` FROM `db1`.`table1` WHERE `id` > ? ORDER BY `id` LIMIT 200" table=db1.table1 tag=cursor
DEBU[8767] found 200 rows                                args="[1089173568]" sql="SELECT `id` FROM `db1`.`table1` WHERE `id` > ? ORDER BY `id` LIMIT 200" table=db1.table1 tag=cursor
[2018/11/16 15:26:51] [error] binlogstreamer.go:57 close sync with err: sync is been closing...
[2018/11/16 15:26:51] [info] binlogsyncer.go:178 syncer is closed
panic: (*logrus.Entry) (0x8ccd40,0xc000be3b30)
goroutine 42 [running]:
github.com/Shopify/ghostferry/vendor/github.com/sirupsen/logrus.Entry.log(0xc00013a1e0, 0xc0001de4e0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/user1/go/src/github.com/Shopify/ghostferry/vendor/github.com/sirupsen/logrus/entry.go:128 +0x5a8
github.com/Shopify/ghostferry/vendor/github.com/sirupsen/logrus.(*Entry).Panic(0xc00013b090, 0xc000959ca8, 0x1, 0x1)
[...]

PR #80, opened 2018-11-16
skipping replication.GenericEvent

tclh123:
The go-mysql doesn't parse those events, so we can't handle this event.
This should fix the binlog_streamer panic https://github.com/Shopify/ghostferry/issues/79.

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -165,6 +165,10 @@ func (s *BinlogStreamer) Run() {
 			// This event can tell us about table structure change which means
 			// the cached schemas of the tables would be invalidated.
 			// TODO: investigate using this to allow for migrations to occur.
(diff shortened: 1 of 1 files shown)


binlog_streamer.go
```diff
@@ -165,6 +165,10 @@ func (s *BinlogStreamer) Run() {
 			// This event can tell us about table structure change which means
 			// the cached schemas of the tables would be invalidated.
 			// TODO: investigate using this to allow for migrations to occur.
+		case *replication.GenericEvent:
+			// go-mysql don't parse this event, it use GenericEvent instead,
```
shuhaowu: "go-mysql don't parse all events and unparsed events are denoted with empty GenericEvent structs"

PR #81, opened 2018-12-17
DO NOT LAND: Verbose slow shop mover to debug move failures

milanatshopify:
Getting more information when the move fails


milanatshopify:
I'm going to be testing with this.

PR #82, opened 2018-12-17
Log PK/binlog position on error

shuhaowu:
When Ghostferry's writes to the target fails, print out the PK value and the binlog positions into the error for better debugging.

### DIFF
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -30,8 +30,19 @@ func (w *BatchWriter) Initialize() {
 
 func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 	return WithRetries(w.WriteRetries, 0, w.logger, "write batch to target", func() error {
-		if batch.Size() == 0 {
+		values := batch.Values()
+		if len(values) == 0 {
 			return nil
+		}
+
+		startPkpos, err := values[0].GetUint64(batch.PkIndex())
+		if err != nil {
+			return err
+		}
+
+		endPkpos, err := values[len(values)-1].GetUint64(batch.PkIndex())
+		if err != nil {
+			return err
 		}
 
 		db := batch.TableSchema().Schema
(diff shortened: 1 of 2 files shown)


batch_writer.go
```diff
@@ -34,6 +34,17 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 			return nil
 		}
 
+		lastRow := batch.Values()[len(batch.Values())-1]
```
milanatshopify: batch.Size() is the same as len(batch.Values())? In that case, we're safe doing the -1, because it's guaranteed to be >= 0.

batch_writer.go
```diff
@@ -34,6 +34,17 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
...
+		startPkpos, err := batch.Values()[0].GetUint64(batch.PkIndex())
+		if err != nil {
+			return err
```
milanatshopify: I know this is just moved code, but we're not logging anything beyond the error, I guess we really don't expect this to fail?

shuhaowu: Yeah this really shouldn't fail as it was set by the dataiterator which should have the correct type.

batch_writer.go
hkdsun: correction: because it's guaranteed to be greater than zero (not greater or equal)

milanatshopify: The [ ... ] is guaranteed to be >= 0, we're not disagreeing here, we're just talking about different things.

shuhaowu: Yeah, `batch.Size()` above basically does the same thing. I refactored them to actually be same in this method.

batch_writer.go
```diff
@@ -30,10 +30,21 @@ func (w *BatchWriter) Initialize() {
 
 func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 	return WithRetries(w.WriteRetries, 0, w.logger, "write batch to target", func() error {
-		if batch.Size() == 0 {
+		values := batch.Values()
+		if len(values) == 0 {
```
hkdsun: Can we delete `Size()` then?

shuhaowu: It's still used in two locations in the `Cursor`. I don't want to change that for now to keep this change small.

PR #85, opened 2018-12-21
When a query fails, remove it from the map for a retry

milanatshopify:
Did I correctly interpret the conversation that we shouldn't cache and reuse when queries fail, or is that a different level cache?

### DIFF
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -62,11 +62,13 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 
 		stmt, err := w.stmtFor(query)
 		if err != nil {
+			w.delStmt(query)
 			return fmt.Errorf("during preparing query near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
 		}
 
 		_, err = stmt.Exec(args...)
 		if err != nil {
+			w.delStmt(query)
 			return fmt.Errorf("during exec query near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
 		}
 
@@ -109,4 +111,13 @@ func (w *BatchWriter) getStmt(query string) (*sql.Stmt, bool) {
 	defer w.mut.RUnlock()
 	stmt, exists := w.statements[query]
 	return stmt, exists
+}
+
+func (w *BatchWriter) delStmt(query string) {
+	w.mut.RLock()
+	defer w.mut.RUnlock()
+
+	if _, exists := w.statements[query]; exists {
+		delete(w.statements, query)
+	}
 }


batch_writer.go
```diff
@@ -110,3 +112,12 @@ func (w *BatchWriter) getStmt(query string) (*sql.Stmt, bool) {
 	stmt, exists := w.statements[query]
 	return stmt, exists
 }
+
+func (w *BatchWriter) delStmt(query string) {
+	w.mut.RLock()
+	defer w.mut.RUnlock()
```
hkdsun: `Lock()` and `Unlock()` since we're doing a write

batch_writer.go
```diff
@@ -110,3 +112,12 @@ func (w *BatchWriter) getStmt(query string) (*sql.Stmt, bool) {
...
+
+	if _, exists := w.statements[query]; exists {
```
hkdsun: No need for the `if`:

>func delete(m map[Type]Type1, key Type)
>
>The delete built-in function deletes the element with the specified key (m[key]) from the map. **If m is nil or there is no such element, delete is a no-op.**

milanatshopify:
Hormoz Kheradmand [14:53]
ah @Jeremy Cole are you saying that prepared statements are per-connection?
that would definitely cause issues

Jeremy Cole [14:54]
yes, they definitely are

Hormoz Kheradmand [14:54]
since we're caching the prepared statements and assuming they'll be valid even if connection gets reset


Jeremy Cole [14:54]
oh :open_mouth:
yes, the prepared statement handle is valid only for a single connection and only for the lifetime of it

Hormoz Kheradmand [14:56]
it's just not clear to me how to detect when the go mysql driver has made a new connection so that we can expire the cache
i guess our cache key can have the connection ID or something as part of it :slightly_smiling_face:
but still not clear why our connection gets killed in the first place consistently every time we move this shop :thinking_face:

hkdsun:
Ah i see. Yeah I think this is the correct thing to do.

shuhaowu:
I'm not sure if this will actually remove the statement from our cache if a connection is closed. According to the mysql go driver source code, [`Exec` on a statement will return `driver.ErrBadConn` to the `database/sql` package][1]. The [`database/sql` code seems to act on this error and tries to re-establish a new connection (?)][2] without notifying the user until a maximum number of attempts are made.

I think this effectively means what we want to do is occurring already behind the scenes inside `database/sql`. [Statements seems to be prepared for each connection that `database/sql` opens][3]. If I'm interpreting the library code correctly, does that not mean this patch is redundant?

[1]: https://github.com/go-sql-driver/mysql/blob/6be42e0ff99645d7d9626d779001a46e39c5f280/statement.go#L48-L51

[2]: https://github.com/golang/go/blob/9e277f7d554455e16ba3762541c53e9bfc1d8188/src/database/sql/sql.go#L2305-L2323

[3]: https://github.com/golang/go/blob/9e277f7d554455e16ba3762541c53e9bfc1d8188/src/database/sql/sql.go#L2377

pushrax:
You're interpreting it correctly, if prepared on a DB, the prepared statement code in database/sql handles everything to keep the prepared statement working as it moves across the connection pool. https://github.com/golang/go/pull/29890 improves the docs.

milanatshopify:
Sounds like we don't have to do anything here, the original worry was invalid.  Closing.

PR #88, opened 2019-01-28
Added a generic binlog catchup process and reconcile the binlogs

shuhaowu:
This PR adds a generic binlog catchup process prior to resuming Ghostferry (in `Ferry.CatchUpOnMissedBinlogsOnResume()`). Right now, all we reconcile is the binlog events: we delete any rows that are modified during the downtime of Ghostferry and recopy them from the source. In the future, we'll add mechanisms to reconcile the iterative verifier, and so on.

Note that there are no added integration tests for this PR because the current master code can already interrupt/resume by simply replaying all the binlog entries as is. The tests that covers that also covers this PR. This does mean this PR is not strictly necessary if we just want to interrupt/resume a run without handling schema changes. The reconciliation process becomes more important with a schema change as we agreed to delete the table whose schema changed and recopy it completely upon resume.


binlog_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+)
+
+// This event generates a REPLACE INTO SQL statement to overwrite a row if an
+// INSERT or UPDATE occured. For a DELETE statement, it generates the
+// corresponding DELETE statement based on the primary key.
+//
+// This is used during the resume when we are catching up to the binlog events
+// missed by Ghostferry while it is down.
+type ReconcilationDMLEvent struct {
+	*DMLEventBase
```
hkdsun: Why do you need to embed this? It doesn't make sense to call this struct a `*DMLEvent` either. It's not really an event itself. You may store the event that produced this as a field on the struct but calling this a DML event itself is very confusing

Is the only goal of this struct to provide a different `AsSQLString` so that it works as-is with the BinlogWriter? I feel like we need to refactor the current `BinlogWriter` to become a `MultiStatementWriter`.


```
type MultiStatementWriter struct{}  // current BinlogWriter implementation

type BinlogWriter struct{
  writer *MultiStatementWriter
}

type BinlogReconciler struct{
  writer *MultiStatementWriter
}
```

binlog_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+				batch = make([]DMLEvent, 0, r.BatchSize)
+			}
+
+			row, err := r.selectRowFromSource(table, pk)
+			if err != nil {
+				r.logger.WithError(err).Error("failed to select row from source")
+				return err
+			}
+
+			batch = append(batch, NewReconciliationDMLEvent(table, pk, row))
```
hkdsun: wait... why do you need `ReconcilationDMLEvent` at all? can't you re-use `RowBatch` here?

ferry.go
```diff
@@ -292,6 +292,102 @@ func (f *Ferry) Initialize() (err error) {
...
+
+		SourceDB:     f.SourceDB,
+		BinlogWriter: f.BinlogWriter,
+	}
+	binlogReconciler.Initialize()
+
+	// TODO: Step 2: setup the iterative verifier reconciler
+
+	// Step 3: initialize the binlog streamer and the throttler
+	binlogStreamer := f.NewBinlogStreamer()
```
hkdsun: wouldn't it be cleaner to have this inside the reconciler?

binlog_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+				r.logger.WithField("rows_replaced", count).Debug("replacing batch")
+				err := r.replaceBatch(batch)
+				if err != nil {
+					return err
+				}
+
+				batch = make([]DMLEvent, 0, r.BatchSize)
+			}
+
+			row, err := r.selectRowFromSource(table, pk)
```
hkdsun: selecting one row at a time? wouldn't that be very slow for high traffic dbs?

binlog_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+
+	r.logger.WithField("row_count", totalModifiedRows).Info("replaced modified rows")
+
+	return nil
+}
+
+func (r *BinlogReconciler) replaceBatch(batch []DMLEvent) error {
+	err := r.BinlogWriter.WriteEvents(batch)
```
hkdsun: these are not binlog events. see my comment earlier about `MultiStatementWriter`

ferry.go
```diff
@@ -292,6 +292,102 @@ func (f *Ferry) Initialize() (err error) {
...
+
+		err := f.Throttler.Run(ctx)
+		if err != nil {
+			f.ErrorHandler.Fatal("throttler", err)
+		}
+	}()
+	defer func() {
+		shutdown()
+		throttlerWg.Wait()
+	}()
```
hkdsun: why is this here? seems very out of place. smells like it should be started before we even get anywhere close to here

binlog_reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (r *BinlogReconciler) AddRowsToStore(events []DMLEvent) error {
+	for _, ev := range events {
+		pk, err := ev.PK()
+		if err != nil {
+			return err
+		}
+
+		r.modifiedRows.AddRow(ev.TableSchema(), pk)
```
hkdsun: do we really wanna `REPLACE` for every table? seems like the usual binlog playing logic should handle most tables?

ferry.go
shuhaowu: The BinlogReconciler is supposed to catch up binlogs events that occured while Ghostferry is down. However, if we want to interrupt/resume the verifier, the process of resuming that will also need an instance of the binlog streamer. Since we don't want two binlog streamers, this is managed directly in the Ferry.

binlog_reconciler.go
shuhaowu: I tried to refactor the BinlogWriter and it turned out to be a bit more difficult and tedious than I would have liked. Since the interface of the DMLEvent is mainly to call `AsSQLString`, I decided that it's easier to work with this version and refactor later.

shuhaowu:
> In the original ghostferry algorithm, what's the difference between starting the binlog streamer from the current position vs an earlier position? Assume there's no schema change.

The original Ghostferry algorithm is insensitive to any starting locations. As long as we don't miss any events, the data copied should be correct. 

If we don't start at the position of interruption and start at later point, we'll need something like the Reconciler. We would only really want to do this in the event of a schema change, because the events between the position of interruption and the position of resume could contain schema changes and thus change the binlog event structure.

> If we decide to go for full interruption, I would like us to not even have a Reconciler struct. Just feed the Ferry the last state and everything else happens automatically. Am I wrong to assume that's possible?

So I did this PR primarily to futureproof the algorithm. It is not strictly necessary if we don't want to handle schema changes for the time being. It might be possible to resume without the reconciler but I'll have to try it out to see how it looks, especially with respect to the iterative verifier.

PR #89, opened 2019-02-04
Moved the ownership of IterativeVerifier from sharding/copydb to Ferry

shuhaowu:
What is done 
----------

This PR moves the initialization and management of IterativeVerifier from the application level to the core library. This is required to interrupt and resume the IterativeVerifier as otherwise the application
layer will have to assume the responsibility of resuming the IterativeVerifier correctly, which is a non-trivial task.

This also removes the `.Tables` member on the DataIterator and made it an argument of `.Run`.

Also added tests in the ruby integration tests for the IterativeVerifier.

Implication for configurations
------------------------------

Since the verifiers will be initialized by the Ferry using configurations specified in Config. The Config interface is almost the same as the previous config, so a small change for sharding is needed:

Before for ghostferry-sharding:

```
{
 "IgnoredVerificationTables": [],
 "IgnoredVerificationColumns": [],
 "MaxExpectedVerifierDowntime": "1m",
 "VerifierIterationConcurrency": 4
}
```

After for ghostferry-sharding:

```
{
  "IterativeVerifierConfig": {
    "IgnoredTables": [],
    "IgnoredColumns": [],
    "MaxExpectedDowntime": "1m"
    "Concurrency": 4
  }
}
```

copydb doesn't require changes because previously it didn't allow for verifier configurations.

Future improvements
-------------------

One issue that exists is the table schema cache must be given to the verifiers after Ferry.Start. This causes some ugly code (Verifier.SetApplicableTableSchemaCache) for now. Once we refactor how the TableSchemaCache is handled, this method should go away.

Another issue is that the VerifyBeforeCutover is no longer controllable by the application level. The application level doesn't even really get an update about this development. This can be addressed in  the future.

### DIFF
--- a/compression_verifier.go
+++ b/compression_verifier.go
@@ -21,12 +21,10 @@ const (
 )
 
 type (
-	columnCompressionConfig map[string]string
-
 	// TableColumnCompressionConfig represents compression configuration for a
 	// column in a table as table -> column -> compression-type
 	// ex: books -> contents -> snappy
-	TableColumnCompressionConfig map[string]columnCompressionConfig
+	TableColumnCompressionConfig map[string]map[string]string
 )
 
 // UnsupportedCompressionError is used to identify errors resulting
--- a/data_iterator.go
+++ b/data_iterator.go
@@ -12,7 +12,6 @@ import (
 
 type DataIterator struct {
 	DB          *sql.DB
-	Tables      []*schema.Table
 	Concurrency int
 
 	ErrorHandler ErrorHandler
@@ -25,7 +24,7 @@ type DataIterator struct {
 	logger         *logrus.Entry
 }
 
-func (d *DataIterator) Run() {
+func (d *DataIterator) Run(tables []*schema.Table) {
 	d.logger = logrus.WithField("tag", "data_iterator")
 	d.targetPKs = &sync.Map{}
 
@@ -36,8 +35,8 @@ func (d *DataIterator) Run() {
 		d.StateTracker = NewCopyStateTracker(0)
 	}
 
-	d.logger.WithField("tablesCount", len(d.Tables)).Info("starting data iterator run")
-	tablesWithData, emptyTables, err := MaxPrimaryKeys(d.DB, d.Tables, d.logger)
+	d.logger.WithField("tablesCount", len(tables)).Info("starting data iterator run")
+	tablesWithData, emptyTables, err := MaxPrimaryKeys(d.DB, tables, d.logger)
 	if err != nil {
 		d.ErrorHandler.Fatal("data_iterator", err)
 	}
--- a/ferry.go
+++ b/ferry.go
@@ -344,7 +344,6 @@ func (f *Ferry) Start() error {
 
 	// TODO(pushrax): handle changes to schema during copying and clean this up.
 	f.BinlogStreamer.TableSchema = f.Tables
-	f.DataIterator.Tables = f.Tables.AsSlice()
 
 	return nil
 }
@@ -411,7 +410,7 @@ func (f *Ferry) Run() {
 
 	go func() {
 		defer dataIteratorWg.Done()
-		f.DataIterator.Run()
+		f.DataIterator.Run(f.Tables.AsSlice())
 	}()
 
 	dataIteratorWg.Wait()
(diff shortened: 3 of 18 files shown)


compression_verifier.go
```diff
@@ -21,12 +21,10 @@ const (
 )
 
 type (
-	columnCompressionConfig map[string]string
```
shuhaowu: Removed this type because it is only used once and it's also now a part of the public API.

ferry.go
```diff
@@ -288,6 +360,26 @@ func (f *Ferry) Initialize() (err error) {
 	f.DataIterator = f.NewDataIterator()
 	f.BatchWriter = f.NewBatchWriter()
 
+	if f.Config.VerifierType != "" {
+		if f.Verifier != nil {
+			return errors.New("VerifierType specified and Verifier is given. these are mutually exclusive options")
```
hkdsun: Why do we support both? Wouldn't it be simpler to have a single codepath?

data_iterator.go
```diff
@@ -25,7 +24,7 @@ type DataIterator struct {
 	logger         *logrus.Entry
 }
 
-func (d *DataIterator) Run() {
+func (d *DataIterator) Run(tables []*schema.Table) {
```
hkdsun: Why this change?

sharding/test/compression_verifier_test.go
```diff
@@ -67,6 +67,13 @@ func (t *CompressionVerifierTestSuite) InsertCompressedRowInDb(id int, schema, d
 	t.Require().Nil(err)
 }
 
+func (t *CompressionVerifierTestSuite) replaceCompressionVerifier(tableCompressions ghostferry.TableColumnCompressionConfig) {
+	var err error
+	iterativeVerifier := t.Ferry.Ferry.Verifier.(*ghostferry.IterativeVerifier)
+	iterativeVerifier.CompressionVerifier, err = ghostferry.NewCompressionVerifier(tableCompressions)
+	t.Require().Nil(err)
+}
```
hkdsun: This is..really ugly. Can you do any better with some more refactoring?

iterative_verifier.go
```diff
@@ -244,6 +245,11 @@ func (v *IterativeVerifier) VerifyBeforeCutover() error {
 	return err
 }
 
+func (v *IterativeVerifier) SetApplicableTableSchemaCache(t TableSchemaCache) {
+	v.Tables = t.AsSlice()
+	v.TableSchemaCache = t
+}
```
hkdsun: I still don't get why you're doing this instead of creating the verifier with the right fields

ferry.go
```diff
@@ -344,7 +436,12 @@ func (f *Ferry) Start() error {
 
 	// TODO(pushrax): handle changes to schema during copying and clean this up.
 	f.BinlogStreamer.TableSchema = f.Tables
-	f.DataIterator.Tables = f.Tables.AsSlice()
+
+	// TODO: refactor the TableSchemaCache logic to remove this weird post initialize
+	//       set tables hack
+	if f.Verifier != nil {
+		f.Verifier.SetApplicableTableSchemaCache(f.Tables)
+	}
```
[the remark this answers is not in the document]
shuhaowu: f.Tables is only loaded in `Start`. I tried to move this to initialize, but I have to change the integration tests quite a bit as the integration tests depends the tables being loaded in Start.

I started this work but I took it out because it confused this PR too much. I'll create another PR to fix this after merging this one.

ferry.go
shuhaowu: I only have this because `Verifier` is an `interface`. While we don't do it in our use cases, you could conceivably configure and pass in your own verifier to the Ferry without having the Ferry initialize it for you. You could even create an entirely different verifier without having to change the Ferry.

The code after this doesn't really change with respect to how the verifier is initialized, so the only "complexity" should be the few lines of code here.

data_iterator.go
shuhaowu: I wanted to refactor the table loading logic from `Start` to `Initialize`. I thought I took this part out, but I guess I didn't do it fully. 

As stated [earlier](https://github.com/Shopify/ghostferry/pull/89#discussion_r259859369), I'll refactor this part anyway, so I'm not going to remove it from this PR anymore.

iterative_verifier.go
shuhaowu: As stated [earlier](https://github.com/Shopify/ghostferry/pull/89#discussion_r259859369), the loading order for the TableSchemaCache is somewhat important to this. I want to get rid of this too but it proved too difficult for one PR. There'll be a follow up PR that removes this method.

sharding/test/compression_verifier_test.go
shuhaowu: I just tried and I can't do this without major rework of this test. At that point, it would be nice to be able to extract these tests out to the ruby tests, where we can more easily handle configuration changes.

Between this and the issue with table loading in the Go integration tests, we might want to think about refactoring all the tests into the ruby tests. However, I'm not sure if we'll have time to do this for now.

shuhaowu:
As stated, I'll have a follow up PR to cleanup the table loading logic. I'm uncertain about how we want to deal with the compression tests, but lets talk about it offline.

I'll merge this for now so we can have a new base to work on.

PR #90, opened 2019-02-05
Partially refactored StateTracker into two stages: copy and verify

shuhaowu:
The previous StateTracker can only handle tracking the copy, not the verify. This commit refactors the StateTracker into two distinct stages: the CopyStateTracker and the VerifierStateTracker, along with their
associated SerializableStates (CopySerializableState and VerifySerializableState).

The Ferry is responsible of initializing and passing the correct stages to the correct components. 

Right now, this PR only really implements the CopyStage. The VerifyStage will be implemented as I make the IterativeVerifier interruptible.

### DIFF
--- a/state_tracker.go
+++ b/state_tracker.go
@@ -19,7 +19,7 @@ import (
 // interrupted and resumed. The state tracker is supposed to be initialized and
 // managed by the Ferry. Each Ghostferry components, such as the `BatchWriter`
 // and `IterativeVerifier` will get passed an instance of either the
-// `CopyStateTracker` or the `VerifyStateTracker`. During the run, these
+// `CopyStateTracker` or the `VerificationStateTracker`. During the run, these
 // components will update their last successful components to the state tracker
 // instances given via the state tracker API defined here.
 //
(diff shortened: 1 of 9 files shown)


state_tracker.go
```diff
@@ -9,13 +9,118 @@ import (
 	"github.com/siddontang/go-mysql/mysql"
 )
 
+// StateTracker design
+// ===================
```
shuhaowu: I feel that it may be beneficial if we started leaving comments like this for all the major components. Thoughts?

state_tracker.go
```diff
@@ -9,13 +9,125 @@ import (
...
+}
+
+type VerifierSerializableState struct {
+	*BinlogAndIterationSerializableState
+
+	// This is not efficient because we have to build this map of a different
+	// type from the original ReverifyStore struct.
+	//
+	// TODO: address this inefficiency later.
+	ReverifyStore map[string][]uint64
```
shuhaowu: This is kind of a placeholder for now as I haven't hooked up the VerifierStage to the IterativeVerifier yet. We can iterate on this in the next PR.

hkdsun:
Almost looks good besides the suggested refactor. We're adding more complexity so let's reduce some debt too 👍

shuhaowu:
I've updated the changes such that the `BinlogAndIterationStateTracker` is split into `BinlogStateTracker` and `IterationStateTracker` per the suggestion above.

Issue #92, opened 2019-02-13
failed to read current binlog position 


ridvankaratas:
Hello,

I have an old MariaDB server and I wanted to use this tool for migration. Source MariaDB server was 10.0.x and binlog_row_image was not supported in this version. That's why I've upgraded it to 10.2.22 and this variable is working fine. 

After granting necessary permissions to the migration user for both source and target servers, I got this error: 

```
ERRO[0000] failed to read current binlog position        error="sql: expected 4 destination arguments in Scan, not 5" tag=binlog_streamer
error: failed to start ferry: sql: expected 4 destination arguments in Scan, not 5
```

I checked your binary codes and did so for the tutorial, but I couldn't find where exactly is locating for binlog position. 

Can you please advise how to fix it?

Thanks in advance.

shuhaowu:
Can you post the output of `SHOW MASTER STATUS` from your database? The code that's giving the error is here: https://github.com/Shopify/ghostferry/blob/908c310e1f4ea079c77c62090a12872e2721fb6f/utils.go#L146

PR #93, opened 2019-02-14
Scan SHOW MASTER STATUS row depends on column count #92

tufanbarisyildirim:
We are currently using MariaDB on production and `Executed_Gtid_Set` column is not returning for `SHOW MASTER STATUS` query
This is causing `sql: expected 4 destination arguments in Scan, not 5 ` when ghostferry wants t read binlog position

Here is our `SHOW MASTER STATUS` result

```
MariaDB [(none)]> SHOW MASTER STATUS;
+--------------------+----------+--------------+------------------+
| File               | Position | Binlog_Do_DB | Binlog_Ignore_DB |
+--------------------+----------+--------------+------------------+
| mariadb-bin.000006 | 39818670 |              |                  |
+--------------------+----------+--------------+------------------+
1 row in set (0.00 sec)
```

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -144,12 +144,28 @@ loop:
 }
 
 func ShowMasterStatusBinlogPosition(db *sql.DB) (mysql.Position, error) {
-	row := db.QueryRow("SHOW MASTER STATUS")
+	rows, err := db.Query("SHOW MASTER STATUS")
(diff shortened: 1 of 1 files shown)


tufanbarisyildirim:
will re-create that request. sorry.

PR #94, opened 2019-02-14
Scan SHOW MASTER STATUS row depending on column count #92

tufanbarisyildirim:
We are currently using MariaDB on production and `Executed_Gtid_Set` column is not returning for `SHOW MASTER STATUS` query
This is causing `sql: expected 4 destination arguments in Scan, not 5 ` when ghostferry wants t read binlog position

Here is our `SHOW MASTER STATUS` result

```
MariaDB [(none)]> SHOW MASTER STATUS;
+--------------------+----------+--------------+------------------+
| File               | Position | Binlog_Do_DB | Binlog_Ignore_DB |
+--------------------+----------+--------------+------------------+
| mariadb-bin.000006 | 39818670 |              |                  |
+--------------------+----------+--------------+------------------+
1 row in set (0.00 sec)
```

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -144,11 +144,20 @@ loop:
 }
 
 func ShowMasterStatusBinlogPosition(db *sql.DB) (mysql.Position, error) {
-	row := db.QueryRow("SHOW MASTER STATUS")
+	rows, err := db.Query("SHOW MASTER STATUS")
 	var file string
 	var position uint32
 	var binlog_do_db, binlog_ignore_db, executed_gtid_set string
-	err := row.Scan(&file, &position, &binlog_do_db, &binlog_ignore_db, &executed_gtid_set)
+	var cols []string
+	if rows.Next() {
+		cols, err = rows.Columns()
+		switch len(cols) {
+		case 4:
+			err = rows.Scan(&file, &position, &binlog_do_db, &binlog_ignore_db)
+		default:
+			err = rows.Scan(&file, &position, &binlog_do_db, &binlog_ignore_db, &executed_gtid_set)
+		}
+	}
 	return NewMysqlPosition(file, position, err)
 }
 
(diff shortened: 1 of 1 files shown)


utils.go
```diff
@@ -144,11 +144,20 @@ loop:
 }
 
 func ShowMasterStatusBinlogPosition(db *sql.DB) (mysql.Position, error) {
-	row := db.QueryRow("SHOW MASTER STATUS")
+	rows, err := db.Query("SHOW MASTER STATUS")
```
hkdsun: should return `if err != nil`

hkdsun: `defer rows.Close()` 🙏

utils.go
```diff
@@ -144,11 +144,20 @@ loop:
...
 	var file string
 	var position uint32
 	var binlog_do_db, binlog_ignore_db, executed_gtid_set string
-	err := row.Scan(&file, &position, &binlog_do_db, &binlog_ignore_db, &executed_gtid_set)
+	var cols []string
+	if rows.Next() {
+		cols, err = rows.Columns()
```
hkdsun: also return `if err != nil`

tufanbarisyildirim:
Thanks for review @hkdsun Updated.

hkdsun:
Looks like the tests are failing in CI with a nil pointer @tufanbarisyildirim 

https://travis-ci.com/Shopify/ghostferry/builds/100945932

hkdsun:
Thanks for your contribution Tufan. We'll wait for @shuhaowu to review and merge this PR 👍

utils.go
```diff
@@ -144,11 +144,29 @@ loop:
 }
 
 func ShowMasterStatusBinlogPosition(db *sql.DB) (mysql.Position, error) {
-	row := db.QueryRow("SHOW MASTER STATUS")
+	rows, err := db.Query("SHOW MASTER STATUS")
+	if rows != nil {
+		defer rows.Close()
```
shuhaowu: You should be able to have the `defer rows.Close()` line after `if err != nil {}` and there should never be a nil pointer, right?

tufanbarisyildirim: right! done.

PR #97, opened 2019-02-15
Find referenced tables if there is any foreign key to create them first.

### LINKED ISSUE #95: Foreign key constraint is incorrectly formed
When it comes to creating a table on the target server with Foreign Key and if referenced table hasn't created yet, it gives an error.

`cannot create table, this may leave the target database in an insane state  error="Error 1005: Can't 
create table publisher.View (errno: 150 \"Foreign key constraint is incorrectly formed\")" 
table=publisher.View error: failed to create databases and tables: Error 1005: Can't create table 
publisher.View (errno: 150 "Foreign key constraint is incorrectly formed")`

Example of a create Query for View Table:

> CREATE TABLE `View` (
>   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
>   `platform_id` int(10) unsigned NOT NULL,
>   `user_id` int(11) NOT NULL,
>   `object_id` int(10) unsigned NOT NULL,
>   `object_type` tinyint(4) NOT NULL,
>   `timestamp` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
>   PRIMARY KEY (`id`),
>   KEY `view_platform_id_foreign` (`platform_id`),
>   CONSTRAINT `view_platform_id_foreign` FOREIGN KEY (`platform_id`) REFERENCES `Platform` (`id`)
> ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci

In this case, `ghostfrerry-copydb` hasn't created `Platform` table yet. 

We have a fix for this too. Just putting this issue here for future reference.

tufanbarisyildirim:
This PR Fixes #95 ,
Needed to check foreign keys before creating a table and check / create referenced tables first from INFORMATION_SCHEMA. Also needed to create another map to keep tables that created in current session to prevent recreating tables that created due to reference before their turn. Will update that request with unit tests too. Just keeping here for any feedback and/or commits from you.

### DIFF
--- a/copydb/copydb.go
+++ b/copydb/copydb.go
@@ -2,7 +2,6 @@ package copydb
 
 import (
 	"fmt"
-	"os"
 	"strings"
 	"sync"
 	"time"
@@ -217,7 +216,7 @@ func (this *CopydbFerry) createTableOnTarget(database, table string) error {
 	fullTableName = fmt.Sprintf("%s.%s", database, table)
 
 	if _, created := this.CreatedTables[fullTableName]; created {
-		logrus.WithField("table", fullTableName).Warn("Table already created. Skipping")
+		logrus.WithField("table", fullTableName).Warn("Table already created in this session. Skipping")
 		return nil
 	}
 
@@ -274,8 +273,7 @@ func (this *CopydbFerry) createReferencedTablesIfAny(database, table string) err
 	stmt, err := this.Ferry.SourceDB.Prepare(constraintsSql)
 
 	if err != nil {
-		_, _ = fmt.Fprintf(os.Stderr, "error preparing constraints sql : %s\n", constraintsSql)
-		return err
+		return fmt.Errorf("error preparing constraints sql : %s", constraintsSql)
 	}
 
 	rows, err := stmt.Query(table, database)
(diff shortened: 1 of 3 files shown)


shuhaowu:
I don't think this would work. Even if you create the table, subsequent BinlogWriter and BatchWriter operations will surely fail due to not being able to satisfy FK constraints. The bug here is that we don't warn the users that Ghostferry doesn't support FK when it detects one on boot: #3.

tufanbarisyildirim:
> I don't think this would work. Even if you create the table, subsequent BinlogWriter and BatchWriter operations will surely fail due to not being able to satisfy FK constraints. The bug here is that we don't warn the users that Ghostferry doesn't support FK when it detects one on boot: #3.

That is why I'm keeping that merge open until we do a complete  migration using it. It needs some more work like checking parent existence before inserting a row on a table has a foreign key and rewriting referenced table name on creation etc. It won't work now but I think Ghostferry should support shouldn't it?

shuhaowu:
> Ghostferry should support shouldn't it?

I'm inclined to think Ghostferry shouldn't support this as it would be a very non-trivial change and there's no guarantee that there's even a solution. See https://github.com/github/gh-ost/issues/331 for more information (gh-ost and Ghostferry are similar enough in this case to have the arguments be applicable to both projects).

tufanbarisyildirim:
It definitely makes sense. I will go over again when I have enough time and back to here to close that PR or add some more comments to continue discussing. Thanks @shuhaowu

shuhaowu:
See #166

PR #98, opened 2019-02-16
Fixed JSON column with binlog streamer

### LINKED ISSUE #86: Error 3144: Cannot create a JSON value from a string with CHARACTER SET binary
Moving a shop  failed with
> Error 3144: Cannot create a JSON value from a string with CHARACTER SET binary
error.

shuhaowu:
JSON column was previously being recognized as a []byte array in the BinlogWriter. This causes the INSERT/UPDATE/DELETE query generated to be with something like `_binary'{"data": 1}'`. However, JSON data are not binary, thus MySQL will return errors  errors such as:

> Error 3144: Cannot create a JSON value from a string with CHARACTER SET binary error.

Fixing this requires us to pass the column types into the methods actually escaping the values (`appendEscapedValues`), as we cannot infer that something is a JSON type from the Go type alone. This turned out to be a rather large list of methods.

To do this the most generic way, I just pass down the table columns all the way (in case we need something else from it in the future).

Fixes #86.

### DIFF
--- a/dml_events.go
+++ b/dml_events.go
@@ -97,15 +97,14 @@ func (e *BinlogInsertEvent) NewValues() RowData {
 }
 
 func (e *BinlogInsertEvent) AsSQLString(target *schema.Table) (string, error) {
-	columns, err := loadColumnsForTable(&e.table, e.newValues)
-	if err != nil {
+	if err := verifyValuesHasTheSameLengthAsColumns(&e.table, e.newValues); err != nil {
 		return "", err
 	}
 
 	query := "INSERT IGNORE INTO " +
 		QuotedTableNameFromString(target.Schema, target.Name) +
-		" (" + strings.Join(columns, ",") + ")" +
-		" VALUES (" + buildStringListForValues(e.newValues) + ")"
+		" (" + strings.Join(quotedColumnNames(&e.table), ",") + ")" +
+		" VALUES (" + buildStringListForValues(e.table.Columns, e.newValues) + ")"
 
 	return query, nil
 }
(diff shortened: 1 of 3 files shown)


dml_events.go
```diff
@@ -253,84 +250,77 @@ func NewBinlogDMLEvents(table *schema.Table, ev *replication.BinlogEvent, pos my
 	}
 }
 
-func loadColumnsForTable(table *schema.Table, valuesToVerify ...RowData) ([]string, error) {
```
shuhaowu: Since I'm passing down `[]TableColumn` anyway, I might as well just do the column name quoting in the (`build*` methods that actually builds the query).

This method is a bit vague on what it does anyway, so I removed it.

dml_events.go
```diff
@@ -253,84 +250,77 @@ func NewBinlogDMLEvents(table *schema.Table, ev *replication.BinlogEvent, pos my
...
-	for _, values := range valuesToVerify {
-		if err := verifyValuesHasTheSameLengthAsColumns(table, values); err != nil {
```
shuhaowu: This method is now instead called manually in the `AsSQLQuery` methods directly.

dml_events.go
```diff
@@ -433,8 +423,12 @@ func appendEscapedString(buffer []byte, value string) []byte {
...
 }
 
-func appendEscapedBuffer(buffer, value []byte) []byte {
-	buffer = append(buffer, "_binary'"...)
+func appendEscapedBuffer(buffer, value []byte, binary bool) []byte {
+	if binary {
+		buffer = append(buffer, "_binary"...)
+	}
+
+	buffer = append(buffer, '\'')
```
milanatshopify: I'm definitely stretching my ability to review this, but why do we need this extra ' here?

shuhaowu: This was in there previously as `_binary'`, as we still need to quote the data. It is just split now depending if you want the binary encoding or not.

test/integration/types_test.rb
```diff
@@ -0,0 +1,43 @@
+require "test_helper"
+
+class TrivialIntegrationTests < GhostferryTestCase
```
hkdsun: `TypesTest`

shuhaowu: Good catch. Thanks.

PR #100, opened 2019-02-25
Moved TableSchemaCache loading from Ferry.Start to Ferry.Initialize

shuhaowu:
We want to do this because having the TableSchemaCache be first loaded in Ferry.Start means we have to then set the instance variables of BinlogStreamer and the verifiers after it is loaded. This effectively means the initializing of the BinlogStreamer and verifiers are split between Initialize and Start, an ugly situation.

This is a surprisingly complicated task, as there is a dependency within the tests: 

1. The tests uses Ferry.SourceDB to seed the database.
2. Ferry.SourceDB is created in Ferry.Initialize()
3. Once f.Tables is populated during Ferry.Initialize(), it means all the tests will see an empty table schema cache.
4. Seeing an empty table schema cache will cause failures.

Instead, what we have to do is to create separate sql.DB instances for the tests to connect to the source and target db without using Ferry. All the seeding logic can thus be moved prior to Ferry.Initialize. However, this results in a rather large PR (mostly one line changing ferry.SourceDB -> testSuite.SourceDB)

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -121,6 +121,7 @@ func (f *Ferry) NewBinlogStreamer() *BinlogStreamer {
 		MyServerId:   f.Config.MyServerId,
 		ErrorHandler: f.ErrorHandler,
 		Filter:       f.CopyFilter,
+		TableSchema:  f.Tables,
 	}
 }
 
@@ -351,8 +352,21 @@ func (f *Ferry) Initialize() (err error) {
 
 	if f.StateToResumeFrom == nil {
 		f.StateTracker = NewStateTracker(f.DataIterationConcurrency * 10)
+
+		// Loads the schema of the tables that are applicable.
+		// We need to do this at the beginning of the run as this is required
+		// in order to determine the PrimaryKey of each table as well as finding
+		// which value in the binlog event correspond to which field in the
+		// table.
+		metrics.Measure("LoadTables", nil, 1.0, func() {
+			f.Tables, err = LoadTables(f.SourceDB, f.TableFilter)
+		})
+		if err != nil {
+			return err
+		}
 	} else {
 		f.StateTracker = NewStateTrackerFromSerializedState(f.DataIterationConcurrency*10, f.StateToResumeFrom)
+		f.Tables = f.StateToResumeFrom.LastKnownTableSchemaCache
 	}
 
 	f.BinlogStreamer = f.NewBinlogStreamer()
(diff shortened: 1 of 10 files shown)


shuhaowu:
Can I get a review on this soon? The change is mostly affecting the Go based integration tests, which I hope to deprecate in favour of the Ruby version for the future.

ferry.go
```diff
@@ -215,6 +217,8 @@ func (f *Ferry) NewIterativeVerifier() (*IterativeVerifier, error) {
 		TargetDB:            f.TargetDB,
 		CompressionVerifier: compressionVerifier,
 
+		Tables:              f.Tables.AsSlice(),
+		TableSchemaCache:    f.Tables,
```
hkdsun: It makes me uneasy to be suggesting this but do you think it's a good idea to return an error if one of these `Ferry.NewThing()` constructors are called before the Ferry is initialized? (perhaps checked with a `Ferry.initialized` boolean field)? 

The ultimate cleanup would be that we get rid of `Ferry.Initialize` in favour of a `NewFerry()` constructor but that's long ways away from our current state

shuhaowu: the `Ferry.NewComponent()` methods, except for the IterativeVerifier, only returns the component structs being created. Adding an error would make the methods more annoying to use.

What if I just `panic` if the Ferry is uninitialized and these methods are called? If such a code was written, it's guarenteed to be a programming error as the resulting components won't even have a valid `SourceDB` and such. This would basically match the current behaviour: components created before Ferry is initialized will likely panic due to nil pointers.


(I agree with the spirit of `NewFerry`, but I don't know how to create a sane version of it without a large amount of positional arguments.)

hkdsun: the `panic` is a good enough measure for now. Add a `TODO: refactor Ferry.Initialize to NewFerry constructor` right by it and we're good to go

ferry.go
```diff
@@ -418,31 +450,6 @@ func (f *Ferry) Start() error {
...
-		metrics.Measure("LoadTables", nil, 1.0, func() {
-			f.Tables, err = LoadTables(f.SourceDB, f.TableFilter)
-		})
-		if err != nil {
-			return err
-		}
-	}
-
-	// TODO(pushrax): handle changes to schema during copying and clean this up.
-	f.BinlogStreamer.TableSchema = f.Tables
```
fjordan: Do we not need to set the TableSchema on the BinlogStreamer in `Initialize` now?

shuhaowu: We do. It occurs directly in NewBinlogStreamer as opposed to within Initialize in an adhoc way.

fjordan: I see it above now. Thanks 👍

Issue #101, opened 2019-03-01
Attempt to Retry Cutover Locks


fjordan:
Currently, Ghostferry will only attempt to acquire the cutover lock a single time before failing. We want to modify Ghostferry to retry acquiring the cutover lock, first using a static value, and then eventually using a dynamic value sent back that tells Ghostferry for how long to wait before attempting to acquire the lock again.

PR #102, opened 2019-03-01
Retry acquisition of cutover lock on error

fjordan:
First step of https://github.com/Shopify/ghostferry/issues/101. 

With this change, the ShardingFerry will now attempt to retry acquiring the cutover lock a configurable number of times while also waiting a configurable number of seconds between each attempt. Later, we can add a "retry-after" header or similar to tell the ferry to wait a dynamic period of time.

### DIFF
--- a/sharding/config.go
+++ b/sharding/config.go
@@ -26,4 +26,22 @@ type Config struct {
 	PrimaryKeyTables []string
 
 	Throttle *ghostferry.LagThrottlerConfig
+
+	// These two values configure the amount of times Ferry should attempt to
+	// retry acquiring the cutover lock, and for how long the Ferry should wait
+	// before attempting another lock acquisition
+	MaxCutoverRetries       int
+	CutoverRetryWaitSeconds int
+}
+
+func (c *Config) ValidateConfig() error {
+	if c.MaxCutoverRetries == 0 {
+		c.MaxCutoverRetries = 1
+	}
+
+	if c.CutoverRetryWaitSeconds == 0 {
+		c.CutoverRetryWaitSeconds = 1
+	}
+
+	return c.Config.ValidateConfig()
 }
--- a/sharding/sharding.go
+++ b/sharding/sharding.go
@@ -113,14 +113,20 @@ func (r *ShardingFerry) Run() {
 
 	r.AbortIfTargetDbNoLongerWriteable()
 
-	var err error
-	client := &http.Client{}
+	var (
+		err          error
+		client       http.Client
+		cutoverStart time.Time
+	)
 
-	cutoverStart := time.Now()
 	// The callback must ensure that all in-flight transactions are complete and
 	// there will be no more writes to the database after it returns.
-	metrics.Measure("CutoverLock", nil, 1.0, func() {
-		err = r.config.CutoverLock.Post(client)
+	err = ghostferry.WithRetries(r.config.MaxCutoverRetries, time.Duration(r.config.CutoverRetryWaitSeconds)*time.Second, r.logger, "get cutover lock", func() (err error) {
+		metrics.Measure("CutoverLock", nil, 1.0, func() {
+			cutoverStart = time.Now()
+			err = r.config.CutoverLock.Post(&client)
+		})
+		return err
 	})
 	if err != nil {
 		r.logger.WithField("error", err).Errorf("locking failed, aborting run")
@@ -167,7 +173,7 @@ func (r *ShardingFerry) Run() {
 	r.Ferry.Throttler.SetDisabled(false)
 
 	metrics.Measure("CutoverUnlock", nil, 1.0, func() {
-		err = r.config.CutoverUnlock.Post(client)
+		err = r.config.CutoverUnlock.Post(&client)
 	})
 	if err != nil {
 		r.logger.WithField("error", err).Errorf("unlocking failed, aborting run")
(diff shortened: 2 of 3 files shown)


config.go
```diff
@@ -278,6 +278,12 @@ type Config struct {
 	// Optional: defaults to false
 	AutomaticCutover bool
 
+	// These two values configure the amount of times Ferry should attempt to
```
milanatshopify: What are the defaults, and where are they set? And what happens if they're not set?

fjordan: We need to default to `1` as a value of `0` retries will retry indefinitely.

fjordan: This is defaulted to `1` below in the `ValidateConfig` method

sharding/sharding.go
```diff
@@ -119,8 +119,11 @@ func (r *ShardingFerry) Run() {
 	cutoverStart := time.Now()
```
hkdsun: How can we accurately report downtime now? This is going to include all retries which is not great

sharding/sharding.go
```diff
@@ -119,8 +119,11 @@ func (r *ShardingFerry) Run() {
...
 	// The callback must ensure that all in-flight transactions are complete and
 	// there will be no more writes to the database after it returns.
-	metrics.Measure("CutoverLock", nil, 1.0, func() {
-		err = r.config.CutoverLock.Post(client)
+	err = ghostferry.WithRetries(r.config.MaxCutoverRetries, time.Duration(r.config.CutoverRetryWaitSeconds)*time.Second, r.logger, "get cutover lock", func() (err error) {
```
hkdsun: `time.Duration(r.config.CutoverRetryWaitSeconds)*time.Second` is confusing. 

`time.Duration` represents nano seconds but it _happens_ to also be an `int64` which works out at the end here. 

I believe what you want is `r.config.CutoverRetryWaitSeconds*time.Second`

fjordan: We need to cast it to `time.Duration` before attempting the operation as this change gives:
```sh
# github.com/Shopify/ghostferry/sharding
./sharding.go:122:91: invalid operation: r.config.Config.CutoverRetryWaitSeconds * time.Second (mismatched types int and time.Duration)
```

sharding/sharding.go
fjordan: Good catch. Would moving this line into the callback func below (after line 122) with a var declaration outside of the block suffice as it will reset with each `POST`?

sharding/sharding.go
hkdsun: ah i see, 👍

sharding/sharding.go
hkdsun: I'm not sure what end result we want to see on our dashboard, I defer to @milanatshopify

config.go
```diff
@@ -278,6 +278,12 @@ type Config struct {
...
+	// retry acquiring the cutover lock, and for how long the Ferry should wait
+	// before attempting another lock acquisition
+	MaxCutoverRetries       int
+	CutoverRetryWaitSeconds int
```
shuhaowu: These variables should be in sharding/config.go because they don't pertain to the base Ferry.

sharding/sharding.go
```diff
@@ -119,8 +119,11 @@ func (r *ShardingFerry) Run() {
...
+		metrics.Measure("CutoverLock", nil, 1.0, func() {
```
shuhaowu: Just a question: the metric measurement should still be on the cutover lock as opposed to surround the whole `WithRetries`?

config.go
fjordan: Is there a `ValidateConfig` method we can use to default these values in the sharding config? If not, any reason we shouldn't add one?

sharding/sharding.go
milanatshopify: We want to see the total time the shop was down during the shop move. That's probably a sum of individual ones, as that is relevant for the customers, though I can see some value in emitting multiple values.

sharding/sharding.go
milanatshopify: Right, likely summing up the individual cutover lock times, so that we get the total impact, without counting the sleeps between retries.

config.go
shuhaowu: We should definitely initialize default values for some config options somewhere. I don't recall off the top of my head where.

fjordan: I added another `ValidateConfig` to the sharding ferry's configuration that checks these two values and sets proper defaults and then calls the embedded struct's (ferry config) `ValidateConfig` method.

sharding/sharding.go
```diff
@@ -167,7 +173,7 @@ func (r *ShardingFerry) Run() {
 	r.Ferry.Throttler.SetDisabled(false)
 
 	metrics.Measure("CutoverUnlock", nil, 1.0, func() {
-		err = r.config.CutoverUnlock.Post(client)
+		err = r.config.CutoverUnlock.Post(&client)
```
shuhaowu: Any reason the type changed?

sharding/config.go
```diff
@@ -26,4 +26,22 @@ type Config struct {
...
 	Throttle *ghostferry.LagThrottlerConfig
+
+	// These two values configure the amount of times Ferry should attempt to
+	// retry acquiring the cutover lock, and for how long the Ferry should wait
+	// before attempting another lock acquisition
+	MaxCutoverRetries       int
+	CutoverRetryWaitSeconds int
+}
+
+func (c *Config) ValidateConfig() error {
```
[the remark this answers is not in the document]
fjordan: That could make sense, sure, but for the sake of consistency we should do the same thing in each package. I was just following convention with another `ValidateConfig` method.

sharding/sharding.go
fjordan: This was changed from
```go
client := &http.Client{}
```
to a var declaration block
```go
var (
	err          error
	client       http.Client
	cutoverStart time.Time
)
```

PR #103, opened 2019-03-11
Hook up interrupt/resume to copydb and sharding

shuhaowu:
A couple of minor things done in this PR:

- Added StateDump to the Error HTTP callback so it can be posted during the ghostferry-sharding usage.
- Changed the flag DumpStateToStdoutOnSignal to DumpStateOnSignal to reflect the fact that we dump the state to not just stdout but also via HTTP.
- Added the ability to pass in a state file to CopyDB.
- Also for CopyDB: Fix the SIGTERM/SIGINT going no where when Ferry.Run finishes but the control server is still up.

I'm doing the CopyDB changes because there's a big chance that we might need to use interrupt/resume with copydb in the next couple of weeks.

### DIFF
--- a/config.go
+++ b/config.go
@@ -279,9 +279,9 @@ type Config struct {
 	AutomaticCutover bool
 
 	// This specifies whether or not Ferry.Run will handle SIGINT and SIGTERM
-	// by dumping the current state to stdout.
-	// This state can be used to resume Ghostferry.
-	DumpStateToStdoutOnSignal bool
+	// by dumping the current state to stdout and the error HTTP callback.
+	// The dumped state can be used to resume Ghostferry.
+	DumpStateOnSignal bool
 
 	// Config for the ControlServer
 	ServerBindAddr string
--- a/copydb/copydb.go
+++ b/copydb/copydb.go
@@ -2,11 +2,8 @@ package copydb
 
 import (
 	"fmt"
-	"os"
-	"os/signal"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
 	"github.com/Shopify/ghostferry"
@@ -114,17 +111,6 @@ func (this *CopydbFerry) Run() {
 
 	// This is where you cutover from using the source database to
 	// using the target database.
-
-	// We have to listen to signals because the Ghostferry builtin signal handler
-	// has quit but this process is still alive
-	go func() {
-		c := make(chan os.Signal, 1)
-		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
-
-		<-c
-		os.Exit(0)
-	}()
-
 	logrus.Info("ghostferry main operations has terminated but the control server remains online")
 	logrus.Info("press CTRL+C or send an interrupt to stop the control server and end this process")
 
--- a/copydb/copydb.go
+++ b/copydb/copydb.go
@@ -2,8 +2,11 @@ package copydb
 
 import (
 	"fmt"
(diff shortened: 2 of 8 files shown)


sharding/test/callbacks_test.go
```diff
@@ -69,12 +69,12 @@ func (t *CallbacksTestSuite) TestFailsRunOnPanicError() {
...
-		errorData["ErrMessage"] = "test error"
-
-		errorDataBytes, jsonErr := json.MarshalIndent(errorData, "", "  ")
+		jsonErr := json.Unmarshal([]byte(resp["Payload"]), &errorData)
 		t.Require().Nil(jsonErr)
-		t.Require().Equal(string(errorDataBytes), resp["Payload"])
+
+		t.Require().Equal("test_error", errorData["ErrFrom"])
+		t.Require().Equal("test error", errorData["ErrMessage"])
+		t.Require().True(errorData["StateDump"] != "")
```
shuhaowu: This is not the best test for now, but the ruby integration test this more thoroughly anyway.

hkdsun: You sure about this? Why don't you assert some keys that we expect to exist do actually exist and have some value in them?

copydb/copydb.go
```diff
@@ -109,11 +112,22 @@ func (this *CopydbFerry) Run() {
...
-	logrus.Info("ghostferry main operations has terminated but the control server remains online")
-	logrus.Info("press CTRL+C or send an interrupt to stop the control server and end this process")
 	// This is where you cutover from using the source database to
 	// using the target database.
 
+	// We have to listen to signals because the Ghostferry builtin signal handler
+	// has quit but this process is still alive
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
```
hkdsun: Does this not override the ghostferry signal handlers? Is there some gopher magic to register multiple signal handlers?

config.go
```diff
@@ -279,9 +279,9 @@ type Config struct {
 	AutomaticCutover bool
 
 	// This specifies whether or not Ferry.Run will handle SIGINT and SIGTERM
-	// by dumping the current state to stdout.
-	// This state can be used to resume Ghostferry.
-	DumpStateToStdoutOnSignal bool
+	// by dumping the current state to stdout and the error HTTP callback.
+	// The dumped state can be used to resume Ghostferry.
+	DumpStateOnSignal bool
```
hkdsun: It still feels like dumping to stdout is all wrong. 

This will spam logs for the mover and we will get asked about it if there's a large number of moves being interrupted all the time.

Don't have to block but be forewarned 🙂

hkdsun:
There weren't significant changes to `sharding`. I assume that wasn't by accident 👍

config.go
shuhaowu: For this concern we can just redirect stdout to /dev/null if needed. stderr contains all the useful logs.

sharding/test/callbacks_test.go
shuhaowu: Changed it to check the state dump more thoroughly.

copydb/copydb.go
shuhaowu: I changed the base signal handler so it doesn't quit after Ferry.Run is finished. Based on if the context in Ferry.Run is canceled or not, it will either exit the process normally or panic.

PR #106, opened 2019-03-18
Fix state dump not showing due to concurrent Fatal

shuhaowu:
If multiple Fatal calls to the ErrorHandler is received from multiple goroutines, the current implementation will only dump the state once using an atomic integer.  Subsequent calls will not dump the state anymore. After dumping the state, panic is called, ending the process.

However, since at some point in the past, the method Fatal was refactored such that it first calls ReportError, which dumps the states, then it panics. However, the atomic integer checking logic is within ReportError. This means, if there are multiple racing goroutines calling .Fatal, there is a chance that the state dump is either incomplete or non-existent.

This PR fixes this issue by restoring the behaviour prior to the Fatal method split by returning a boolean from ReportError, letting Fatal know if it should panic.

### DIFF
--- a/error_handler.go
+++ b/error_handler.go
@@ -11,7 +11,9 @@ import (
 )
 
 type ErrorHandler interface {
-	ReportError(from string, err error)
+	// Usually called from Fatal. When called from Fatal, if this method returns
+	// true, Fatal should panic, otherwise it should not.
+	ReportError(from string, err error) bool
 	Fatal(from string, err error)
 }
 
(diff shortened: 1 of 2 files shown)


fjordan:
The change LGTM and makes sense, but are there any tests we can add to ensure the change behaves as intended?

shuhaowu:
I can look into adding some tests in the next PR. For the time being the code is straightforward enough that you can see it is correct intuitively.

Let me ship this to get some downstream stuff working first, then I can add some tests about if the error handler dumps states correctly and what not.

shuhaowu:
I can look into adding some tests in the next PR. For the time being the code is straightforward enough that you can see it is correct intuitively.

Let me ship this to get some downstream stuff working first, then I can add some tests about if the error handler dumps states correctly and what not.

shuhaowu:
By up a method do you mean `Fatal`? That would work as well. We do call `ReportError` at places without calling `Fatal` and I suppose sending multiple error callbacks is not so bad.

I'll make the change in a follow up now.

PR #108, opened 2019-03-25
Resume Ghostferry without LastKnownSchemaCache

shuhaowu:
If LastKnownSchemaCache is nil, Ghostferry resumes by first loading the tables on the source database.

This is useful if one doesn't want to save the LastKnownSchemaCache, which could be many megabytes of data.

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -367,12 +367,20 @@ func (f *Ferry) Initialize() (err error) {
 
 	if f.StateToResumeFrom == nil {
 		f.StateTracker = NewStateTracker(f.DataIterationConcurrency * 10)
+	} else {
+		f.StateTracker = NewStateTrackerFromSerializedState(f.DataIterationConcurrency*10, f.StateToResumeFrom)
+	}
 
-		// Loads the schema of the tables that are applicable.
-		// We need to do this at the beginning of the run as this is required
-		// in order to determine the PrimaryKey of each table as well as finding
-		// which value in the binlog event correspond to which field in the
-		// table.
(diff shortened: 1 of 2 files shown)


ferry.go
```diff
@@ -367,20 +367,27 @@ func (f *Ferry) Initialize() (err error) {
...
-		// which value in the binlog event correspond to which field in the
-		// table.
+	// Loads the schema of the tables that are applicable.
+	// We need to do this at the beginning of the run as this is required
+	// in order to determine the PrimaryKey of each table as well as finding
+	// which value in the binlog event correspond to which field in the
+	// table.
+	//
+	// If this is a resuming run and the last known table schema cache is not given
+	// we'll regenerate it from the source database, assuming it has not been
```
fjordan: > assuming it has not been changed

Is there any way we can fingerprint the schema and just throw an error if we identify a change?

shuhaowu: Not easily now. In the longish term I want to replace the TableSchemaCache from a simple map to something more manageable (and therefore serializable). We can do this then.

hkdsun: If we make the mover interruptible as-is, it means we only compare schema fingerprints before and after each run rather than before the first run and at the end of the final run.

We either have to persist the schema fingerprint from the first run somewhere and assert it doesn't change between interruptions or throw an error here if we identify a change, like Forrest suggested.

Do I understand this correctly?

shuhaowu: You're understanding this correctly. I'm already building in the functionality downstream to compare schema fingerprints before and after each initial/resuming run.

PR #109, opened 2019-03-28
Moved error handler atomic check to Fatal

shuhaowu:
Follow up to #106: moved atomic check from `ReportError` to `Fatal` to avoid returning a boolean on `ReportError`.


fjordan:
Still not a fan of the `panic`s everywhere and would prefer to just return errors to the main thread, which can then `panic`.

shuhaowu:
> return errors to the main thread

I tried this with a context based approach. I couldn't get a solution to work without deadlocking issues. The code also become even more unmanageable as context has to be manually managed everywhere. See #62. Perhaps there's a better solution out there that I didn't think of.

PR #112, opened 2019-05-16
Fix getting uint pk error in iterative verifier

everpcpc:
go-mysql support uint64

fix panic with:
'reflect: call of reflect.Value.Int on uint32 Value'
and
'reflect: call of reflect.Value.Int on uint64 Value'

### DIFF
--- a/dml_events.go
+++ b/dml_events.go
@@ -15,17 +15,23 @@ import (
 
 type RowData []interface{}
 
-// The mysql driver never actually gives you a uint64 from Scan, instead you
-// get an int64 for values that fit in int64 or a byte slice decimal string
-// with the uint64 value in it.
+// The mysql driver never actually gives you a uint64 from Scan,
+// instead you get an int64 for values that fit in int64 or a byte slice
+// decimal string with the uint64 value in it.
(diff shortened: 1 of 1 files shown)


dml_events.go
```diff
@@ -15,17 +15,22 @@ import (
 
 type RowData []interface{}
 
-// The mysql driver never actually gives you a uint64 from Scan, instead you
-// get an int64 for values that fit in int64 or a byte slice decimal string
-// with the uint64 value in it.
+// The old version mysql driver never actually gives you a uint64 from Scan,
```
shuhaowu: Why the old version? Did something change in the new version? When?

everpcpc: A:
* int unsigned and bigint unsigned both
* rechecking and trying to reproduce this

everpcpc: updated.
It's go-mysql which gives right uint64.

shuhaowu:
Thanks for looking after this. However, I'm still not clear what table structure would lead to this? We use Ghostferry on tables with unsigned ints and this has never been an issue. We always get a string representation of the integer, which we parse.

PR #113, opened 2019-05-17
Fix assertParamSet not working with value receiver

everpcpc:


shuhaowu:
Should the `Source DatabaseConfig` and similar lines within the `Config` struct be pointers as well with this change?

everpcpc:
Update.
But I think it's better to keep them consistent in case of future mistake.

shuhaowu:
Can you update these lines as well https://github.com/Shopify/ghostferry/blob/master/config.go#L196-L201

everpcpc:
There seems no need to update these lines? They would not affect this problem I think.

config.go
```diff
@@ -58,7 +58,7 @@ type DatabaseConfig struct {
 	TLS *TLSConfig
 }
 
-func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
+func (c DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
```
shuhaowu: Why was this reverted? Since some methods are defined as pointer receivers, all methods on this struct should be pointer receivers. From the [golang FAQ](https://golang.org/doc/faq#methods_on_values_or_pointers)

> Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used. See the section on method sets for details.

shuhaowu:
Thanks!

PR #114, opened 2019-05-17
Add config SkipBinlogRowImageCheck

everpcpc:


ferry.go
```diff
@@ -682,6 +682,10 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
 
 	row = db.QueryRow("SHOW VARIABLES LIKE 'binlog_row_image'")
 	err = row.Scan(&name, &value)
+	if err == sql.ErrNoRows {
+		f.logger.Warning("binlog_row_image not found, indicating old version and default to FULL!")
```
shuhaowu: Which version of MySQL does this?

shuhaowu: Which version of MySQL does this? I don't think we officially support anything other than 5.7 (and I think it works on 5.6).

everpcpc: `binlog_row_image` was introduced in `mysql-5.6.2` and `MariaDB-10.1.6`.
I think maybe officially support dose not conflict with making this compatible with old version/other flavor as long as the change dose not affect the current version?

shuhaowu: I don't know what the implication of this change is with respect to the current version of MySQL. If it is possible to get a ErrNoRows and have binlog_row_image default to not FULL, silent data corruption could occur. 

Given that this behaviour could occur in MySQL 5.6, we can add a Config variable that allows this error to be ignored by Ghostferry. This way the user has to consciously make the choice to ignore this error.

shuhaowu:
Next time add a comment if you want another review, otherwise I won't get notified.

shuhaowu:
Thanks for contributing!

everpcpc:
> Next time add a comment if you want another review, otherwise I won't get notified.

got, thx~

PR #117, opened 2019-06-10
Use a custom TableSchema struct

shuhaowu:
Instead of relying on a third party struct for such a crucial piece of information, we use an extended version of this struct so we can store custom information on it. The main purpose is to store which columns are compressed to help with implementing the InlineVerifier.

DMLEvents and RowBatch also no longer hold their own copy of the table schema structure. They originally held their own copy as we thought about dealing with schema changes dynamically. Since our present plan is to pause when a schema change occurs and resume later, we don't need this anymore. This simplifies the code as all schema structures comes from the TableSchemaCache.

Lastly, RowBatch.AsSQLQuery and DMLEvent.AsSQLString have been changed such that they take the schema and table name as string arguments instead of the *schema.Table struct.

### DIFF
--- a/dml_events.go
+++ b/dml_events.go
@@ -38,8 +38,8 @@ func (r RowData) GetUint64(colIdx int) (res uint64, err error) {
 type DMLEvent interface {
 	Database() string
 	Table() string
-	TableSchema() *schema.Table
-	AsSQLString(target *schema.Table) (string, error)
+	TableSchema() *TableSchema
+	AsSQLString(string, string) (string, error)
 	OldValues() RowData
 	NewValues() RowData
 	PK() (uint64, error)
@@ -47,10 +47,8 @@ type DMLEvent interface {
 }
 
 // The base of DMLEvent to provide the necessary methods.
-// This desires a copy of the struct in case we want to deal with schema
-// changes in the future.
 type DMLEventBase struct {
-	table schema.Table
+	table *TableSchema
 	pos   mysql.Position
 }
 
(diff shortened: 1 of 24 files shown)


dml_events.go
```diff
@@ -38,19 +38,17 @@ func (r RowData) GetUint64(colIdx int) (res uint64, err error) {
...
 	PK() (uint64, error)
 	BinlogPosition() mysql.Position
 }
 
 // The base of DMLEvent to provide the necessary methods.
-// This desires a copy of the struct in case we want to deal with schema
-// changes in the future.
 type DMLEventBase struct {
-	table schema.Table
+	table *TableSchema
```
shuhaowu: No more copies!

test/go/iterative_verifier_test.go
```diff
@@ -393,12 +393,12 @@ func (t *ReverifyStoreTestSuite) SetupTest() {
 func (t *ReverifyStoreTestSuite) TestAddEntryIntoReverifyStoreWillDeduplicate() {
 	pk1 := uint64(100)
 	pk2 := uint64(101)
-	// different references to schema.Table shouldn't cause an issue.
```
shuhaowu: I don't really understand this comment. I used the same copy of TableSchema and all seem to work fine...

hkdsun:
> They originally held their own copy as we thought about dealing with schema changes dynamically

http://wiki.c2.com/?YouArentGonnaNeedIt

dml_events.go
```diff
@@ -62,8 +60,8 @@ func (e *DMLEventBase) Table() string {
 	return e.table.Name
 }
 
-func (e *DMLEventBase) TableSchema() *schema.Table {
-	return &e.table
```
fjordan: this (`&e.table`) was because we were dereferencing the struct pointer to the `schema.Table` pointer, but is no longer needed because `e.table` is now a struct and not a pointer?

PR #118, opened 2019-06-19
Report Ghostferry internal progress to a HTTP endpoint

shuhaowu:
This PR allows progress to be reported to an HTTP endpoint periodically so an Ghostferry management software can display more detailed information about a move.

The data looks like this:

```
{
    "CurrentState"=>"done", 
    "CustomPayload"=>"", 
    "Tables"=>{
        "gftest.test_table_1"=>{"LastSuccessfulPK"=>1111, "TargetPK"=>1111, "CurrentAction"=>"completed"}
    }, 
    "LastSuccessfulBinlogPos"=>{"Name"=>"mysql-bin.000005", "Pos"=>31730235}, 
    "BinlogStreamerLag"=>0.100753822, 
    "Throttled"=>false,
    "VerifierType"=>"", 
    "FinalBinlogPos"=>{"Name"=>"mysql-bin.000005", "Pos"=>31730235},
    "PKsPerSecond"=>9400,
    "ETA"=>0, 
    "TimeTaken"=>0.996557813
}
```

This PR does not have detailed progress reporting for the IterativeVerifier (no detailed table information). The `CurrentState` field will be changed to `verify-before-cutover` when the IterativeVerifier is running so we're not completely in the dark. Since we are still on track to remove the IterativeVerifier, I did not bother to hook up the progress from the verifier as it will need a lot of work. 

Note on implementation: there was a `Status` struct in the code previously. However, the way it is structured was too coupled with the `ControlServer`. I don't want to untangle that right now so I created a brand new `Progress` struct. In the future, we should remove the `Status` struct and have the control server rely on the `Progress` struct directly. The `Status` struct is renamed as `StatusDeprecated` to reflect this future.

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -6,6 +6,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
+	"net/http"
 	"os"
 	"os/signal"
 	"strings"
@@ -482,6 +484,24 @@ func (f *Ferry) Run() {
 		handleError("throttler", f.Throttler.Run(ctx))
 	}()
 
+	if f.Config.ProgressCallback.URI != "" {
+		supportingServicesWg.Add(1)
+		go func() {
+			defer supportingServicesWg.Done()
+
+			frequency := time.Duration(f.Config.ProgressReportFrequency) * time.Millisecond
+
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(frequency):
+					f.ReportProgress()
+				}
+			}
+		}()
+	}
+
 	if f.DumpStateOnSignal {
 		go func() {
 			c := make(chan os.Signal, 1)
@@ -496,7 +516,6 @@ func (f *Ferry) Run() {
 				os.Exit(0)
 			}
 		}()
-
 	}
 
 	binlogWg := &sync.WaitGroup{}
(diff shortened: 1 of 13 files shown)


test/helpers/db_helper.rb
```diff
@@ -108,7 +108,7 @@ def seed_simple_database_with_single_table
     # Create some holes in the data.
     delete_statement = source_db.prepare("DELETE FROM #{full_table_name(DEFAULT_DB, DEFAULT_TABLE)} WHERE id = ?")
     140.times do
-      delete_statement.execute(Random.rand(max_id) + 1)
+      delete_statement.execute(Random.rand(max_id))
```
shuhaowu: Since we report progress at the end of the run, I want to check the LastSuccessfulPK and TargetPK in tests. This requires the last id to be predictable.

ferry.go
```diff
@@ -627,6 +663,86 @@ func (f *Ferry) SerializeStateToJSON() (string, error) {
 	return string(stateBytes), err
 }
 
+func (f *Ferry) Progress() *Progress {
```
fjordan: Is it worth breaking up any of the below functionality into separate methods? Seems like there is a lot going on here.

progress.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type TableProgress struct {
+	LastSuccessfulPK uint64
+	TargetPK         uint64
+	CurrentAction    string // Possible values are defined via the constants TableAction*
+}
+
+type Progress struct {
+	// Possible values are defined via the constants State*
```
milanatshopify: Maybe mention these are in ferry.go (as much as that may age), because that file isn't modified, so it won't show up in the diff.

ferry.go
shuhaowu: I personally rather get the ugliness over with all in one method rather than putting it in multiple places.

shuhaowu:
> This looks nice. Do we have any idea if there is a measurable performance impact from doing all the collecting of status?

The performance impact should be very minimal as it is done in a separate goroutine in the background.

Issue #99, opened 2019-02-25
Investigate the possibility of removing the IterativeVerifier


shuhaowu:
After some discussion with @pushrax and @hkdsun, we think it may be possible that all the IterativeVerifier does is check if a type, encoding, or truncation issue (henceforth referred to as encoding issues) caused data corruption to occur during the BatchWriter and BinlogWriter. Examples of this are: #23 and https://github.com/siddontang/go-mysql/issues/205. It is unclear what other issues the IterativeVerifier can catch. This is especially true because the IterativeVerifier uses the same BinlogStreamer and data iteration as the Ferry: if there are any bugs within those structs, they would equally affect the IterativeVerifier.

If the assumptions above are indeed true, that is, the IterativeVerifier exists purely to catch encoding issues due to the need to translate data from MySQL to Golang and back to MySQL: the verification of data correctness could instead be done directly after the writes occurred. We can then eliminate the the IterativeVerifier and the overall architecture of Ghostferry would be much simpler. A simpler architecture also means the likelihood of bugs would be lower, so it's an appealing idea. It would also make implementing interrupt/resume much easier.

We should dig a little bit further on whether or not the IterativeVerifier only ever catch encoding issues. If so, we should decide if we want to get rid of the IterativeVerifier in favour of inline verification.

### Things to investigate

- Better reasoning/analysis on what the IterativeVerifier actually covers.
- Find examples of IterativeVerifiers failing a verification and investigate the causes.
- Understand what "checking directly after writes" means: what does it catch and what does it miss?
- Decide if we are OK with removing the IterativeVerifier in favour of checking inline.

### Some more optional thoughts

The original design of Ghostferry only included a CHECKSUM TABLE verifier. This verifier is implemented effectively outside of Ghostferry. It ensures the entire Ghostferry algorithm is implemented correctly (assuming CHECKSUM TABLE is implemented correctly). The IterativeVerifier was conceived as a drop-in replacement to CHECKSUM TABLE so we can verify partial table moves. This means the original mental model for this "drop-in" IterativeVerifier is that it would "ensures the entire Ghostferry algorithm is implemented correctly". It's unclear there was any direct challenge against this mental model.

So instead of getting rid of the IterativeVerifier, we could make the current IterativeVerifier match the mental model above more closely. However, it's equally unclear how to do this without reimplementing the binlog streaming and data iteration.

shuhaowu:
[...]

PR #120, opened 2019-06-24
InlineVerifier for DataIterator/BatchWriter

shuhaowu:
This PR adds verification inline while selecting data in DataIterator (Cursor) and writing data in BatchWriter. It does not add the verification to the BinlogStreamer so the PR is a reasonable size. That will come next. 

### Motivation

Some more notes on the InlineVerifier topic is available in #99. The tl;dr of it is:

1. Making the IterativeVerifier interruptible will significantly complicate the code base.
2. The InlineVerifier will allow us to get interrupt/resume for almost free.
3. The only type of corruption that both the IterativeVerifier and the InlineVerifier can guarantee detection is corruption due to an encoding-type bug.
4. The IterativeVerifier have can detect some additional types of corruption  (such as implementation bugs in interrupt/resume like an incorrect restart location). However, in none of these types can it guarantee detect and requires some luck. 
5. Simpler code base due to implementing only the InlineVerifier gives us a better chance at reducing bugs as it is easier to understand, test, and formally model.

### How it works

How the InlineVerifier works in the context of the DataIterator as implemented in this PR is as follows:

1. While selecting the data in `DataIterator`, the `SELECT * FROM ...` gets replaced with `SELECT *, MD5(...) FROM ...`.
2. The fingerprint is stored in the `RowBatch` struct.
3. When the `BatchWriter` tries to write the `RowBatch`, instead of inserting it directly with a single statement, it:
    1. Opens a transaction.
    2. Inserts the data as done previously
    3. `SELECT pk, MD5(....) FROM ...`
    4. Checks the MD5 against the fingerprint stored on `RowBatch`.

As one can note, the number of round trips per batch written has been increased from 1 to 4. I suspect this _should_ be okay, as inserting 200 rows should take the majority of the time. Furthermore, since we are eventually going to remove the iterative verifier in favour of using this, the overall clock time of a Ghostferry run should decrease.

Another complication is the presence of compressed columns that has already been copied to the target. If a compressed column is specified, step 3.3 will be modified to `SELECT pk, MD5, compressedCol1, compressedCol2....`. The MD5 will exclude the compressed columns. Additionally, the decompressed columns are compared in Golang after step 3.4.

### DIFF
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -8,8 +8,9 @@ import (
 )
 
 type BatchWriter struct {
-	DB           *sql.DB
-	StateTracker *CopyStateTracker
+	DB             *sql.DB
+	InlineVerifier *InlineVerifier
+	StateTracker   *CopyStateTracker
 
 	DatabaseRewrites map[string]string
 	TableRewrites    map[string]string
@@ -62,9 +63,34 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 			return fmt.Errorf("during prepare query near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
 		}
 
-		_, err = stmt.Exec(args...)
+		tx, err := w.DB.Begin()
 		if err != nil {
+			return fmt.Errorf("unable to begin transaction in BatchWriter: %v", err)
+		}
+
+		_, err = tx.Stmt(stmt).Exec(args...)
+		if err != nil {
+			tx.Rollback()
 			return fmt.Errorf("during exec query near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
+		}
+
+		if w.InlineVerifier != nil {
+			mismatches, err := w.InlineVerifier.CheckFingerprintInline(tx, db, table, batch)
+			if err != nil {
+				tx.Rollback()
+				return fmt.Errorf("during fingerprint checking for pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
+			}
+
+			if len(mismatches) > 0 {
+				tx.Rollback()
+				return fmt.Errorf("row fingerprints for pks %v do not match", mismatches)
+			}
+		}
+
+		err = tx.Commit()
+		if err != nil {
+			tx.Rollback()
+			return fmt.Errorf("during commit near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
 		}
 
 		// Note that the state tracker expects us the track based on the original
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -3,7 +3,6 @@ package ghostferry
 import (
 	"database/sql"
 	"fmt"
-	"sync"
 
 	"github.com/sirupsen/logrus"
 )
@@ -17,13 +16,12 @@ type BatchWriter struct {
 
 	WriteRetries int
 
-	mut        sync.RWMutex
-	statements map[string]*sql.Stmt
-	logger     *logrus.Entry
+	stmtCache *StmtCache
+	logger    *logrus.Entry
 }
 
 func (w *BatchWriter) Initialize() {
-	w.statements = make(map[string]*sql.Stmt)
+	w.stmtCache = NewStmtCache()
 	w.logger = logrus.WithField("tag", "batch_writer")
 }
 
@@ -59,9 +57,9 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 			return fmt.Errorf("during generating sql query at pk %v -> %v: %v", startPkpos, endPkpos, err)
 		}
 
-		stmt, err := w.stmtFor(query)
+		stmt, err := w.stmtCache.StmtFor(w.DB, query)
 		if err != nil {
-			return fmt.Errorf("during preparing query near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
+			return fmt.Errorf("during prepare query near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
 		}
 
 		_, err = stmt.Exec(args...)
@@ -77,35 +75,4 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 
 		return nil
 	})
-}
-
-func (w *BatchWriter) stmtFor(query string) (*sql.Stmt, error) {
-	stmt, exists := w.getStmt(query)
-	if !exists {
-		return w.newStmtFor(query)
-	}
-	return stmt, nil
-}
-
-func (w *BatchWriter) newStmtFor(query string) (*sql.Stmt, error) {
-	stmt, err := w.DB.Prepare(query)
-	if err != nil {
-		return nil, err
-	}
-
-	w.storeStmt(query, stmt)
-	return stmt, nil
-}
-
-func (w *BatchWriter) storeStmt(query string, stmt *sql.Stmt) {
-	w.mut.Lock()
-	defer w.mut.Unlock()
-	w.statements[query] = stmt
-}
-
-func (w *BatchWriter) getStmt(query string) (*sql.Stmt, bool) {
-	w.mut.RLock()
-	defer w.mut.RUnlock()
-	stmt, exists := w.statements[query]
-	return stmt, exists
 }
--- a/config.go
+++ b/config.go
@@ -171,6 +171,10 @@ type IterativeVerifierConfig struct {
 	//	1. Snappy (https://google.github.io/snappy/) as "SNAPPY"
 	//
 	// Optional: defaults to empty map/no compression
+	//
+	// Note that the IterativeVerifier is in the process of being deprecated.
+	// If this is specified, ColumnCompressionConfig should also be filled out in
+	// the main Config.
 	TableColumnCompression TableColumnCompressionConfig
 }
 
@@ -187,6 +191,25 @@ func (c *IterativeVerifierConfig) Validate() error {
 	}
 
 	return nil
+}
+
+// SchemaName => TableName => ColumnName => CompressionAlgorithm
+// Example: blog1 => articles => body => snappy
+//          (SELECT body FROM blog1.articles => returns compressed blob)
+type ColumnCompressionConfig map[string]map[string]map[string]string
+
+func (c ColumnCompressionConfig) CompressedColumnsFor(schemaName, tableName string) map[string]string {
+	t1, found := c[schemaName]
+	if !found {
+		return nil
+	}
+
+	t2, found := t1[tableName]
+	if !found {
+		return nil
+	}
+
+	return t2
 }
 
 type Config struct {
@@ -309,11 +332,35 @@ type Config struct {
 
 	// Only useful if VerifierType == Iterative.
 	// This specifies the configurations to the IterativeVerifier.
+	//
+	// This option is being in the processing of being deprecated.
 	IterativeVerifierConfig IterativeVerifierConfig
 
 	// For old versions mysql<5.6.2, MariaDB<10.1.6 which has no related var
 	// Make sure you have binlog_row_image=FULL when turning on this
 	SkipBinlogRowImageCheck bool
+
+	// This config is necessary for a special case of Ghostferry:
+	//
+	// - If you are copying a table where the data is already partially on the
+	//   target through some other means.
+	//   - In this case, for an applicable row: the PK on both the source and the
+	//     target are the same. INSERT IGNORE will simply skip copying that row.
+	//     This leaves the target data without modifications.
+	//   - While this is not a supported use case, careful examination of your
+	//     data may reveal that Ghostferry will not cause a data corruption.
+	//     This is left as an exercise for the reader.
+	// - In this case, the InlineVerifier will verify the data to be correct.
+	// - However, if this data is compressed with a non-determinstic algorithm
+	//   such as snappy, the compressed blob may not be equal even when the
+	//   uncompressed data is equal.
+	// - This column signals to the InlineVerifier that it needs to decompress
+	//   the data to compare identity.
+	//
+	// Note: a similar option exists in IterativeVerifier. However, the
+	// IterativeVerifier is being deprecated and this will be the correct place
+	// to specify it if you don't need the IterativeVerifier.
+	ColumnCompressionConfig ColumnCompressionConfig
 }
 
 func (c *Config) ValidateConfig() error {
--- a/config.go
+++ b/config.go
@@ -200,17 +200,17 @@ func (c *IterativeVerifierConfig) Validate() error {
 type ColumnCompressionConfig map[string]map[string]map[string]string
 
 func (c ColumnCompressionConfig) CompressedColumnsFor(schemaName, tableName string) map[string]string {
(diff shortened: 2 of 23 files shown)


inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/golang/snappy"
+	"github.com/sirupsen/logrus"
+)
+
+type InlineVerifier struct {
+	SourceDB *sql.DB
+	TargetDB *sql.DB
+
+	sourceStmtCache *StmtCache
+	targetStmtCache *StmtCache
```
shuhaowu: We only use the targetStmtCache for now for storing statements executing on the target. When we hook up the InlineVerifier with the BinlogStreamer, we'll need a StmtCache for the source db.

I included in here to show that I didn't refactor the StmtCache for only a single use case.

inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+	logger          *logrus.Entry
+}
+
+func (v *InlineVerifier) VerifyBeforeCutover() error {
+	// TODO: Iterate until the reverify queue is small enough
```
shuhaowu: This method will be implemented when the binlog portion of the InlineVerifier is hooked up

inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+	return nil
+}
+
+func (v *InlineVerifier) VerifyDuringCutover() (VerificationResult, error) {
+	// TODO: verify everything within the reverify queue.
+	return VerificationResult{}, nil
+}
+
+func (v *InlineVerifier) StartInBackground() error {
+	// not needed?
```
shuhaowu: These methods (`StartInBackground, Wait, Result`) are to satisfy the `Verifier` interface. They might be needed to play nice with the ControlServer interface for later.

test/lib/go/integrationferry.go
```diff
@@ -236,11 +236,27 @@ func main() {
...
 		config.VerifierType = ghostferry.VerifierTypeIterative
 		config.IterativeVerifierConfig = ghostferry.IterativeVerifierConfig{
 			Concurrency: 2,
 		}
+	} else if verifierType != "" {
+		config.VerifierType = verifierType
+	}
+
+	// This is currently a hack to customize the Ghostferry configuration.
+	// TODO: allow Ghostferry config to be specified by the ruby test directly.
```
shuhaowu: A bit hacky for now. I'll look into refactoring this another time.

cursor.go
```diff
@@ -80,6 +81,10 @@ func (c *Cursor) Each(f func(*RowBatch) error) error {
 		c.ColumnsToSelect = []string{"*"}
 	}
 
+	if c.SelectFingerprint {
+		c.ColumnsToSelect = append(c.ColumnsToSelect, c.Table.RowMd5Query())
+	}
```
hkdsun: There's two places in `Cursor` that check this flag.. This looks pretty leaky to me.

How about making the caller pass in the `ColumnsToSelect` option:

```go
func (c *Cursor) Each(columnsToSelect []string, f func(*RowBatch) error) error {
    ...
}
```

And since the caller has passed in the selection clause, they know what format of data to expect:
```go
// in data_iterator.go

err := cursor.Each(columnsIncludingMD5, func(batch *RowBatch) error {
  if c.SelectFingerprint {
    batchDataWithoutFingerprints []RowData
    fingerprints := make(map[uint64][]byte)

    for _, rowData := range batch.Values {
      // wrangle the resultset into `batchDataWithoutFingerprints` and `fingerprints`
    }

    batch = &RowBatch{
      values: batchDataWithoutFingerprints,
      fingerprints: fingerprints,
      pkIndex: batch.pkIndex,
      table: batch.Table,
    }
  }
  ...
}
```

config.go
```diff
@@ -171,6 +172,10 @@ type IterativeVerifierConfig struct {
 	//	1. Snappy (https://google.github.io/snappy/) as "SNAPPY"
 	//
 	// Optional: defaults to empty map/no compression
+	//
+	// Note that the IterativeVerifier is in the process of being deprecated.
+	// If this is specified, ColumnCompressionConfig should also be filled out in
+	// the main Config.
```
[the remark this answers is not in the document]
shuhaowu: I would like to get rid of the compression_verifier and iterative_verifier all together, but it would have to be another day. At that point I'll fix this comment.

ferry.go
```diff
@@ -73,7 +73,8 @@ type Ferry struct {
 	// If VerifierType is specified and this is nil on Ferry initialization, a
 	// Verifier will be created by Initialize. If an IterativeVerifier is to be
 	// created, IterativeVerifierConfig will be used to create the verifier.
-	Verifier Verifier
+	Verifier       Verifier
+	inlineVerifier *InlineVerifier
```
hkdsun: I don't think you need this here?

test/integration/inline_verifier_test.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def teardown
+    drop_triggers
+  end
+
+  def test_corrupted_insert_is_detected_inline_with_batch_writer
+    seed_random_data(source_db, number_of_rows: 3)
+    seed_random_data(target_db, number_of_rows: 0)
+
+    result = source_db.query("SELECT id FROM #{DEFAULT_FULL_TABLE_NAME} LIMIT 1")
```
hkdsun: `ORDER BY RAND()` ?

test/integration/inline_verifier_test.rb
```diff
@@ -0,0 +1,95 @@
+require "test_helper"
+
+class InlineVerifierTest < GhostferryTestCase
```
hkdsun: There's no test that has to deal with multiple batches of copied rows, right? I think we should have that

inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetDb, targetTable string, sourceBatch *RowBatch) ([]uint64, error) {
+	table := sourceBatch.TableSchema()
+
+	fingerprintQuery := table.FingerprintQuery(targetDb, targetTable, sourceBatch.Size())
+	stmt, err := v.targetStmtCache.StmtFor(v.TargetDB, fingerprintQuery)
+	if err != nil {
+		return nil, err
+	}
+
+	args := make([]interface{}, len(sourceBatch.Values()))
```
hkdsun: `s/args/pks`?

inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+func (v *InlineVerifier) Result() (VerificationResultAndStatus, error) {
+	// not implemented for now
+	return VerificationResultAndStatus{}, nil
+}
+
+func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetDb, targetTable string, sourceBatch *RowBatch) ([]uint64, error) {
+	table := sourceBatch.TableSchema()
+
+	fingerprintQuery := table.FingerprintQuery(targetDb, targetTable, sourceBatch.Size())
+	stmt, err := v.targetStmtCache.StmtFor(v.TargetDB, fingerprintQuery)
```
hkdsun: `s/stmt/fingerprintStmt`?

test/integration/inline_verifier_test.rb
shuhaowu: Do you mean multiple corrupted rows spread across multiple batches? The InlineVerifier will cause an error on the first corrupted row it detect tho.

batch_writer.go
```diff
@@ -59,16 +58,41 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
...
+			return fmt.Errorf("during prepare query near pk %v -> %v (%s): %v", startPkpos, endPkpos, query, err)
 		}
 
-		_, err = stmt.Exec(args...)
+		tx, err := w.DB.Begin()
 		if err != nil {
+			return fmt.Errorf("unable to begin transaction in BatchWriter: %v", err)
+		}
+
+		_, err = tx.Stmt(stmt).Exec(args...)
```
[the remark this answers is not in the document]
shuhaowu: The present code is correct. The documentation you quoted above is somewhat misleading. I verified that the [parent statement is not being closed](https://github.com/golang/go/issues/25329#issuecomment-388054264) (also is [commented in the database/sql code](https://github.com/golang/go/blob/release-branch.go1.10/src/database/sql/sql.go#L2514-L2518)). There's also not [another call to the prepare the statement unless it is a new connection](https://github.com/golang/go/blob/release-branch.go1.10/src/database/sql/sql.go#L2046-L2067).

table_schema_cache.go
```diff
@@ -34,6 +35,57 @@ type TableSchema struct {
...
+		}
+	}
+
+	hashStrs := make([]string, len(columns))
+	for i, column := range columns {
+		hashStrs[i] = fmt.Sprintf("MD5(COALESCE(%s, 'NULL'))", normalizeAndQuoteColumn(column))
+	}
+
+	t.rowMd5Query = fmt.Sprintf("MD5(CONCAT(%s)) AS __ghostferry_row_md5", strings.Join(hashStrs, ","))
+	return t.rowMd5Query
```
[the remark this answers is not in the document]
shuhaowu: Done.

shuhaowu:
Updated according to comments and rebased to master.

cursor.go
shuhaowu: I tried this, but I still needed two ifs within DataIterator.Run

```
cursor := CursorConfig.NewCursor(....)
if dataiterator.SelectFingerprint { 
  cursor.ColumnsToSelect = ... // This is also harder to write here because Cursor.Each set a default value for ColumnsToSelect if not specified
}
cursor.Each(func(batch *RowBatch) {
  if dataiterator.SelectFingerprint {
    // Extract fingerprint from batch
  }
  batch = &RowBatch{....}
  ...
})
```

table_schema_cache.go
```diff
@@ -34,6 +35,57 @@ type TableSchema struct {
...
+	columnsToSelect[0] = quoteField(t.GetPKColumn(0).Name)
+	columnsToSelect[1] = t.RowMd5Query()
+	i := 2
+	for columnName, _ := range t.CompressedColumns {
+		columnsToSelect[i] = quoteField(columnName)
+		i += 1
+	}
+
+	return fmt.Sprintf(
+		"SELECT %s FROM %s WHERE %s IN (%s)",
```
fjordan: Do we need to add an explicit `ORDER BY` here?  I'm concerned if we leave it to implicit ordering it may not return deterministically for the source and target.

inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+
+		pk, err := strconv.ParseUint(string(rowData[0]), 10, 64)
+		if err != nil {
+			return nil, err
+		}
+
+		targetFingerprints[pk] = rowData[1]
+		targetDecompressedData[pk] = make(map[string][]byte)
+
+		for i := 2; i < len(columns); i++ {
```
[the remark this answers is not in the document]
shuhaowu: I can add a comment here, this way we don't need an extra variable? Would that be acceptable?

inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+			sourceDecompressedData[pk][col.Name], err = v.DecompressData(table, col.Name, compressedData)
+		}
+	}
+
+	mismatches := v.CompareHashes(sourceFingerprints, targetFingerprints)
+	if len(mismatches) > 0 {
+		return mismatches, nil
+	}
+
+	mismatches = v.CompareDecompressedData(sourceDecompressedData, targetDecompressedData)
```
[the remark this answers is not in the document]
shuhaowu: Actually in a follow up PR I'm guaranteed to call this method every time as I need the exact PKs that mismatched for reverification purposes.

I implemented this because I assume it can be understood that if the sourceDecompressedData and targetDecompressedData are empty, there would be no mismatches. Is this not how you understood this method?

table_schema_cache.go
shuhaowu: All consumers of this method will read the rows and build a map of PK => fingerprints. We'll have to do this whether or not this query is done via an ORDER BY for ease of comparison (especially later with the background reverifier).

inline_verifier.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		sourceDecompressedData[pk] = make(map[string][]byte)
+		for idx, col := range table.Columns {
+			var compressedData []byte
+			var ok bool
+			if _, ok = table.CompressedColumns[col.Name]; !ok {
+				continue
```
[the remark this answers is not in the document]
shuhaowu: It looks like the [memory overhead of an empty map is around 128bytes](https://stackoverflow.com/questions/15313105/memory-overhead-of-maps-in-go/), so it shouldn't be a big deal. I can tweak this code slightly, tho.

table_schema_cache.go
fjordan: Didn't realize it was a map 👍

inline_verifier.go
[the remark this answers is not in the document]
fjordan: Works for me

inline_verifier.go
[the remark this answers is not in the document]
fjordan: I saw that once I looked at the method, but it wasn't obvious at first and I was wondering why it was always called if the above block `continue`d out of the execution path

test/integration/inline_verifier_test.rb
hkdsun: Hmm I see.. May still be worth enabling the trigger after a batch or two have been copied?

cursor.go
hkdsun: > two ifs within DataIterator.Run

That sounds more reasonable than having it be within a general-purpose cursor IMO

inline_verifier.go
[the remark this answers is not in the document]
shuhaowu: I'm going to leave this code as is now, as I refactor these two method into a standalone method in the next PR where I add the reverifier for the binlog streamer

It'll looke like this: `v.compareHashesAndData(sourceFingerprints, targetFingerprints, sourceDecompressedData, targetDecompressedData)`

shuhaowu:
I made the `SelectFingerprint` change in the latest commit. It's short. Care to take another look before I merge this?

inline_verifier.go
```diff
@@ -91,6 +91,9 @@ func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetDb, targetTabl
 		targetFingerprints[pk] = rowData[1]
 		targetDecompressedData[pk] = make(map[string][]byte)
 
+		// Note that the FingerprintQuery returns the columns: pk, fingerprint,
+		// compressedData1, compressedData2, ...
+		// If there are no compressed data, only 2 columns are returned.
```
fjordan: nit: `If there are no compressed data, only 2 columns are returned and this block will be skipped`

test/integration/inline_verifier_test.rb
shuhaowu: I don't think it will be that different, also getting the timing right is still a bit difficult.

inline_verifier.go
shuhaowu: I'll change this in the next PR as I moved this code around.

PR #123, opened 2019-07-15
Removed StateTracker complexity

shuhaowu:
The `StateTracker` code was complex as it was split between the `CopyStateTracker` and the `VerifierStateTracker`. Since we are moving towards the `InlineVerifier`, which requires very little state, we don't need this complexity anymore. 

This PR removes the extra complexity and only retains a single `StateTracker` and `SerializableState`.

### DIFF
--- a/state_tracker.go
+++ b/state_tracker.go
@@ -17,126 +17,43 @@ import (
 //
 // The state tracker keeps track of the progress of Ghostferry so it can be
 // interrupted and resumed. The state tracker is supposed to be initialized and
-// managed by the Ferry. Each Ghostferry components, such as the `BatchWriter`
-// and `IterativeVerifier` will get passed an instance of either the
-// `CopyStateTracker` or the `VerificationStateTracker`. During the run, these
+// managed by the Ferry. Each Ghostferry components, such as the `BatchWriter`,
+// will get passed an instance of the StateTracker. During the run, the
 // components will update their last successful components to the state tracker
-// instances given via the state tracker API defined here.
+// instance given via the state tracker API defined here.
 //
-// The states stored in the state tracker can be copied into a serialization
-// friendly struct (`SerializableState`), which can then be dumped using
-// something like JSON. Assuming the rest of Ghostferry used the API of the
-// state tracker correctly, this can be done at any point during the Ghostferry
-// run and the resulting state is can be resumed from without incurring data
-// loss. The same `SerializableState` can be used as an input to `Ferry`, which
-// if specified will instruct to `Ferry` to start from a particular location.
-//
-// Code Design
-// -----------
-//
-// In a Ghostferry run, there are two "stages" of operation: the copy stage and
-// the verification stage. Both stages must emit their states to the state
-// tracker in order for them to be interruptible and resumable. These two
-// stages are very similar: they both iterate over the data and tail the
-// binlog. However, there are some minor differences. Example: the verifier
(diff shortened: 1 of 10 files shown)


state_tracker.go
```diff
@@ -160,65 +77,68 @@ func newSpeedLogRing(speedLogCount int) *ring.Ring {
...
+	CopyRWMutex   *sync.RWMutex
 
-func NewBinlogStateTracker() *BinlogStateTracker {
-	return &BinlogStateTracker{
-		RWMutex:                     &sync.RWMutex{},
-		lastProcessedBinlogPosition: mysql.Position{},
-	}
-}
+	lastWrittenBinlogPosition                 mysql.Position
+	lastStoredBinlogPositionForInlineVerifier mysql.Position
```
hkdsun: nit: `lastVerifiedBinlogPosition` is simpler? Do we need to introduce this in this PR? Seems to me like a followup to https://github.com/Shopify/ghostferry/pull/122

state_tracker.go
```diff
@@ -17,126 +17,43 @@ import (
...
-// The binlog writer and the verify binlog positions are different because the
-// binlog writer is buffered in a background go routine. Its position can race
-// with respect to the verifier binlog position. The minimum position between
-// the two are always safe to resume from.
 func (s *SerializableState) MinBinlogPosition() mysql.Position {
-	if s.VerifierStage == nil {
-		return s.CopyStage.LastProcessedBinlogPosition
+	nilPosition := mysql.Position{}
+	if s.LastWrittenBinlogPosition == nilPosition {
+		return s.LastStoredBinlogPositionForInlineVerifier
```
hkdsun: If there was no `LastWrittenBinlogPosition` (nothing was written by the binlog writer), how could we have a `LastStoredBinlogPositionForInlineVerifier`?

state_tracker.go
```diff
@@ -17,126 +17,43 @@ import (
...
+	CompletedTables                           map[string]bool
+	LastWrittenBinlogPosition                 mysql.Position
+	LastStoredBinlogPositionForInlineVerifier mysql.Position
 }
 
-// The binlog writer and the verify binlog positions are different because the
-// binlog writer is buffered in a background go routine. Its position can race
-// with respect to the verifier binlog position. The minimum position between
-// the two are always safe to resume from.
 func (s *SerializableState) MinBinlogPosition() mysql.Position {
```
hkdsun: Why is this method defined on this struct? Seems like a consumer concern

state_tracker.go
shuhaowu: Both the `LastWrittenBinlogPosition` and the `LastStoredBinlogPositionForInlineVerifier` are initialized during `Ferry.Start`, right after the binlog streamer is connected. This is done to address a bug where if the application database got not writes to it while Ghostferry is running, Ghostferry could not resume.

state_tracker.go
shuhaowu: I was more thinking of this as an "attribute" of the SerializableState and hence put it here.

state_tracker.go
shuhaowu: It's not quite the last verified binlog position. As soon as the binlog event is added to the reverify store, we record this position. The reverify store will be dumped anyway so we won't lose anything. 

Either way, I'll remove it for now and send a follow up PR after #122.

shuhaowu:
Removed the Verifier state tracking from here in 3caa901. I'll re-add it in a PR following #122.

PR #124, opened 2019-07-23
Re-adds InlineVerifier state into StateTracker

shuhaowu:
Review this after #122 as it would make more sense.


inline_verifier.go
```diff
@@ -516,6 +543,7 @@ func (v *InlineVerifier) verifyAllEventsInStore() (bool, map[string]map[string][
 		if err != nil {
 			return false, nil, err
 		}
+		v.reverifyStore.RemoveVerifiedBatch(batch)
```
hkdsun: Hmm what if a PK contained within `batch` gets modified again between L525 and this line?

hkdsun: As mentioned on Slack, removing an entry _after_ it's been handled still makes sense, so I'd like to find a good solution here.

Perhaps we should store entries in the reverifyStore as events with unique IDs so that their removal is a lot safer. WDYT?

shuhaowu: You're right, there's a race condition here. Unique ID (perhaps via the mysql binlog position?) should work but I'd need to alter the data structure of the BinlogVerifyStore again. I'll think about it for a bit.

hkdsun: Yeah that feels safe enough. Generating a UUID on event creation is another optino

fjordan: Agree with the concerns and proposed solutions.

state_tracker.go
```diff
@@ -34,13 +34,27 @@ type SerializableState struct {
...
+	nilPosition := mysql.Position{}
+	if s.LastWrittenBinlogPosition == nilPosition {
+		return s.LastStoredBinlogPositionForInlineVerifier
+	}
+
+	if s.LastStoredBinlogPositionForInlineVerifier == nilPosition {
+		return s.LastWrittenBinlogPosition
+	}
+
+	if s.LastWrittenBinlogPosition.Compare(s.LastStoredBinlogPositionForInlineVerifier) >= 0 {
```
fjordan: What exactly is this comparing? We're returning the lesser of the two `LastWrittenBinlogPosition` and `LastStoredBinlogPositionForInlineVerifier`?

shuhaowu: These two positions can be out of sync because the BinlogWriter independently from adding the binlog events to the BinlogVerifyStore. However, to resume, we need to resume from a single binlog coordinate, which should be the minimum one.

inline_verifier.go
shuhaowu: Generating an unique ID and then try to remove the event from the BinlogVerifyStore was too cumbersome to implement. It'll taint a lot of the code within the InlineVerifier and the unique ID leaks all over the place. In essence, we do not have the right data structure to support this.

However, I've come up with a simpler solution that works with our code for now. Right now the BinlogVerifyStore is essentially `map[uint64]struct{}`, or a simple set. The simple solution is to change it to `map[uint64]int`, where the value is the number of times the row has been changed. This way, `RemoveVerifiedBatch` simply decrements the counter while `Add` increments the counter. If the counter reaches 0, `RemoveVerifiedBatch` will delete the entry from the map to keep the serialized map smaller.

This will result in some wasted work as we may reverify a row multiple times unnecessarily. However, any other changes likely will require the BinlogVerifyStore to be something other than a simple map and not worth the effort for now IMO. This workaround should allow us to move forward and ship full-resumability and we can more easily revisit this problem in a more isolated PR.

shuhaowu:
I've not really figured out a way to test the race condition. There's no hooks into this part of the code yet for the Ruby integration tests to pause.

I'm not quite sure even how to verify this fix. I could run a TLA+ analysis but it'll take a little bit of time as I'm a bit rusty. Any other ideas? Is this verification even necessary or is the fix "simple enough"?

PR #125, opened 2019-08-08
Use go modules instead of dep

kirinrastogi:
I had some free time and I was curious about this project.

Go is adding first class support for go modules, this *should* be the last time you switch your dependency management tool.

There shouldn't be much to review, as the `vendor/` folder will continue to be used.
I updated the go version to `1.12.7` as well.

Note that now instead of running `dep ensure -v` you run `go mod verify` or `go mod download`

I am getting a test failure locally, I am investigating if it is related to this go version bump:

```
    --- FAIL: TestTableSchemaCache/TestFingerprintQuery (0.06s)
        assertions.go:247: 
                          
	Error Trace:	table_schema_cache_test.go:151
            
	Error:      	Not equal: 
            
	            	expected: "SELECT `id`,MD5(CONCAT(MD5(COALESCE(`id`, 'NULL')))) AS __ghostferry_row_md5,`data` FROM `s`.`t` WHERE `id` IN (?,?,?,?,?,?,?,?,?,?)"
            
	            	actual  : "SELECT `id`,MD5(CONCAT(MD5(COALESCE(`id`, 'NULL')),MD5(COALESCE(`data`, 'NULL')))) AS __ghostferry_row_md5,`data` FROM `s`.`t` WHERE `id` IN (?,?,?,?,?,?,?,?,?,?)"

```


shuhaowu:
Hey I did this last year: b34dc02e10a5fcbd34450cd4cdb39a537986ca73#diff-bd247e83efc3c45ae9e8c47233249f18

Your test failure looks like one of the cases I documented: https://github.com/Shopify/ghostferry/issues/107

shuhaowu:
> this should be the last time you switch your dependency management tool.

:trollface:

kirinrastogi:
Neat, I did not see that issue :eyes:

shuhaowu:
Actually that case is a different issue. I recently introduced it into the tests and I believe it's due to the fact that go maps iterate over in random orders. I think I address it in one of my dev branches that I haven't PRed.

Don't worry about it and hit retry until it passes.

kirinrastogi:
They passed on the first run 🥇

shuhaowu:
One thing is updating to Golang 1.12.7 makes it hard on Ubuntu bionic. I'll have to figure out a work around to continue working on this. I'm not sure if I'll have time this week to do it.

kirinrastogi:
The last two commits ensure that `go` uses the vendor directory. I'm not sure how you manage your env on ubuntu but it would be useful to have
```bash
GO111MODULE=on
GOFLAGS=-mod=vendor
```

set

shuhaowu:
Thanks again for the help. I've been doing my yearly go dependency management reading today and I noted from an article:

> Our aim is for Go 1.13, scheduled for August 2019, to enable module mode by default (that is, to change the default from auto to on) and deprecate GOPATH mode. In order to do that, we’ve been working on better tooling support along with better support for the open-source module ecosystem. 

I understand the interface is relatively stable, but given that Go1.13 is supposed to be out soon, I think we should hold off for now, especially since we would otherwise still need to add those environment variables into the Makefile.

PR #128, opened 2019-09-03
Explicitly link tutorial and documentation

yariksheptykin:
When clicking on the "documentation" link one lands on page prompting to select a branch, which seems unintended. This MR links documentation directly to its master version. It also explicitly links "tutorial" to its page in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,5 +45,5 @@ Run tests:
 Test copydb:
 
 - `make copydb && ghostferry-copydb -verbose examples/copydb/conf.json`
-- For a more detailed tutorial, see the
-  [documentation](https://shopify.github.io/ghostferry).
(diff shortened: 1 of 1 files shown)


shuhaowu:
If we ever tag a release, won't this link to the wrong documentation version?

yariksheptykin:
Thanks for taking a look a my PR @shuhaowu ! Well, if this lands, master is going be hard-wired in the link. So regardless of tagging we will always point to the latest documentation. I seemed to me that the page prompting to choose the version was somewhat unintended. But if that's by design, feel free to close the PR.

shuhaowu:
It was a design decision to have a version switcher, despite us never having used it.

PR #129, opened 2019-09-06
sharding: Use InlineVerifier by default

shuhaowu:
Note that CompressionVerifierTest is removed from sharding/test. The CompressionVerifier is superceded by the InlineVerifier which has the features to verify compressed columns without a separate struct. The compression verifier tests rely on a hard coded IterativeVerifier on `sharding.Ferry`. It would require major changes to deprecated functionality if we want to keep these tests, which is deprecated itself.

Hopefully over the next while I can remove most of IterativeVerifier once we further validate the InlineVerifier.


shuhaowu:
Oops I ran the wrong test suites locally and CI just failed on a real bug. Looks like the InlineVerifier is not playing nicely with the delta copier. I'll fix it and send another comment once I'm done.

shuhaowu:
The issue was that the join table tests only inserts into joined tables, which are filtered out by the `ApplicableEvents` filter within the BinlogStreamer. This means the list of events the BinlogStreamer emits to the event listeners is empty. The InlineVerifier binlog event handler does not expect this and thus panics. The fix is made within the BinlogStreamer such that the event listeners are not invoked if all events are filtered.

I'm ready for another review now.

PR #132, opened 2019-09-10
Introduce database abstraction layer

kgalieva:
Refactoring database calls as a first step to introduce comments for Ghostferry's sql queries.

Next step will be to add comments to GhostferryDB's functions.


ghostferry_db.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+	"database/sql"
+)
+
+type GhostferryDB struct {
+	SqlDB *sql.DB
+}
+
+type GhostferryTx struct {
+	Tx *sql.Tx
```
shuhaowu: Can you embed the pointer directly as opposed having a member variable? This should reduce the amount of `tx.Tx`.

ghostferry_db.go
```diff
@@ -0,0 +1,44 @@
+package ghostferry
+
+import (
+	"context"
+	"database/sql"
+)
+
+type GhostferryDB struct {
+	SqlDB *sql.DB
```
shuhaowu: Ditto.

ghostferry_db.go
```diff
@@ -0,0 +1,44 @@
+package ghostferry
+
+import (
+	"context"
+	"database/sql"
+)
+
+type GhostferryDB struct {
```
shuhaowu: Is `ghostferry.GhostferryDB` a bit verbose? Maybe just DB is fine?

shuhaowu: Or DBWrapper to differentiate it from sql.DB. I don't have a strong preference here.

kgalieva:
It's after BFCM now, could I please get a review?

testhelpers/data_writer.go
```diff
@@ -101,7 +101,7 @@ type DataWriter interface {
 	Run()
 	Stop()
 	Wait()
-	SetDB(*sql.DB)
+	SetDB(db *ghostferry.DB)
```
fjordan: is the named parameter `db` needed here? Maybe just omit it like it was prior?

wait_until_replica_is_caught_up_to_master.go
```diff
@@ -11,7 +10,7 @@ import (
 )
 
 type ReplicatedMasterPositionFetcher interface {
-	Current(*sql.DB) (mysql.Position, error)
+	Current(db *DB) (mysql.Position, error)
```
fjordan: Same comment here. I don't believe we need the named parameter to this interface method

kgalieva:
@shuhaowu could you please review?

shuhaowu:
While the code is fine, I'd much rather see the wrapper be in a separate package (perhaps `github.com/Shopify/ghostferry/sqlwrapper`) and then imported via `import sql "github.com/Shopify/ghostferry/sqlwrapper"` as opposed to changing every `sql.DB` line in the codebase. I'm concerned about if we actually need this code so if we want to merge it now, we should make it so it's easily removable.

hkdsun:
@shuhaowu that's a clever pattern 😄

kgalieva:
@shuhaowu I can't use `sql` alias in sqlwrapper import, because it will conflict with "database/sql".

shuhaowu:
I just tried it, it seems to work: [apptest.tar.gz](https://github.com/Shopify/ghostferry/files/4149766/apptest.tar.gz).

If there are some other issues that I wasn't able to replicate, you can also do something like this in `sqlwrapper` ( but this shouldn't be needed ):

```
import sqlorig "database/sql"

type DB struct {
	*sqlorig.DB

	C string
}

```

kgalieva:
The problem is that I'm not replacing "database/sql" completely. Some files (f.e. `cursor.go`) will import both "database/sql"  and "github.com/Shopify/ghostferry/sqlwrapper". If I use `sql` alias for "github.com/Shopify/ghostferry/sqlwrapper", then I'll need to override an alias for "database/sql" to something like `sqlorig`. Is this what you are suggesting?

shuhaowu:
Yeah whatever yields the least amount of code changes is preferred. I looked through cursor.go and the only thing is `sql.Rows` in two instances? In this case you can alias the original sql library to minimize the changes necessary.

kgalieva:
@shuhaowu could I get another round of review please?

config.go
```diff
@@ -126,7 +127,8 @@ func (c *DatabaseConfig) SqlDB(logger *logrus.Entry) (*sql.DB, error) {
 		logger.WithField("dsn", MaskedDSN(dbCfg)).Info("connecting to database")
 	}
 
-	return sql.Open("mysql", dbCfg.FormatDSN())
+	sqlDB, err := sqlorig.Open("mysql", dbCfg.FormatDSN())
+	return &sql.DB{sqlDB}, err
```
shuhaowu: Is it not possible to add this to sqlwrapper?

sqlwrapper/ghostferry_db.go
```diff
@@ -0,0 +1,82 @@
...
+	return tx.Tx.Query(query, args...)
+}
+
+func (tx Tx) QueryRowContext(ctx context.Context, query string, args ...interface{}) *sqlorig.Row {
+	return tx.Tx.QueryRowContext(ctx, query, args...)
+}
+
+func (tx Tx) QueryRow(query string, args ...interface{}) *sqlorig.Row {
+	return tx.Tx.QueryRow(query, args...)
+}
```
shuhaowu: I thought newline at EOF is inserted automatically by go fmt?

test/go/binlog_streamer_test.go
```diff
@@ -29,10 +30,11 @@ func (this *BinlogStreamerTestSuite) SetupTest() {
 	this.Require().Nil(err)
 
 	sourceDSN := sourceConfig.FormatDSN()
-	this.sourceDb, err = sql.Open("mysql", sourceDSN)
+	sqlDB, err := sqlorig.Open("mysql", sourceDSN)
 	if err != nil {
 		this.Fail("failed to connect to source database")
 	}
+	this.sourceDb = &sql.DB{sqlDB}
```
shuhaowu: Yeah we should just have sqlwrapper.Open do this wrapping automatically.

test/go/iterative_verifier_collation_test.go
```diff
@@ -30,7 +31,8 @@ func (t *IterativeVerifierCollationTestSuite) SetupTest() {
 
 	unsafeDSN := unsafeConfig.FormatDSN()
 
-	t.unsafeDb, err = sql.Open("mysql", unsafeDSN)
+	sqlDB, err := sqlorig.Open("mysql", unsafeDSN)
+	t.unsafeDb = &sql.DB{sqlDB}
```
shuhaowu: ditto

kgalieva:
@shuhaowu I think the tests have race conditions and CI build is randomly timing out. I made a few empty commits to make the build pass. The PR is ready for review.

kgalieva:
Here is a PR with similar failures on the master branch https://github.com/Shopify/ghostferry/pull/154

shuhaowu:
The stalling failures are likely caused by this: https://github.com/Shopify/ghostferry/issues/107#issuecomment-505441991

PR #133, opened 2019-09-13
Additional info logs in DataIterator

shuhaowu:
Note: I need this for some downstream tests so I can more reliably interrupt Ghostferry based on logging outputs. That said, this is probably nice in case someone wants to run Ghostferry without `-verbose`.

If debug level logs are turned off (as in ghostferry-sharding), there are no visibility of the DataIterator's progress in the logs. This PR adds at most 50 lines of logs while the DataIterator runs.

The log lines are emitted after a table is queued. Since the queue is an unbuffered channel and consumed by the tableIterators, logging here will reliably report the approximate number of tables copied by Ghostferry.

### DIFF
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -7,6 +7,15 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+type BatchWriterVerificationFailed struct {
+	mismatchedPks []uint64
+	table         string
+}
+
+func (e BatchWriterVerificationFailed) Error() string {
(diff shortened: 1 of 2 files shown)


shuhaowu:
I added another logging change where verification failures are more clearly marked in the DataIterator. This way it's easier for downstream consumers to distinguish between a generic data iteration error vs a verification failure.

shuhaowu:
Adding a level between `debug` and `info` is not easy because logrus doesn't support this built in.

PR #134, opened 2019-09-18
Move to Go Modules

epk:
Continuing from #125 

With 1.13 finally out, I think it is safe to move forward with Go Modules in general with this project.

There are a few hurdles if we really want ghostferry to be vgo/Go modules compliant not just support it in a hacky manner:

### github.com/go-sql-driver/mysql

`github.com/go-sql-driver/mysql` seems to have versioned releases but they don't adhere to semantic versioning or have a `go.mod` in their repo.

Currently `ghostferry` uses `github.com/go-sql-driver/mysql@v1.3.0`

Running `go mod tidy` (which is a recommended step before cutting a new release see: https://github.com/golang/go/wiki/Modules#releasing-modules-all-versions) 
bumps `github.com/go-sql-driver/mysql` to `v1.4.1` (the latest minor version 🤔🤔🤔) which breaks bunch of tests with:

```go
  --- FAIL: TestIterativeVerifierCollationTestSuite/TestVerifyCompressedOncePass (0.00s)
        Error Trace:    unit_test_suite.go:66
                    	iterative_verifier_test.go:26
                    	iterative_verifier_collation_test.go:23
                    	suite.go:88
        Error:      	Expected nil, but got: &errors.errorString{s:"this user requires mysql native password authentication."}
```

### Versioning

From Go Module FAQ (https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher):

> if you are adopting modules for the first time for a pre-existing repository or set of packages that have already been tagged v2.0.0 or higher before adopting modules, then the recommended best practice is to increment the major version when first adopting modules

Due to how they have defined the semantics in vgo we'll need to adopt versioned releases of `ghostferry` which will require tagging on GitHub and merging this into a `v2` branch and set that as the default on GitHub in addition to what I have done in https://github.com/Shopify/ghostferry/pull/134/commits/13ddb396e322f21639b9aa41ad36992addface8e

The route I went with in https://github.com/Shopify/ghostferry/pull/134/commits/13ddb396e322f21639b9aa41ad36992addface8e is the less backwards compatible one as it will break any projects using `ghostferry` with Go versions 1.10.2 or older and using `dep`:
> Note: creating a new branch is not required. If instead you have been previously releasing on master and would prefer to tag v3.0.0 on master, that is a viable option. (However, be aware that introducing an incompatible API change in master can cause issues for non-modules users who issue a go get -u given the go tool is not aware of semver prior to Go 1.11 or when module mode is not enabled in Go 1.11+).
Pre-existing dependency management solutions such as dep currently can have problems consuming a v2+ module created in this way. See for example dep#1962.

The safer option is to have a major subdirectory `v2/` which is backwards compatible and move all of the code from this PR in the `v2/` subdirectory. Though that might require changing how we run tests/makefiles etc.

I think going with the safer `v2/` subdirectory is the better option here but I will let you make that call :D 

cc @shuhaowu 

Sorry for the massive delete diff, I got rid of `/vendor` because vendoring is a PITA with Go modules because of the need to call go with the flag `-mod=vendor` or set `GOFLAGS=-mod=vendor` across all {dev,build,ci,...} environments. It does not seem to add any benefit just increases complexity and repo size.


shuhaowu:
I took a look at go modules post 1.13 and realize the effort needed is vastly more than I have time for right now. Further, I don't want to risk having build issues if we roll out a new version of Ghostferry with full resumability within the next few weeks. Thus, I'm going to delay this until at least December.

PR #135, opened 2019-09-23
Ignore certain columns during InlineVerification

shuhaowu:
This code is used in the same special case that handles the CompressedColumns: if data on the source and target already agrees with each other except for a non-critical field (say the a last_updated timestamp that does not affect the actual data integrity), the InlineVerifier should be able to be skip that particular column during verification to avoid a failed run. This option previously existed within the IterativeVerifier but I neglected to bring it over to the InlineVerifier. 

I must admit, I don't quite like how the code is structured (w.r.t. adding yet another parameter to `TableSchema` and another argument to `LoadTables()`). However, I can't see any other simple ways to get this to work. If there's some sort of cleanup/refactor we can do here, I'm all ears.

### DIFF
--- a/config.go
+++ b/config.go
@@ -250,6 +250,24 @@ func (c ColumnCompressionConfig) CompressedColumnsFor(schemaName, tableName stri
 	return columnsConfig
 }
 
+// SchemaName => TableName => ColumnName => struct{}{}
+// These columns will be ignored during InlineVerification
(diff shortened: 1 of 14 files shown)


shuhaowu:
> If we add another, then it probably warrants it.

Agreed. Thanks for the review.

PR #136, opened 2019-09-25
Less noisy logs from the MD5 transform

shuhaowu:
Cursor emitted logs that includes the entire query. With the InlineVerifier, this query could be huge and thus clutters up the log. This change omits the columns selected, which reduces this clutter.

Example log line before this change:

```
{
    "args":[923],
    "level":"debug",
    "msg":"found 178 rows",
    "sql":"SELECT *, MD5(CONCAT(MD5(COALESCE(`id`, 'NULL')),MD5(COALESCE(`data`, 'NULL')))) AS __ghostferry_row_md5 FROM `gftest`.`test_table_1` WHERE `id` \u003e ? ORDER BY `id` LIMIT 200 FOR UPDATE",
    "table":"gftest.test_table_1",
    "tag":"cursor",
    "time":"2019-09-25T14:11:35-04:00"
}
```

Example after:
```
{
    "args":[911],
    "level":"debug",
    "msg":"found 187 rows",
    "sql":"SELECT [omitted] FROM  `gftest`.`test_table_1` WHERE `id` \u003e ? ORDER BY `id` LIMIT 200 FOR UPDATE",
    "table":"gftest.test_table_1",
    "tag":"cursor",
    "time":"2019-09-25T14:08:35-04:00"
}
```

### DIFF
--- a/cursor.go
+++ b/cursor.go
@@ -3,6 +3,7 @@ package ghostferry
 import (
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/Masterminds/squirrel"
 	"github.com/siddontang/go-mysql/schema"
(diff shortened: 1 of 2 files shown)


cursor.go
```diff
@@ -166,8 +167,15 @@ func (c *Cursor) Fetch(db SqlPreparer) (batch *RowBatch, pkpos uint64, err error
 		return
 	}
 
+	// With the inline verifier, the columns to be selected may be very large as
+	// the query generated will be very large. The code here simply hides the
+	// columns from the logger to not spam the logs.
+
+	splitQuery := strings.Split(query, "FROM")
+	loggedQuery := fmt.Sprintf("SELECT [omitted] FROM %s", splitQuery[1])
```
fjordan: Can we add a simple test for this? Also, a nitpick, but maybe `omitted-md5-string` instead of just `omitted`?

shuhaowu: It's not just md5-string, there could also be other columns in there. I don't think it's necessarily worth to rebuild the other columns and omit only the md5 string tho.

fjordan: Right, `omitted-md5-transform` then. Just something a bit more descriptive.

PR #137, opened 2019-09-26
Further tests for the InlineVerifier

shuhaowu:
Is a decently straightforward PR, although I do fix an issue with `NULL` vs `'NULL'`.

- Tested that InlineVerifier works across multiple charsets and collations.
- Tested positive and negative zero floating point values.
- Tested NULL values.

### Appendix: Technical analysis of the `NULL` vs `'NULL'` problem

One thing I uncovered is that `NULL` is being treated the same as `'NULL'` by the InlineVerifier. Thus, we're prone to [false](https://www.wired.com/2015/11/null/) [negatives](https://www.wired.com/story/null-license-plate-landed-one-hacker-ticket-hell/) involving NULL. The cause of the issue is as folllows:

- To generate a checksum, for each row, we compute the MD5 of each column and concatenate the result. Then we generate yet another MD5 for this concatenated string.
- MD5 checksum is computed on the mysql server, using the `MD5` function. However, `MD5(NULL) -> NULL`, and `CONCAT(NULL, 'data') -> NULL`. This means every row that has a NULL column will have a NULL checksum, which is unacceptable.
- To get around this, we use the `COALESCE` function, which returns the first non null entry. Specifically, we used `MD5(COALASCE(column, 'NULL'))`. If the column is NULL, we would thus compute `MD5('NULL')`. This causes the collision with the string `'NULL'`.

The current workaround uses a magic string that's unlikely to be found in nature. Although it may be possible to use `CONCAT_WS` and `NULLIF` to accomplish this without the magic string. There are some reverted commits for reference. The commit at the end also adds a test case that will be triggered if the naive `CONCAT_WS` method is implemented in the future.

### DIFF
--- a/table_schema_cache.go
+++ b/table_schema_cache.go
@@ -81,10 +81,10 @@ func (t *TableSchema) RowMd5Query() string {
 
 	hashStrs := make([]string, len(columns))
 	for i, column := range columns {
-		hashStrs[i] = fmt.Sprintf("MD5(COALESCE(%s, 'NULL'))", normalizeAndQuoteColumn(column))
+		hashStrs[i] = fmt.Sprintf("MD5(%s)", normalizeAndQuoteColumn(column))
 	}
 
-	t.rowMd5Query = fmt.Sprintf("MD5(CONCAT(%s)) AS __ghostferry_row_md5", strings.Join(hashStrs, ","))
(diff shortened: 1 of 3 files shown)


shuhaowu:
I'm going to merge this for now. If there are any other concerns/comments about this, we can address it in a follow up PR.

Issue #96, opened 2019-02-14
Support iteration on non-PK unique integer column


ridvankaratas:
hi again

I got this error after fixing other things and wondered why its not supported? what is the deal with it or is there any planned feature about it?

anyway, we will try to fix it but wanted to hear from your side

hkdsun:
AFAIK, we don't use multi-column primary keys internally and there is no work planned on our part for this. 

I'll be happy to review a plan for this. I believe @shuhaowu had started a branch some time ago to contain all usage of primary keys in a struct and that will likely be where you want to start looking.

shuhaowu:
Supporting composite and non-uint64 PK has gotten more difficult I believe. That branch would have only supported single column strings as PK. Doing it in a more generic way requires some more work. 

There's no plan to do this on our side, but PRs are welcome and we'll review them.

shuhaowu:
We can potentially solve this issue as long as you have a pagination column (a unique integer column that can be iterated over in ascending order). Ghostferry needs to be changed to take a configurable pagination column and iterate over that column instead of the primary key. Any PRs to solve this issue should probably take a variant of this approach as oppose to locking onto the primary key.

jarthorn:
@shuhaowu we are really interested in this as part of our [Order database performance](https://vault.shopify.io/projects/9773) project. In this case we will be retaining an int `id` iteration column, it would just not be the primary key any more. Do you have any bandwidth to either take this on, or help us out with a high level sketch of how to fix this? We are willing to contribute someone to work on it if you have other priorities!

shuhaowu:
This is fixed on the latest master.

PR #138, opened 2019-10-07
Iterate by column

Ajwah:
This addresses [Support iteration on non-PK unique integer column](https://github.com/Shopify/ghostferry/issues/96)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@
 
 .vagrant/
 build/
+.bundle/
+.vscode/
--- a/config.go
+++ b/config.go
@@ -268,6 +268,37 @@ func (c ColumnIgnoreConfig) IgnoredColumnsFor(schemaName, tableName string) map[
 	return columnsConfig
 }
 
+// CascadingPaginationColumnConfig to configure pagination columns to be
+// used. The term `Cascading` to denote that greater specificity takes
+// precedence.
+type CascadingPaginationColumnConfig struct {
+	// PerTable has greatest specificity and takes precedence over the other options
+	PerTable map[string]map[string]string // SchemaName => TableName => ColumnName
+
+	// FallbackColumn is a global default to fallback to and is more specific than the default fallback,
+	// which would be the PK
+	FallbackColumn string
+}
+
+// PaginationColumnFor is a helper function to retrieve the column name to paginate by
+func (c *CascadingPaginationColumnConfig) PaginationColumnFor(schemaName, tableName string) (string, bool) {
+	if c == nil {
+		return "", false
+	}
+
+	tableConfig, found := c.PerTable[schemaName]
+	if !found {
+		return "", false
+	}
+
+	column, found := tableConfig[tableName]
+	if !found {
+		return "", false
+	}
+
+	return column, true
+}
+
 type Config struct {
 	// Source database connection configuration
 	//
@@ -433,6 +464,12 @@ type Config struct {
 	// - Putting the column in this config will cause the InlineVerifier to skip
 	//   this column for verification.
 	IgnoredColumnsForVerification ColumnIgnoreConfig
+
+	// Ghostferry requires a single numeric column to paginate over tables. Inferring that column is done in the following exact order:
+	// 1. Use the PerTable pagination column, if configured for a table. Fail if we cannot find this column in the table.
+	// 2. Use the FallbackColumn pagination column, if configured. Fail if we cannot find this column in the table.
+	// 3. Use the table's primary key column as the pagination column. Fail if the primary key is a composite key or is not numeric.
+	CascadingPaginationColumnConfig *CascadingPaginationColumnConfig
 }
 
 func (c *Config) ValidateConfig() error {
--- a/sharding/sharding.go
+++ b/sharding/sharding.go
@@ -213,7 +213,7 @@ func (r *ShardingFerry) copyPrimaryKeyTables() error {
 	r.config.TableFilter.(*ShardedTableFilter).PrimaryKeyTables = pkTables
 	r.config.CopyFilter.(*ShardedCopyFilter).PrimaryKeyTables = pkTables
 
-	sourceDbTables, err := ghostferry.LoadTables(r.Ferry.SourceDB, r.config.TableFilter, r.config.CompressedColumnsForVerification, r.config.IgnoredColumnsForVerification)
+	sourceDbTables, err := ghostferry.LoadTables(r.Ferry.SourceDB, r.config.TableFilter, r.config.CompressedColumnsForVerification, r.config.IgnoredColumnsForVerification, r.config.CascadingPaginationColumnConfig)
 	if err != nil {
 		return err
 	}
--- a/table_schema_cache.go
+++ b/table_schema_cache.go
@@ -36,6 +36,7 @@ type TableSchema struct {
 
 	CompressedColumnsForVerification map[string]string   // Map of column name => compression type
 	IgnoredColumnsForVerification    map[string]struct{} // Set of column name
+	PaginationKeyColumn              *schema.TableColumn
 
 	rowMd5Query string
 }
@@ -101,6 +102,10 @@ func (t *TableSchema) RowMd5Query() string {
 
 type TableSchemaCache map[string]*TableSchema
 
+func fullTableName(schemaName, tableName string) string {
+	return fmt.Sprintf("%s.%s", schemaName, tableName)
+}
+
 func QuotedTableName(table *TableSchema) string {
 	return QuotedTableNameFromString(table.Schema, table.Name)
 }
@@ -134,7 +139,7 @@ func MaxPrimaryKeys(db *sql.DB, tables []*TableSchema, logger *logrus.Entry) (ma
 	return tablesWithData, emptyTables, nil
 }
 
-func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig ColumnCompressionConfig, columnIgnoreConfig ColumnIgnoreConfig) (TableSchemaCache, error) {
+func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig ColumnCompressionConfig, columnIgnoreConfig ColumnIgnoreConfig, cascadingPaginationColumnConfig *CascadingPaginationColumnConfig) (TableSchemaCache, error) {
 	logger := logrus.WithField("tag", "table_schema_cache")
 
 	tableSchemaCache := make(TableSchemaCache)
@@ -189,18 +194,12 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
 			tableLog := dbLog.WithField("table", tableName)
 			tableLog.Debug("caching table schema")
 
-			// Sanity check
-			if len(tableSchema.PKColumns) != 1 {
-				err = fmt.Errorf("table %s has %d primary key columns and this is not supported", tableName, len(tableSchema.PKColumns))
+			paginationKeyColumn, err := tableSchema.paginationKeyColumn(cascadingPaginationColumnConfig)
+			if err != nil {
 				logger.WithError(err).Error("invalid table")
 				return tableSchemaCache, err
 			}
-
-			if tableSchema.GetPKColumn(0).Type != schema.TYPE_NUMBER {
-				err = fmt.Errorf("table %s is using a non-numeric primary key column and this is not supported", tableName)
-				logger.WithError(err).Error("invalid table")
-				return tableSchemaCache, err
-			}
+			tableSchema.PaginationKeyColumn = paginationKeyColumn
 
 			tableSchemaCache[tableSchema.String()] = tableSchema
 		}
@@ -209,6 +208,84 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
 	logger.WithField("tables", tableSchemaCache.AllTableNames()).Info("table schemas cached")
 
 	return tableSchemaCache, nil
+}
+
+func (t *TableSchema) findColumnByName(name string) (*schema.TableColumn, error) {
+	for _, column := range t.Columns {
+		if column.Name == name {
+			return &column, nil
+		}
+	}
+	return nil, NonExistingPaginationKeyColumnError(t.Schema, t.Name, name)
+}
+
+// NonExistingPaginationKeyColumnError exported to facilitate black box testing
+func NonExistingPaginationKeyColumnError(schema, table, paginationKey string) error {
+	return fmt.Errorf("Pagination Key `%s` for %s non existent", paginationKey, QuotedTableNameFromString(schema, table))
+}
+
+// NonExistingPKError exported to facilitate black box testing
+func NonExistingPKError(schema, table string) error {
+	return fmt.Errorf("%s has no Primary Key to default to for Pagination purposes. Kindly specify a Pagination Key for this table in the CascadingPaginationColumnConfig", QuotedTableNameFromString(schema, table))
+}
+
+// CompositePKError exported to facilitate black box testing
+func CompositePKError(schema, table string) error {
+	return fmt.Errorf("%s attempted to default to a Composite Primary Key for Pagination purposes whereas this is not supported. Kindly specify a Pagination Key for this table in the CascadingPaginationColumnConfig", QuotedTableNameFromString(schema, table))
+}
+
+// NonNumericPaginationKeyError exported to facilitate black box testing
+func NonNumericPaginationKeyError(schema, table, paginationKey string) error {
+	return fmt.Errorf("Pagination Key `%s` for %s is non-numeric", paginationKey, QuotedTableNameFromString(schema, table))
+}
+
+func (t *TableSchema) paginationKeyColumn(cascadingPaginationColumnConfig *CascadingPaginationColumnConfig) (*schema.TableColumn, error) {
+	var err error
+	var paginationKeyColumn *schema.TableColumn
+
+	if paginationColumn, found := cascadingPaginationColumnConfig.PaginationColumnFor(t.Schema, t.Name); found {
+		// Check that supplied pagination key column is in actual effect an existing column
+		paginationKeyColumn, err = t.findColumnByName(paginationColumn)
+	} else if cascadingPaginationColumnConfig == nil || cascadingPaginationColumnConfig.FallbackColumn == "" {
+		// default to Primary Key if possible
+		switch len(t.PKColumns) {
+		case 0:
+			err = NonExistingPKError(t.Schema, t.Name)
+		case 1:
+			paginationKeyColumn = &t.Columns[t.PKColumns[0]]
+		default:
+			err = CompositePKError(t.Schema, t.Name)
+		}
+	} else {
+		// Check that supplied pagination key column to fallback to is in actual effect an existing column
+		paginationKeyColumn, err = t.findColumnByName(cascadingPaginationColumnConfig.FallbackColumn)
+	}
+
+	if paginationKeyColumn != nil && paginationKeyColumn.Type != schema.TYPE_NUMBER {
+		return nil, NonNumericPaginationKeyError(t.Schema, t.Name, paginationKeyColumn.Name)
+	}
+
+	return paginationKeyColumn, err
+}
+
+/*
+GetPKColumn is a method originally belonging to go-mysql/schema.Table
+and stands for: `Get Primary Key Column`
+
+As `TableSchema` embeds `schema.Table`, previously the method call to
+`TableSchema` would delegate automatically to the method for
+`schema.Table`. This was in spirit with the initial assumption that the
+column to iterate over in any given table is the Primary Key Column.
+
+In order to accommodate user provided pagination columns through
+configuration in the least painful way, this method has been
+reimplemented. Now this method stands for: `Get Pagination Key Column`.
+
+The param `index` currently serves no purpose but to maintain the same
+method signature as the original method.
+*/
+func (t *TableSchema) GetPKColumn(index int) *schema.TableColumn {
+	return t.PaginationKeyColumn
 }
 
 func (c TableSchemaCache) AsSlice() (tables []*TableSchema) {
@@ -228,8 +305,7 @@ func (c TableSchemaCache) AllTableNames() (tableNames []string) {
 }
 
 func (c TableSchemaCache) Get(database, table string) *TableSchema {
-	fullTableName := fmt.Sprintf("%s.%s", database, table)
-	return c[fullTableName]
+	return c[fullTableName(database, table)]
 }
 
 func showDatabases(c *sql.DB) ([]string, error) {
@@ -282,7 +358,6 @@ func showTablesFrom(c *sql.DB, dbname string) ([]string, error) {
 func maxPk(db *sql.DB, table *TableSchema) (uint64, bool, error) {
 	primaryKeyColumn := table.GetPKColumn(0)
 	pkName := quoteField(primaryKeyColumn.Name)
-
 	query, args, err := sq.
 		Select(pkName).
 		From(QuotedTableName(table)).
--- a/test/go/table_schema_cache_test.go
+++ b/test/go/table_schema_cache_test.go
@@ -19,6 +19,23 @@ type TableSchemaCacheTestSuite struct {
 
 var ()
 
+func dropTestTables(this *TableSchemaCacheTestSuite) {
+	var query string
+	var err error
+	for _, tablename := range this.tablenames {
+		query = fmt.Sprintf("DROP TABLE %s.%s", testhelpers.TestSchemaName, tablename)
+		_, err = this.Ferry.SourceDB.Exec(query)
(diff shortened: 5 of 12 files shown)


table_schema_cache.go
```diff
@@ -189,10 +190,7 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
 			tableLog := dbLog.WithField("table", tableName)
 			tableLog.Debug("caching table schema")
 
-			// Sanity check
-			if len(tableSchema.PKColumns) != 1 {
-				err = fmt.Errorf("table %s has %d primary key columns and this is not supported", tableName, len(tableSchema.PKColumns))
-				logger.WithError(err).Error("invalid table")
+			if err := tableSchema.setPaginationKeyColumn(dbname, tableName, paginateByColumnConfig); err != nil {
 				return tableSchemaCache, err
 			}
```
hkdsun: few lines down, we have the error message "table ... is using a non-numeric primary key column and this is not supported"

could you reword that?

table_schema_cache.go
```diff
@@ -189,10 +190,7 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
 			tableLog := dbLog.WithField("table", tableName)
 			tableLog.Debug("caching table schema")
 
-			// Sanity check
-			if len(tableSchema.PKColumns) != 1 {
-				err = fmt.Errorf("table %s has %d primary key columns and this is not supported", tableName, len(tableSchema.PKColumns))
-				logger.WithError(err).Error("invalid table")
+			if err := tableSchema.setPaginationKeyColumn(dbname, tableName, paginateByColumnConfig); err != nil {
```
hkdsun: A little philosophical but does it make sense to figure out the column before even constructing the new `TableSchema`s?

i.e. on L176 we have `tableSchemas = append(tableSchemas, &TableSchema{...})`. We could infer the pagination column before even calling `append`

table_schema_cache.go
```diff
@@ -211,6 +209,70 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
...
+In order to accommodate user provided pagination columns through
+configuration in the least painful way, this method has been
+reimplemented. Now this method stands for: `Get Pagination Key Column`.
+
+The param `index` currently serves no purpose but to maintain the same
+method signature as the original method.
+*/
+func (ts *TableSchema) GetPKColumn(index int) *schema.TableColumn {
+	return ts.PaginationKeyColumn
+}
```
hkdsun: This is clever but easily confused with primary key. We discussed this in Slack. You suggested two followup cleanup PRs but would prefer to not work on the follow-ups yourself:

* Rename this method to something like `GetPaginationColumn`
* Rename references like `minPK`, `maxPK`, etc

table_schema_cache.go
```diff
@@ -211,6 +209,70 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
...
+		default:
+			ts.PaginationKeyColumn = ts.findColumnByName("id")
+			err = LackPaginationKeyError(dbname, table, true)
+		}
+	}
+	if ts.PaginationKeyColumn == nil {
+		return err
+	}
+
+	return nil
```
hkdsun: This whole method is a bit weird.. I'd prefer if it didn't set `ts.PaginationKeyColumn` directly but only inferred what it should be based on configuration. i.e. lift up the field assignment one level up.

I would then rewrite it as something like:

```
func (ts *TableSchema) setPaginationKeyColumn(dbname, table string, paginateByColumnConfig PaginateByColumnConfig) (*schema.TableColumn, error) {
    if paginateByColumnConfig == nil {
        return inferPaginationColumn(ts, dbname, table)
    } else {
        // ...
    } 
```

.gitignore
```diff
@@ -15,3 +15,5 @@
 
 .vagrant/
 build/
+.bundle/
+.vscode/
```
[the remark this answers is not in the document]
shuhaowu: Or alternatively it can go into .git/info/exclude for each repo.

config.go
```diff
@@ -433,6 +451,8 @@ type Config struct {
 	// - Putting the column in this config will cause the InlineVerifier to skip
 	//   this column for verification.
 	IgnoredColumnsForVerification ColumnIgnoreConfig
+
+	PaginateByColumnConfig PaginateByColumnConfig
```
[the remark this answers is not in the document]
Ajwah: What about:
```
/* Configuration option in which you can specify an alternate column to paginate over as opposed to the default Primary Key Column. This is helpful when:
-1 Primary Key column does not comply to the requirement of being auto incrementing, numeric and unique
-2 Primary Key column is a composite
*/ 
PaginateByColumnConfig PaginateByColumnConfig
```

table_schema_cache.go
Ajwah: There are multiple perspectives to this:
1. Should we adopt an optimistic or pessimistic outlook on user supplied configuration? For the optimistic outlook, you make the assumption that the user supplied sane configuration unless proven otherwise. For the pessimistic outlook you make the inverse assumption. 

If you adopt the optimistic outlook, then you basically want to make the usage as permissible as possible and thus you accommodate for the scenario where the user supplied incorrect configuration but which never harms the system as it is not being applied. e.g.:
The line: `tableSchemas, err = tableFilter.ApplicableTables(tableSchemas)` filters out any tables that are ignored. In the case that the user supplied incorrect configuration for a table that he has blacklisted, then because it gets filtered out, the subsequent loop will never error out for `tableSchema.setPaginationKeyColumn` as the loop simply skips that table.

In the pessimistic outlook, you would want to run the sanity checks for all tables, whether or not blacklisted. 

It is the same 'being-optimistic-vs-pessimistic-about-user-input' debate you would have with regards to the design of a programming language whether you want it to be a dynamically typed language or a statically typed one. 
 
Which outlook do you feel is best for this project? From my side, I mainly placed it here as the previous sanity check hinted to me that the project adopted an optimistic outlook on user configuration supplied:

```
if len(tableSchema.PKColumns) != 1 {	
				err = fmt.Errorf("table %s has %d primary key columns and this is not supported", tableName, len(tableSchema.PKColumns))	
				logger.WithError(err).Error("invalid table")	
				return tableSchemaCache, err
```

Likewise the function introduced by me `setPaginationKeyColumn` adopts the `optimistic` outlook by it only running on the tables that are effectively whitelisted

In a `pessimistic` outlook I would promote the same to the other loop so it will always error out despite the table being blacklisted.

2. How explicit do we want to be with our code base to facilitate readability at first glance?
We declared a type: 
```
type TableSchema struct {
	*schema.Table

	CompressedColumnsForVerification map[string]string   // Map of column name => compression type
	IgnoredColumnsForVerification    map[string]struct{} // Set of column name
	PaginationKeyColumn *schema.TableColumn

	rowMd5Query string
}
```

If we want to be explicit, then the following code:
```			
tableSchemas = append(tableSchemas, &TableSchema{
				Table: tableSchema,
				CompressedColumnsForVerification: columnCompressionConfig.CompressedColumnsFor(dbname, table),
				IgnoredColumnsForVerification: columnIgnoreConfig.IgnoredColumnsFor(dbname, table),
			})
```
can be a source of confusion where a developer will have to look through the code base to see where the missing fields are being set. It would be more clearer to have `PaginationKeyColumn` to be explicitly set as:
```			tableSchemas = append(tableSchemas, &TableSchema{
				Table: tableSchema,
				CompressedColumnsForVerification: columnCompressionConfig.CompressedColumnsFor(dbname, table),
				IgnoredColumnsForVerification: columnIgnoreConfig.IgnoredColumnsFor(dbname, table),
				PaginationKeyColumn: paginationKeyColumn,
			})
```
If we adopt the pessimistic outlook, then we could state:
```
if paginationKeyColumn, err = GetPaginationKeyColumn(...); err != nil {
...
}
```
If we adopt the optimistic outlook then we could state:
```			
tableSchemas = append(tableSchemas, &TableSchema{
				Table: tableSchema,
				CompressedColumnsForVerification: columnCompressionConfig.CompressedColumnsFor(dbname, table),
				IgnoredColumnsForVerification: columnIgnoreConfig.IgnoredColumnsFor(dbname, table),
				PaginationKeyColumn: nil // will be set shortly,
			})
```
Do you feel that being explicit facilitates better code maintainability? 
What about the key: `rowMd5Query`? Should we also set it explicitly:
```			
tableSchemas = append(tableSchemas, &TableSchema{
				Table: tableSchema,
				CompressedColumnsForVerification: columnCompressionConfig.CompressedColumnsFor(dbname, table),
				IgnoredColumnsForVerification: columnIgnoreConfig.IgnoredColumnsFor(dbname, table),
				PaginationKeyColumn: nil // will be set shortly,
                                   rowMd5Query: "" // will be set later when verification being done,
			})
```

Lastly, I made `PaginationKeyColumn` public. Would it make sense for it to be private instead?

.gitignore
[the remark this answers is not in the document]
Ajwah: If this is fine then I will keep it as is.

table_schema_cache.go
Ajwah: Will address this in the next PR

table_schema_cache.go
Ajwah: Will address this as soon as the above discussion is resolved.

table_schema_cache.go
```diff
@@ -189,15 +190,12 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
...
 
-			// Sanity check
-			if len(tableSchema.PKColumns) != 1 {
-				err = fmt.Errorf("table %s has %d primary key columns and this is not supported", tableName, len(tableSchema.PKColumns))
-				logger.WithError(err).Error("invalid table")
+			if err := tableSchema.setPaginationKeyColumn(dbname, tableName, paginateByColumnConfig); err != nil {
 				return tableSchemaCache, err
 			}
 
 			if tableSchema.GetPKColumn(0).Type != schema.TYPE_NUMBER {
```
shuhaowu: I know that you override the `GetPKColumn` and I spoke with you about it, but this is pretty confusing now that I see it. I think we should change this if it is not too difficult.

table_schema_cache.go
shuhaowu: I agree with @hkdsun on this point.

shuhaowu:
The failing test seems real:

```
    --- FAIL: TestTableSchemaCache/TestLoadTablesRejectTablesWithoutNumericPK (0.15s)
	Error Trace:	table_schema_cache_test.go:76
	Error:      	"table test_table_4 is using a non-numeric pagination key column and this is not supported" does not contain "non-numeric primary key"
```

FYI: the best way to find a failing Go test is to search for the word `FAIL` with match case and whole word turned on.

table_schema_cache.go
Ajwah: Yes I agree. I will change it as a separate pull request. I want to keep this pull request relevant to the functionality required.

table_schema_cache.go
hkdsun: Reading the code again, the whole `LoadTables` function needs a cleanup and this PR is not making it any worse than before. I don't really wanna block the PR on this discussion so it's fine as is, though improving the readability of the `setPaginationKeyColumn` function still seems like a good idea.

table_schema_cache.go
```diff
@@ -189,15 +190,14 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
...
 				return tableSchemaCache, err
 			}
+			tableSchema.PaginationKeyColumn = paginationKeyColumn
 
 			if tableSchema.GetPKColumn(0).Type != schema.TYPE_NUMBER {
-				err = fmt.Errorf("table %s is using a non-numeric primary key column and this is not supported", tableName)
+				err = fmt.Errorf("table %s is using a non-numeric pagination key column and this is not supported", tableName)
 				logger.WithError(err).Error("invalid table")
 				return tableSchemaCache, err
 			}
```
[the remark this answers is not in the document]
shuhaowu: This code will need some cleanup anyway, after we cleanup `ApplicableTables`. If no one gets to it I'll make it happen as it is a simple enough change.

table_schema_cache.go
```diff
@@ -211,6 +211,77 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
...
+
+func (ts *TableSchema) paginationKeyColumn(dbname, table string, cascadingPaginationColumnConfig *CascadingPaginationColumnConfig) (*schema.TableColumn, error) {
+	var err error
+	var paginationKeyColumn *schema.TableColumn
+
+	if paginationColumn, found := cascadingPaginationColumnConfig.PaginationColumnFor(dbname, table); found {
+		paginationKeyColumn = ts.findColumnByName(paginationColumn)
+		err = NonExistingPaginationKeyColumnError(dbname, table, paginationColumn)
+	} else {
+		if cascadingPaginationColumnConfig == nil || cascadingPaginationColumnConfig.FallbackColumn == "" {
```
shuhaowu: Instead of nesting like this, can you just have:

```
if cascadingPAginationColumnConfig == nil {
  paginationKeyColumn = findViaPK
} else if cascadingPaginationColumnConfig.FallbackColumn == "" {
  paginationKEyColumn = findViaFallback
} else if paginationColumn, found := cascadingPaginationColumnConfig.PaginationColumnFor(dbname, table); found {
  ...
} else {
  // do we need this clause or does the above already cover all possibilities?
}
```

table_schema_cache.go
```diff
@@ -211,6 +211,77 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
...
+			default:
+				err = CompositePKError(dbname, table)
+			}
+		} else {
+			paginationKeyColumn = ts.findColumnByName(cascadingPaginationColumnConfig.FallbackColumn)
+			err = NonExistingPaginationKeyColumnError(dbname, table, cascadingPaginationColumnConfig.FallbackColumn)
+		}
+	}
+
+	if paginationKeyColumn == nil {
```
shuhaowu: `if err != nil { return err }`?

shuhaowu: Actually you could likely get rid of this whole block and just `return paginationKeyColumn, err`

Ajwah: I currently need it because I am always populating `err`

table_schema_cache.go
Ajwah: This cannot be split up as you suggested:
```
if cascadingPAginationColumnConfig == nil {
  paginationKeyColumn = findViaPK
} else if cascadingPaginationColumnConfig.FallbackColumn == "" {
  paginationKEyColumn = findViaFallback
```
as both cases need to `findViaPK`. 
So the following would work:
if cascadingPAginationColumnConfig == nil {
  paginationKeyColumn = findViaPK
} else if cascadingPaginationColumnConfig.FallbackColumn == "" {
  paginationKEyColumn = findViaPK
```
but I am not sure that this increases clarity.

Nevertheless, I can introduce more helper functions if this is deemed important to understand that the checks being done need to reflect the cases as documented at: https://github.com/Shopify/ghostferry/blob/090abdce25b3821280e0503a5b6d26be0f05d423/config.go#L271-L281

Comment (shuhaowu) [reply table_schema_cache.go]: I meant to write:

```
if cascadingPAginationColumnConfig == nil || cascadingPaginationColumnConfig.FallbackColumn == "" {
  paginationKeyColumn = findViaPK
} else if cascadingPaginationColumnConfig.FallbackColumn != "" {
  paginationKEyColumn = findViaFallbackCol
```

I was just suggesting that we turn this method flat (right now it nested 2 levels deep). If you think it's still not clear, you add some inline comments as well.

File: table_schema_cache.go (hunk 10 above)
Comment (shuhaowu) [reply table_schema_cache.go]: I feel like this style is not very common in Go. Looking at the `findColumnByName` method and how it is invoked, it seems to me that method should return a (tablecol, err) tuple, where err == nil if a column can be found.

File: table_schema_cache.go [hunk 11]
```diff
@@ -232,33 +226,44 @@ func CompositePKError(schema, table string) error {
...
+		err = NonExistingPaginationKeyColumnError(dbname, table, cascadingPaginationColumnConfig.FallbackColumn)
 	}
 
 	if paginationKeyColumn == nil {
 		return nil, err
 	}
 
+	if paginationKeyColumn.Type != schema.TYPE_NUMBER {
+		err = NonNumericPaginationKeyError(dbname, table, paginationKeyColumn.Name)
+		logger.WithError(err).Error("invalid table")
```
shuhaowu: I don't think you actually need to log in this here. Probably should log the overall error at line 195, otherwise I don't think any of the other errors will be logged?

hkdsun:
Awesome. Thanks for the contribution and bearing with us through the reviews @Ajwah 🥇 

Squash and `gofmt ./...` before merging pls 🙏

table_schema_cache.go
Ajwah: Good catch!

config.go
```diff
@@ -268,6 +268,37 @@ func (c ColumnIgnoreConfig) IgnoredColumnsFor(schemaName, tableName string) map[
...
+
+	tableConfig, found := c.PerTable[schemaName]
+	if !found {
+		return "", false
+	}
+
+	column, found := tableConfig[tableName]
+	if !found {
+		return "", false
+	}
```
shuhaowu: Minor nit: probably can get rid of the boolean and just check if `column == ""`.

sharding/sharding.go
```diff
@@ -213,7 +213,7 @@ func (r *ShardingFerry) copyPrimaryKeyTables() error {
 	r.config.TableFilter.(*ShardedTableFilter).PrimaryKeyTables = pkTables
 	r.config.CopyFilter.(*ShardedCopyFilter).PrimaryKeyTables = pkTables
 
-	sourceDbTables, err := ghostferry.LoadTables(r.Ferry.SourceDB, r.config.TableFilter, r.config.CompressedColumnsForVerification, r.config.IgnoredColumnsForVerification)
+	sourceDbTables, err := ghostferry.LoadTables(r.Ferry.SourceDB, r.config.TableFilter, r.config.CompressedColumnsForVerification, r.config.IgnoredColumnsForVerification, r.config.CascadingPaginationColumnConfig)
```
shuhaowu: Question: this is calling to the PrimaryKeyTables, which are tables where there's only one entry consists of the sharding key. Does this need this option passed in? Should we explicitly not pass it in? @hkdsun 

(This is not an objection, I just want to make sure we're doing the right thing as I'm less familiar with PrimaryKeyTables).

table_schema_cache.go
```diff
@@ -211,6 +206,81 @@ func LoadTables(db *sql.DB, tableFilter TableFilter, columnCompressionConfig Col
 	return tableSchemaCache, nil
 }
 
+func (t *TableSchema) findColumnByName(name string, err error) (*schema.TableColumn, error) {
+	for _, column := range t.Columns {
+		if column.Name == name {
+			return &column, nil
+		}
+	}
+	return nil, err
```
shuhaowu: At first I thought this `err` is always `nil` because no where in this function is it initialized. Then I see that you're passing the same `NonExistingPaginationKeyColumnError(dbname, table, name)` in as an `err` argument and returning it. 

We can just build that error in this function directly. dbname and table are all information available directly on the  TableSchema struct that this is a method of.

test/go/table_schema_cache_test.go
```diff
@@ -65,30 +76,159 @@ func (this *TableSchemaCacheTestSuite) TestLoadTablesWithoutFiltering() {
...
+	table := "test_table_4"
+	paginationColumn := "id"
+	query := fmt.Sprintf("CREATE TABLE %s.%s (%s varchar(20) not null, data TEXT, primary key(%s))", testhelpers.TestSchemaName, table, paginationColumn, paginationColumn)
+	_, err := this.Ferry.SourceDB.Exec(query)
+	this.Require().Nil(err)
+
+	_, err = ghostferry.LoadTables(this.Ferry.SourceDB, this.tableFilter, nil, nil, nil)
+
+	this.Require().NotNil(err)
+	this.Require().EqualError(err, ghostferry.NonNumericPaginationKeyError(testhelpers.TestSchemaName, table, paginationColumn).Error())
```
shuhaowu: TIL `EqualError`

test/go/table_schema_cache_test.go
```diff
@@ -65,30 +76,159 @@ func (this *TableSchemaCacheTestSuite) TestLoadTablesWithoutFiltering() {
...
+				table: paginationColumn,
+			},
+		},
+	}
+
+	query := fmt.Sprintf("CREATE TABLE %s.%s (%s bigint(20) not null, data TEXT)", testhelpers.TestSchemaName, table, paginationColumn)
+	_, err := this.Ferry.SourceDB.Exec(query)
+	this.Require().Nil(err)
+
+	_, err = ghostferry.LoadTables(this.Ferry.SourceDB, this.tableFilter, nil, nil, cascadingPaginationColumnConfig)
```
shuhaowu: Maybe we can quickly assert that the table `pagination_by_column_config_right_scenario_1` does indeed have the right PaginationColumn set? Is this necessary?

test/go/table_schema_cache_test.go
```diff
@@ -65,30 +76,159 @@ func (this *TableSchemaCacheTestSuite) TestLoadTablesWithoutFiltering() {
...
+
+	this.Require().Nil(err)
+}
+func (this *TableSchemaCacheTestSuite) TestLoadTablesCascadingPaginationColumnConfigRightScenario2() {
+	dropTestTables(this) // needed because they interfere with this test
```
shuhaowu: Can you elaborate what "they" means? It's unclear to me. Shouldn't this method be in the test setup method instead?

test/go/table_schema_cache_test.go
```diff
@@ -65,30 +76,159 @@ func (this *TableSchemaCacheTestSuite) TestLoadTablesWithoutFiltering() {
...
+	paginationColumn := "identity"
+	cascadingPaginationColumnConfig := &ghostferry.CascadingPaginationColumnConfig{
+		FallbackColumn: paginationColumn,
+	}
+
+	query := fmt.Sprintf("CREATE TABLE %s.%s (%s bigint(20) not null, data TEXT)", testhelpers.TestSchemaName, table, paginationColumn)
+	_, err := this.Ferry.SourceDB.Exec(query)
+	this.Require().Nil(err)
+
+	_, err = ghostferry.LoadTables(this.Ferry.SourceDB, this.tableFilter, nil, nil, cascadingPaginationColumnConfig)
```
shuhaowu: Ditto on asserting that the PaginationColumn is set correctly?

test/go/table_schema_cache_test.go
Ajwah: The test setup introduces different tables: `test_table_1, _2, _3, and _4` Those tables are interfering with this test

shuhaowu: I'm a bit confused why the following code doesn't already address this problem as it drops the entire schema.

https://github.com/Shopify/ghostferry/blob/b8147c5ff8653099f6c5e24128aa6fd41ff401e1/testhelpers/unit_test_suite.go#L81-L90

Ajwah: I only need to drop the tables. The database can remain. 
If you feel that there is merit in this helper function, would you suggest me to promote it to `unit_test_suite.go`?

table_schema_cache.go
Ajwah: Nice suggestion! That access would allow me to also simplify `paginationKeyColumn`-method even further

test/go/table_schema_cache_test.go
shuhaowu: No I meant the code I linked should have ran before your test so you shouldn't need to drop additional tables since you are using the same schema. If that is the case you shouldn't have needed this function at all.

Just leave the code here for now if you dont want to dig. We can make a note to deal with it later.

Ajwah: I am not following yet. The TestSetup does: https://github.com/Shopify/ghostferry/blob/97d62446c2f7d978cb6e727e3024bdddd27f4e3f/test/go/table_schema_cache_test.go#L42-L45
which always is called by: https://github.com/Shopify/ghostferry/blob/97d62446c2f7d978cb6e727e3024bdddd27f4e3f/test/go/table_schema_cache_test.go#L338-L341 before each test. 

So by the time the test: `TestLoadTablesCascadingPaginationColumnConfigRightScenario2()` is invoked, those 4 tables are present and populated.

I cannot run `dropTestTables` inside the setup because other tests rely on the existence of those tables, e.g. `test_table_1, _2, _3, and _4`. 

What am I misunderstanding?

sroysen:
I was talking with @shuhaowu about this PR.
It occurred to me that based on the use case this change was aiming to solve, where the `id` of a table would still be used as the rightmost element of the primary/clustered key, and where the id would still be an auto increment integer, ghostferry automatically identify any auto increment field in a table, and use it to iterate over that table (innodb only allows one field of type auto increment per table): That would cover all cases: The traditional one where id is the primary key, and the case where you have a composite primary key, but where you still have a field of type auto increment.

PR #142, opened 2019-10-21
Ruby Integration Test for iterate by column

Ajwah:


test/integration/iterate_by_column_test.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  private
+  def create_tables(db_name, table_name, create_query)
+    [source_db, target_db].each do |db|
+      db.query("CREATE DATABASE IF NOT EXISTS #{db_name}")
+      db.query("CREATE TABLE IF NOT EXISTS #{DbHelper.full_table_name(db_name, table_name)} #{create_query}")
+    end
+  end
+
+  def seed_dynamically(full_table_name, columns, amount_rows:)
```
Ajwah: If this is deemed to make sense, then I will promote as helper function like: https://github.com/Shopify/ghostferry/blob/b5a11f7f8b8ae58d8811653e3a310e0bdcd2e87f/test/helpers/db_helper.rb#L88

test/helpers/ghostferry_helper.rb
```diff
@@ -231,6 +231,10 @@ def start_ghostferry(resuming_state = nil)
           environment["GHOSTFERRY_IGNORED_COLUMN"] = @config[:ignored_column]
         end
 
+        if @config[:cascading_pagination_column_config]
+          environment["GHOSTFERRY_CASCADING_PAGINATION_COLUMN_CONFIG"] = @config[:cascading_pagination_column_config]
+        end
+
```
shuhaowu: Not for you, but at some point this needs to be cleaned up.

test/integration/iterate_by_column_test.rb
```diff
@@ -0,0 +1,118 @@
...
+      FallbackColumn: fallback_column
+    }.reduce({}) do |a,(k,v)|
+      if v
+        a.merge!({k => v})
+      end
+      a
+    end.to_json
+  end
+
+  def assert_source_target_eql_data(full_table_name)
```
Ajwah: This has been introduced as https://github.com/Shopify/ghostferry/blob/85f2a57620c9580d280162c9081aeb27dc822fda/test/test_helper.rb#L84 requires a bit extra customization, in terms of a parameter for a different table(which is trivial) and in terms of the specificity of: https://github.com/Shopify/ghostferry/blob/b5a11f7f8b8ae58d8811653e3a310e0bdcd2e87f/test/helpers/db_helper.rb#L126 which has other tests depending on it

test/integration/iterate_by_column_test.rb
```diff
@@ -0,0 +1,118 @@
...
+        table_name => pagination_key_column,
+      },
+    })
+    assert_source_target_eql_data(DbHelper.full_table_name(db_name, table_name))
+  end
+
+  def test_specifying_pagination_column_by_fallback_column_succeeds
+    pagination_key_column = 'some_column'
+    db_name = DEFAULT_DB
+    table_name = DEFAULT_TABLE
```
shuhaowu: Are these variables assignments necessary?

test/integration/iterate_by_column_test.rb
shuhaowu: We can leave it here for now until we need it somewhere else.

shuhaowu:
This is low risk enough to be merged as is.

PR #146, opened 2020-01-17
Do not record InlineVerifier pos in state dump if not verifying

shuhaowu:
If the VerifierType is not Inline, there's no point in recording the InlineVerifier binlog position in the state dump. This is done right now because the code records the binlog position at the beginning of a run. This actually somewhat misleading as it makes people using Ghostferry without verification might think the InlineVerifier is running due to its presence in the state dump.

### DIFF
--- a/test/go/state_tracker_test.go
+++ b/test/go/state_tracker_test.go
@@ -38,6 +38,32 @@ func (s *StateTrackerTestSuite) TestMinBinlogPosition() {
 		},
 	}
 	s.Require().Equal(serializedState.MinBinlogPosition(), mysql.Position{"mysql-bin.00002", 10})
+
+	serializedState = &ghostferry.SerializableState{
+		LastWrittenBinlogPosition: mysql.Position{
+			Name: "",
+			Pos:  0,
+		},
+
+		LastStoredBinlogPositionForInlineVerifier: mysql.Position{
+			Name: "mysql-bin.00002",
+			Pos:  10,
+		},
+	}
+	s.Require().Equal(serializedState.MinBinlogPosition(), mysql.Position{"mysql-bin.00002", 10})
+
+	serializedState = &ghostferry.SerializableState{
+		LastStoredBinlogPositionForInlineVerifier: mysql.Position{
+			Name: "",
+			Pos:  0,
+		},
+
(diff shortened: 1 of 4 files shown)


test/go/state_tracker_test.go
```diff
@@ -38,6 +38,19 @@ func (s *StateTrackerTestSuite) TestMinBinlogPosition() {
...
+			Name: "",
+			Pos:  0,
+		},
+
+		LastStoredBinlogPositionForInlineVerifier: mysql.Position{
+			Name: "mysql-bin.00002",
+			Pos:  10,
+		},
+	}
+	s.Require().Equal(serializedState.MinBinlogPosition(), mysql.Position{"mysql-bin.00002", 10})
```
fjordan: This is a good case to cover, but with regards to the changes in this PR, I would have expected the `LastStoredBinlogPositionForInlineVerifier` to be empty and for `LastWrittenBinlogPosition` to have the values of `mysql.Position{"mysql-bin.00002", 10}` given we're now only setting `LastStoredBinlogPositionForInlineVerifier` if the `InlineVerifier` is being used.

test/integration/interrupt_resume_test.rb
```diff
@@ -105,6 +105,26 @@ def test_interrupt_resume_when_table_has_completed
...
+      batches_written += 1
+      if batches_written >= 2
+        ghostferry.term_and_wait_for_exit
+      end
+    end
+
+    dumped_state = ghostferry.run_expecting_interrupt
+    assert_basic_fields_exist_in_dumped_state(dumped_state)
+    assert_equal "", dumped_state["LastStoredBinlogPositionForInlineVerifier"]["Name"]
+    assert_equal 0, dumped_state["LastStoredBinlogPositionForInlineVerifier"]["Pos"]
```
fjordan: Probably also want to also ensure `LastWrittenBinlogPosition` contains the correct values here

shuhaowu: It's actually pretty difficult to assert this because the binlog position is not something I can easily control.

fjordan: Good point. I guess we could refute nil values then

PR #151, opened 2020-01-22
Added more JSON types to integration test

shuhaowu:
Apparently valid json includes: `"null", "true", "false", "<any number>"`

This adds it into the integration test and test all statements (INSERT/UPDATE/DELETE).

However, there is apparently a case where a `"null"` value can be inserted but show up as an empty string in the binlog. See https://github.com/siddontang/go-mysql/pull/267 and https://github.com/siddontang/go-mysql/pull/266 (in Chinese). I can't reproduce this locally in the test case although a workaround should be provided by @danieloliveira079 in a separate PR.


danieloliveira079:
I will investigate that a bit further and share my findings later on

PR #152, opened 2020-01-22
Fixing panic encountered in BinlogWriter when operations are performed on the tables with JSON NOT NULL columns

shuhaowu:
The bug this attempts to workaround is here: https://bugs.mysql.com/bug.php?id=98496

This effectively implements the suggested fix as above.. except this occurs during binlog copying from source to target. During the data copy process, we've already coerced `''` into `'null'` as mysql server returns `'null'` when it sees `''` internally.

Other related bugs:

- https://github.com/siddontang/go-mysql/issues/267 
- https://github.com/siddontang/go-mysql/issues/266

### DIFF
--- a/dml_events.go
+++ b/dml_events.go
@@ -423,6 +423,11 @@ func appendEscapedString(buffer []byte, value string) []byte {
 
 func appendEscapedBuffer(buffer, value []byte, isJSON bool) []byte {
 	if isJSON {
+		// See https://bugs.mysql.com/bug.php?id=98496
+		if len(value) == 0 {
+			value = []byte("null")
+		}
+
 		buffer = append(buffer, "CAST("...)
 	} else {
 		buffer = append(buffer, "_binary"...)
--- a/test/integration/types_test.rb
+++ b/test/integration/types_test.rb
@@ -10,11 +10,7 @@ class TypesTest < GhostferryTestCase
   JSON_NUMBER = '42'
 
   def test_json_colum_not_null_with_no_default_is_invalid_this_is_fine
-    # See:
-    # - https://github.com/mysql/mysql-server/commit/f538d494292127768094084a082997b94a96f2db
-    # - https://bugs.mysql.com/bug.php?id=87734
-    # - https://github.com/siddontang/go-mysql/issues/266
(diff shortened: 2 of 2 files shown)


dml_events.go
```diff
@@ -423,6 +423,11 @@ func appendEscapedString(buffer []byte, value string) []byte {
 
 func appendEscapedBuffer(buffer, value []byte, isJSON bool) []byte {
 	if isJSON {
+		// TODO: link to issue
+		if len(value) == 0 {
+			value = []byte("null")
```
shuhaowu: This would be changing the actual data on MySQL from the "invalid null value" that it stores to a real `'null'` value. I think this is fine because half of MySQL already treats the "invalid null value" as `'null'` and one of the only place this is visible is within the replication.

shuhaowu: > However, MySQL will silently change this to a JSON-type null value instead of an actual NULL or empty value.

Not quite. MySQL will internally store the empty string and when will transform it to the JSON `'null'` when serving the value to the consumer. It just didn't do this conversion in the binlog in 5.7.

fjordan: Yes, I omitted the `on read` part from the end. Thanks!

test/integration/types_test.rb
```diff
@@ -9,6 +9,82 @@ class TypesTest < GhostferryTestCase
...
+      # Need to make sure we don't flush binlogs until we affirmatively see the
+      # 3 rows on the target and issue the DELETE statements
+      start = Time.now
+
+      loop do
+        sleep 0.1
+        res = nonstrict_target_db.query("SELECT COUNT(*) AS cnt FROM #{DEFAULT_FULL_TABLE_NAME}")
+        if res.first["cnt"] == 4
+          nonstrict_source_db.query("UPDATE #{DEFAULT_FULL_TABLE_NAME} SET data = 'true' WHERE id = 2")
+          nonstrict_source_db.query("UPDATE #{DEFAULT_FULL_TABLE_NAME} SET data = 'null' WHERE id = 3")
```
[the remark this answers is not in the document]
shuhaowu: Already covered in a separate test added before this PR: https://github.com/Shopify/ghostferry/blob/master/test/integration/types_test.rb#L202

PR #153, opened 2020-02-05
[WIP] Make Data Iterator Concurrency Aware

danieloliveira079:
This PR is an experiment that our team is implementing and it is not ready for review or it is intended to be merged into the master branch. It still requires more testing and refactoring.

The idea here is to test if it would be possible to leverage the power of ghostfery and copydb to speed up the data transfer between source and target databases using concurrency.

All the core business logic were kept. That includes the inline-verifier, fingerprints, binlog streamer and writer, etc.

At this point, we are just playing and trying different approaches and implementations. **This code is not meant to be used in production environments**.

Our team is running a few experiments and gathering data to be analysed. The tool itself does not provide much in terms of metrics. Consequently, insights are coming from other sources.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@
 build/
 .bundle/
 .vscode/
+
+.idea/
+bin/
--- a/config.go
+++ b/config.go
@@ -77,7 +77,7 @@ func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
 			return nil, fmt.Errorf("failed to build TLS config: %v", err)
 		}
 
-		cfgName := fmt.Sprintf("%s@%s:%s", c.User, c.Host, c.Port)
+		cfgName := fmt.Sprintf("%s@%s:%d", c.User, c.Host, c.Port)
 
 		err = mysql.RegisterTLSConfig(cfgName, tlsConfig)
 		if err != nil {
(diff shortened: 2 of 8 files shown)


shuhaowu:
Wow thanks so much for this work. 

> The tool itself does not provide much in terms of metrics. 

What kind of metrics are you interested in? There's a `/progress` callback that Ghostferry exposes (not sure about the control server). The control server is a bit out of date with the rest of the code base and could be updated.

danieloliveira079:
> Wow thanks so much for this work.
> 
> > The tool itself does not provide much in terms of metrics.
> 
> What kind of metrics are you interested in? There's a `/progress` callback that Ghostferry exposes (not sure about the control server). The control server is a bit out of date with the rest of the code base and could be updated.

I am working on some instrumentation using Prometheus and I will update this PR soon. I've been working on some metrics for tracking the time taken by each batch to be processed and also the number of rows processed.

In addition to that, I've tested different `fetch` process. One is using `SELECT LIMIT` and the other `OFFSET`. I kind of expected those results but seeing on a chart makes easier to digest and think about.

Below you can find some findings.

**SELECT LIMIT**
image

**SELECT OFFSET**
image

As you can see the time running the offset experiment was pretty short. It degrades quite a lot across time and also stresses the Source DB up to an unsustainable point.  

On the other hand, the select limit approach is quite stable and didn't cause much overhead to the server.

For this experiment, the concurrency was set to a very low number (4).

I am gonna post more details next week.

PR #155, opened 2020-02-27
Add maintenance comment to all sql change queries

kgalieva:
# The Problem
Need to differentiate Technical Ghostferry's binlog events from Core's Business binlog events

# The Solution
Add `/*maintenance:ghostferry*/` comment to all change queries that are executed by Ghostferry.

### DIFF
--- a/config.go
+++ b/config.go
@@ -49,12 +49,15 @@ func (this *TLSConfig) BuildConfig() (*tls.Config, error) {
 }
 
 type DatabaseConfig struct {
-	Host      string
-	Port      uint16
-	User      string
-	Pass      string
-	Collation string
-	Params    map[string]string
+	Host       string
+	Port       uint16
+	User       string
+	Pass       string
+	Collation  string
+	Params     map[string]string
+	// SQL query comments to differentiate Ghostferry's binlog events
+	// Optional: defaults to empty string (no comments)
+	Marginalia string
 
 	TLS *TLSConfig
 }
@@ -126,7 +129,7 @@ func (c *DatabaseConfig) SqlDB(logger *logrus.Entry) (*sql.DB, error) {
 		logger.WithField("dsn", MaskedDSN(dbCfg)).Info("connecting to database")
 	}
 
-	return sql.Open("mysql", dbCfg.FormatDSN())
+	return sql.Open("mysql", dbCfg.FormatDSN(), c.Marginalia)
 }
 
 func (c *DatabaseConfig) assertParamSet(param, value string) error {
--- a/sqlwrapper/ghostferry_db.go
+++ b/sqlwrapper/ghostferry_db.go
@@ -7,23 +7,25 @@ import (
 
 type DB struct {
 	*sqlorig.DB
+	marginalia string
 }
 
 type Tx struct {
 	*sqlorig.Tx
+	marginalia string
 }
 
-func Open(driverName, dataSourceName string) (*DB, error) {
+func Open(driverName, dataSourceName, marginalia string) (*DB, error) {
 	sqlDB, err := sqlorig.Open(driverName, dataSourceName)
-	return &DB{sqlDB}, err
+	return &DB{sqlDB, marginalia}, err
 }
 
 func (db DB) PrepareContext(ctx context.Context, query string) (*sqlorig.Stmt, error) {
-	return db.DB.PrepareContext(ctx, query)
+	return db.DB.PrepareContext(ctx, Annotate(query, db.marginalia))
 }
 
 func (db DB) ExecContext(ctx context.Context, query string, args ...interface{}) (sqlorig.Result, error) {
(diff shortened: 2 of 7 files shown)


shuhaowu:
The approach here of a static comment is ok to identify Ghostferry, but we just talked with @fjordan that we need more granular tagging. Thus, we should make this comment configurable via Ghostferry config struct. The caller of Ghostferry can then customize the comment, such as putting in a job id, tenant id, and possibly more.

cc: @fjordan

kgalieva:
I've pushed a change to make configs configurable. Touched a lot of files to pass it through `sql.Open`. Not sure if there is a simpler way to do it.

kgalieva:
@fjordan @shuhaowu could you please review?

config.go
```diff
@@ -116,7 +116,7 @@ func (c *DatabaseConfig) Validate() error {
 	return nil
 }
 
-func (c *DatabaseConfig) SqlDB(logger *logrus.Entry) (*sql.DB, error) {
+func (c *DatabaseConfig) SqlDB(logger *logrus.Entry, comment string) (*sql.DB, error) {
```
shuhaowu: Can this comment not be a struct member on `DatabaseConfig`? It feels like a configuration to me. If there's a concern of needing to specify this for the source master, source, and target databases, we can have `Config.QueryComment` be automatically propagated to each `DatabaseConfig` near the code where we set defaults?

config.go
```diff
@@ -479,6 +479,10 @@ type Config struct {
 	// 2. Use the table's primary key column as the pagination column. Fail if the primary key is not numeric or is a composite key without a FallbackColumn specified.
 	// 3. Use the FallbackColumn pagination column, if configured. Fail if we cannot find this column in the table.
 	CascadingPaginationColumnConfig *CascadingPaginationColumnConfig
+
+	// SQL query comments to differentiate Ghostferry's binlog events
+	// Optional: defaults to empty string (no comments)
+	QueryComment string
```
fjordan: By moving this piece of configuration to within the `DatabaseConfig`, we can remove the need to pass it around to the various constructors that will use it and just reference it internally in the `SqlDB` method and in `ghostferry_db.go`.

I would also recommend renaming this to `Marginalia` as it is consistent with our other naming internally. If we think it's a problem that this is confusing because it's not using ActiveRecord (https://github.com/shopify/marginalia), then we can use a name like `Annotation`.

sqlwrapper/ghostferry_db.go
```diff
@@ -86,3 +88,7 @@ func (tx Tx) QueryRowContext(ctx context.Context, query string, args ...interfac
 func (tx Tx) QueryRow(query string, args ...interface{}) *sqlorig.Row {
 	return tx.Tx.QueryRow(query, args...)
 }
+
+func Comment(query, comment string) string {
```
fjordan: I think `annotate` is a better name here. It doesn't need to be exported so it can be lowercased.

test/go/race_conditions_integration_test.go
```diff
@@ -16,6 +16,14 @@ func setupSingleEntryTable(f *testhelpers.TestFerry, sourceDB, targetDB *sql.DB)
 	testhelpers.SeedInitialData(targetDB, "gftest", "table1", 0)
 }
 
+func addMaintenanceComments(queries []string, comment string) []string {
```
fjordan: Should this be named `addMaintenanceComments`? I think `addAnnotations` or `annotate` may be a better name.

kgalieva:
I updated the code. @fjordan @shuhaowu could you please review the changes?

sqlwrapper/ghostferry_db.go
```diff
@@ -86,3 +88,7 @@ func (tx Tx) QueryRowContext(ctx context.Context, query string, args ...interfac
 func (tx Tx) QueryRow(query string, args ...interface{}) *sqlorig.Row {
 	return tx.Tx.QueryRow(query, args...)
 }
+
+func Annotate(query, marginalia string) string {
```
fjordan: does this need to be exported? I think it can be only used internally if we don't need to call it from the test below

test/go/race_conditions_integration_test.go
```diff
@@ -37,6 +45,8 @@ func TestSelectUpdateBinlogCopy(t *testing.T) {
 			}(queries[i])
 		}
 
+		queries = annotate(queries, testcase.Ferry.Source.Marginalia)
```
fjordan: what's the purpose of annotating these queries? Are the annotations being asserted?

kgalieva: The queries that are executed contain comments. This is why we need to wait for the annotated queries in this test.

sqlwrapper/ghostferry_db.go
kgalieva: it's used in race_conditions_integration_test.go to annotated test queries

Issue #157, opened 2020-03-08
Data corruption in binary and varbinary columns 


kolbitsch-lastline:
When retrieving updates via binlog , the replication module is stripping trailing 0-bytes from the data in data updates. This affects the data inserted via insert/update statements as well as the "where" clause for update/delete statements.
As a result, a target DB starts to diverge in the data that it stores.

IMO this is a bug in the upstream library we use, and I have opened a ticket with them:

https://github.com/siddontang/go-mysql/issues/477

However, existing users of the library may break if the library is changed, which is why I think we may need to work around the bug in ghostferry directly.

To do this, we need to update go-mysql to parse BINARY/VAR_BINARY table column definition (it currently assumes they are simple string columns) and extend input array before using the data in calls like _BinlogUpdateEvent. AsSQLString_

shuhaowu:
Can you post an example dataset and query so we can add an integration test?

shuhaowu:
Does the InlineVerifier catch this corruption? It should, right?

kolbitsch-lastline:
> Can you post an example dataset and query so we can add an integration test?

should be reproducable with literally any update of a BINARY column type that does uses the full length of the column but happens to have 0-bytes at the end. I can work on writing an integration test in the project - I have a patch that should fix it and can commit it together with the change

Unfortunately it looks like the replication library cannot know of the truncation (unless I'm misreading the docs, the mysql replication protocol doesn't transmit the max-size of a column update), so we'll need to extend the schema extraction and handle the padding in ghostferry (but it seems really simple).

> Does the InlineVerifier catch this corruption? It should, right?

I am not overly familiar with this part yet, but from what I have learned so far, I would assume that it does

kolbitsch-lastline:
ok, sent you 2 pull requests - the first one is to illustrate the problem, the second one contains a potential fix - but the second one should probably not be merged as-is, as I'm actually changing the vendor'ed module to show what I'd propose changing

let's discuss the details there

kolbitsch-lastline:
FYI: tracking the upstream vendor module change here:

https://github.com/siddontang/go-mysql/issues/480

let's see what's the feedback we get there

**NOTE:** To increase chances to get it into upstream, I've made the change a bit more generic (added not only the `FixedSize` but also the `MaxSize` property). If that gets merged in the current PR version, we'll need to trivially update the _ghostferry_ code when pulling in latest master (but we'll have to change other things to work with upstream _master_ anyways)

PR #158, opened 2020-03-10
Unit-test for binary-column data corruption (#157)

kolbitsch-lastline:
There seems to be a bug in the propagation of data from the binlog
streamer to the binlog writer when using fixed-length BINARY columns.

This commit adds an integration test that currently works only under
very specific conditions, and prepares the code for the fix to the data
propagation logic.

### DIFF
--- a/test/go/trivial_integration_test.go
+++ b/test/go/trivial_integration_test.go
@@ -6,6 +6,7 @@ import (
 	sql "github.com/Shopify/ghostferry/sqlwrapper"
 	"math/rand"
 	"testing"
+	"time"
 
 	"github.com/Shopify/ghostferry/testhelpers"
 )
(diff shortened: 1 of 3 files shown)


test/go/trivial_integration_test.go
```diff
@@ -167,6 +140,91 @@ func TestCopyDataWithNullInColumn(t *testing.T) {
 	testcase.Run()
 }
 
+func TestCopyDataInFixedSizeBinaryColumn(t *testing.T) {
```
shuhaowu: The golang integration tests are deprecated. Can you write the tests in the ruby integration test instead? Specifically it can go under https://github.com/Shopify/ghostferry/blob/master/test/integration/types_test.rb.

shuhaowu:
I assume this is superceded by #159. Reopen if necessary.

Issue #156, opened 2020-03-06
Resuming ghostferry-copydb after interrupt fails due to missed TableMapEvent


kolbitsch-lastline:
The interrupt-resume feature as described in

https://shopify.github.io/ghostferry/master/copydbinterruptresume.html

works well, if the interrupt does not happen within batched RowsEvent events being processed.

A mysql replication event for changing data is always started by sending a TableMapEvent (describing the table to be changed), followed by one or more RowsEvent (containing the data to be changed). If **multiple consecutive** RowsEvent events are sent **for the same table**, the TableMapEvent is typically skipped (after sending it once).

Thus, if the interrupt happens after receiving the TableMapEvent but before receiving the last RowsEvent, ghostferry will try to resume from the last processed RowsEvent, causing the _replication/BinlogSyncer_ syncer to crash with an **invalid table id < table-ID >, no corresponding table map event** exception.

This is due to the following code:

```
var ok bool
e.Table, ok = e.tables[e.TableID]
if !ok {
	if len(e.tables) > 0 {
		return errors.Errorf("invalid table id %d, no corresponding table map event", e.TableID)
	} else {
		return errMissingTableMapEvent(errors.Errorf("invalid table id %d, no corresponding table map event", e.TableID))
	}
}
```

Note that this is not a bug in the _replication_ module, because ghostferry points the syncer at a resume position after the TableMapEvent , and the code **cannot** satisfy this (without ignoring the fact that it doesn't know the table ID).

Changing the replication to ignore unknown table IDs is tempting but very tricky: we could cache the table IDs previously seen in ghostferry and "fill the gap" this way. Unfortunately the parsing of the RowsEvent data relies on the table schema. And, conceptually, it also seems quite unclean.

Thus, IMO the correct approach is to make sure the syncer is not pointed at a location from which it cannot resume.

Ghostferry today only stores the resume position via its _lastStreamedBinlogPosition_ property. In my variant of ghostferry (see https://github.com/Shopify/ghostferry/issues/26) , I have changed the _BinlogStreamer_ class to keep track of an additional position: the _lastResumeBinlogPosition_ . The idea is to analyze each received event from the syncer and check if it's a RowsEvent or not. If not, the event position is considered "safe to be resumed from" and the property is updated - otherwise it is kept at its current value.
Thus, the resume position is a tuple of "last received/processed event" and "last position we can resume from".

[...]

PR #160, opened 2020-03-12
Fix resuming binlog streaming #156

kolbitsch-lastline:
A mysql replication event for changing data is always started by
sending a TableMapEvent (describing the table to be changed),
followed by one or more RowsEvent (containing the data to be
changed). If multiple consecutive RowsEvent events are sent for the
same table, the TableMapEvent is typically skipped (after sending it
once), meaning that resuming from such a binlog is not possible.

Resuming at a binlog position between the TableMapEvent and a
RowsEvent is not possible, as the binlog streamer would miss the
table definition for following DML statements.

This commit tracks the position of the most recently processed
TableMapEvent and uses this for resuming (skipping any actual events
between the last write and resume position that were already
processed).


fjordan:
👋 Apologies for the delay in reviews here. 

Thank you for your contribution - this does indeed solve the problem you're describing. 

After running into this problem while doing some binlog streaming against the target for a separate and new feature we're looking to soon add, I took a slightly different approach that led to the same basic resolution, but with fewer changes (primarily by moving the last resumable position to the `DMLEvent` instead of creating a new `BinlogEvent` struct). You can see thee proposed alternative here: https://github.com/Shopify/ghostferry/pull/168

I also borrowed your test and slightly modified the assertions a bit, but I wanted to make sure you knew I was intentionally reusing your work 🙏 😄

kolbitsch-lastline:
glad to hear it was useful and thanks for the pointer

Issue #163, opened 2020-03-15
copydb: State export on stdout makes resuming difficult, if tool invocation fails


kolbitsch-lastline:
Exporting state on stdout works really well, but it has an annoying gotcha: if the tool invocation fails (e.g., a configuration error, DB connectivity error, or something similar) the output on stdout will not contain a valid resume state, as we never entered the main loop of _ghostferry-copydb_. As a result, the caller must be aware of any startup failure and has to take special precaution to not overwrite the statefile generated by a previous tool invocation.

For example: if we run _ghostferry-copydb_ in a loop, where the previous iteration's stdout is written to disk and passed as CLI argument to the next iteration, any iteration where tool invocation fails will inevitably overwrite (and thus lose) the state.

One way to handle this situation would be to configure _ghostferry_ to dump state to a file, and to only write the updated state once we are in a condition where this makes sense - we have successfully started, loaded the original state (if any), are able to export the state, etc. At the same time, we could import the state on _copydb_ start (if that file already exists).

Note that this could also be handled outside the _ghostferry_ library or _ghostferry-copydb_ by putting more intelligence into a caller. But, by embedding more careful handling in the library/tool itself, we can provide users with safe default handling and avoid different solutions to a shared problem.
Clearly whatever change we make in the library/tool must allow for backwards compatibility to maintain the original behavior if that is expected by a caller

kolbitsch-lastline:
FYI: I have a **proof-of-concept patch** that I'd like to open up for review. I understand that it may need a few unit-tests, but before I spend time writing those, I'd like to get some general feedback on the idea in general.

shuhaowu:
We designed Ghostferry to work with container technologies such as Docker. Dumping to a file is somewhat problematic as the container. This is why we dump to stdout for now. I don't think the *core library* should make the decision to dump to a file given this. That said, if you want to make `copydb` dump its state to a file, that's acceptable.

What you're saying is that some of the error messages are dumped to stdout? We can also make these go into stderr like all log messages to be consistent.

kolbitsch-lastline:
> We designed Ghostferry to work with container technologies such as Docker. Dumping to a file is somewhat problematic as the container. This is why we dump to stdout for now. I don't think the core library should make the decision to dump to a file given this.

yeah, same here, only that we run everything via Kubernetes. This makes capturing data on stdout (and pass to a second invocation) even more difficult.

[...]

PR #164, opened 2020-03-15
Allow state tracking via fixed file

kolbitsch-lastline:
The previous handling of state across runs of ghostferry was only
allowing the use of stdout. In cases where the tools fails to start,
this mean that the output in stdout was not a valid state, and the
caller has to take special precaution to not override the previous
state.

This commit extends the state handling to allow passing a state file in
the configuration. This state is loaded (if not overridden with the
state file provided as command-line argument), and the state is written
to this file only when the tool shuts down in a way that the output will
be a valid state.

### DIFF
--- a/config.go
+++ b/config.go
@@ -401,6 +401,13 @@ type Config struct {
 	// by dumping the current state to stdout and the error HTTP callback.
 	// The dumped state can be used to resume Ghostferry.
 	DumpStateOnSignal bool
+	// When dumping state is enabled, the file to which to write the state. If
+	// this is not set, use stdout.
+	//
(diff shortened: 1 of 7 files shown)


kolbitsch-lastline:
argh, had a typo in my commit message and duplicated https://github.com/Shopify/ghostferry/issues/163 . closing this again.

shuhaowu:
If you have a typo in your commit message you can always use `git commit --amend`.

kolbitsch-lastline:
> If you have a typo in your commit message you can always use git commit --amend.

I know - that's what I did... but I realized only after pushing, and it seems that GH automatically creates an issue if there is a `#NNN` in the comment :-)

Issue #161, opened 2020-03-14
Support tables with foreign key constraints


kolbitsch-lastline:
Ghostferry currently documents that it cannot work with foreign-key constraints (_FKCs_).

Intuitively this makes sense for two reasons:
- data is inserted in per-tables batches, and it's not guaranteed that inserts on one table are not inserted before other tables that rows with _FKCs_ may refer to, and
- binlog writing happens in parallel with batch processing, so we may be inserting "new" data that refers to rows that the batch-writer has not yet processed.

However: mysql allows disabling foreign-key constraint checks on a per-session basis, and it does not re-validate constraints when this is disabled. As a result, we may temporarily disable constraint enforcement until the database is back in a consistent state. The only issue that does arise is that tables must be created in an order that satisfies their inter-dependencies.

The golang sql mysql driver even allows disabling constraints on a DB connection using a simple configuration change, making support even easier.

I have a working version of the above table creation change and was curious if you guys think it's a useful addition to the **ghostferry-copydb** tool. I understand it's somewhat hack-ish, but it could be useful in many scenarios where ghostferry cannot be used today.

Also: am I overlooking something with my assumption that disabling FKCs during the copy process is a problem?

shuhaowu:
See https://github.com/github/gh-ost/issues/507.

We have no plans to add this as we don't use FK in our tables. We would like to avoid changes to the core library to support FK. I'm also not very familiar with what the implications of FK is in terms of data consistency and do not have the spare time to review its implications. We do not want to have a ghostferry "mainline" application on master that performs something we don't know for sure to be safe.

That said, you could fork copydb into a separate application that handles FK. We could put it into a "staging" directory in the main ghostferry repo so people can try it at their own risk.

kolbitsch-lastline:
> We have no plans to add this as we don't use FK in our tables. We would like to avoid changes to the core library to support FK. I'm also not very familiar with what the implications of FK is in terms of data consistency and do not have the spare time to review its implications. We do not want to have a ghostferry "mainline" application on master that performs something we don't know for sure to be safe.

that makes sense, but please note that the change is entirely in the database/table creation by introducing an order. Thus, no changes in the main pipeline are really needed and any breakage would have very limited impact (would be detected right away)

> That said, you could fork copydb into a separate application that handles FK. We could put it into a "staging" directory in the main ghostferry repo so people can try it at their own risk.

[...]

PR #166, opened 2020-03-17
Allow specifying table creation order #161

kolbitsch-lastline:
This commit introduces the ability for ghostferry-copydb to create
tables in a specific order. This can be useful if tables to be created
contain foreign-key constraints (FKCs).

NOTE: This does not mean that ghostferry supports FKCs! However, with
this feature and by disabling FKCs in the target DB, it is
theoretically possible to migrate DBs with FKCs - this is experimental
and not recommended for production usage. Use with care!

To disable the FKCs, one must add the following config to the target DB
configuration:

    "Params": {
        "foreign_key_checks": "0"
    }

### DIFF
--- a/copydb/config.go
+++ b/copydb/config.go
@@ -44,6 +44,14 @@ type Config struct {
 	// Filter configuration for tables to copy
 	Tables FilterAndRewriteConfigs
 
+	// Specifies the order in which to create database tables as <db>.<table> .
+	// Names refer to original databases and tables (that is, before renaming
+	// occurs).
+	// If a table is to be created on start and appears in this list, it is
+	// created before any other table, and is created in the order listed here.
+	// All tables not specified in this list are created in arbitrary order.
+	TablesToBeCreatedFirst []string
+
 	// If you're running Ghostferry from a read only replica, turn this option
 	// on and specify SourceReplicationMaster and ReplicatedMasterPositionQuery.
 	RunFerryFromReplica bool
--- a/copydb/copydb.go
+++ b/copydb/copydb.go
@@ -60,7 +60,7 @@ func (this *CopydbFerry) CreateDatabasesAndTables() error {
 	// We need to create the same table/schemas on the target database
 	// as the ones we are copying.
 	logrus.Info("creating databases and tables on target")
-	for tableName := range this.Ferry.Tables {
+	for _, tableName := range this.Ferry.Tables.GetTableListWithPriority(this.config.TablesToBeCreatedFirst) {
 		t := strings.Split(tableName, ".")
 
 		err := this.createDatabaseIfExistsOnTarget(t[0])
--- a/table_schema_cache.go
+++ b/table_schema_cache.go
@@ -293,6 +293,31 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
 	return c[fullTableName(database, table)]
 }
 
+// Helper to sort a given map of tables with a second list giving a priority.
+// If an element is present in the input and the priority lists, the item will
+// appear first (in the order of the priority list), all other items appear in
+// the order given in the input
+func (c TableSchemaCache) GetTableListWithPriority(priorityList []string) (prioritzedTableNames []string) {
+	// just a fast lookup if the list contains items already
+	contains := map[string]struct{}{}
+	if len(priorityList) >= 0 {
+		for _, tableName := range priorityList {
+			// ignore tables given in the priority list that we don't know
+			if _, found := c[tableName]; found {
+				contains[tableName] = struct{}{}
+				prioritzedTableNames = append(prioritzedTableNames, tableName)
+			}
+		}
+	}
+	for tableName, _ := range c {
+		if _, found := contains[tableName]; !found {
+			prioritzedTableNames = append(prioritzedTableNames, tableName)
+		}
+	}
+
+	return
+}
+
 func showDatabases(c *sql.DB) ([]string, error) {
 	rows, err := c.Query("show databases")
 	if err != nil {
--- a/table_schema_cache.go
+++ b/table_schema_cache.go
@@ -297,7 +297,7 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
 // If an element is present in the input and the priority lists, the item will
 // appear first (in the order of the priority list), all other items appear in
 // the order given in the input
-func (c TableSchemaCache) GetTableListWithPriority(priorityList []string) (prioritzedTableNames []string) {
+func (c TableSchemaCache) GetTableListWithPriority(priorityList []string) (prioritizedTableNames []string) {
 	// just a fast lookup if the list contains items already
 	contains := map[string]struct{}{}
 	if len(priorityList) >= 0 {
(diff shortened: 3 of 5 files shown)


copydb/config.go
```diff
@@ -44,6 +44,15 @@ type Config struct {
...
 	Tables FilterAndRewriteConfigs
 
+	// Hints in which order to create database tables, as <db>.<table> . If a
+	// table is to be created on start and appears in this list, it is created
+	// before any other table, and is created in the order listed here
+	//
+	// This is useful when tables have dependencies amongst each other. By
+	// default, create them in an arbitrary order. Names refer to original
+	// databases and tables - that is, before renaming occurs.
+	TableCreationOrder []string
```
shuhaowu: `TablesToBeCreatedFirstInOrder`?

copydb/config.go
```diff
@@ -44,6 +44,15 @@ type Config struct {
 	// Filter configuration for tables to copy
 	Tables FilterAndRewriteConfigs
 
+	// Hints in which order to create database tables, as <db>.<table> . If a
+	// table is to be created on start and appears in this list, it is created
+	// before any other table, and is created in the order listed here
```
shuhaowu: + "All tables not specified in this list are created in arbitrary order."

copydb/copydb.go
```diff
@@ -60,7 +60,8 @@ func (this *CopydbFerry) CreateDatabasesAndTables() error {
 	// We need to create the same table/schemas on the target database
 	// as the ones we are copying.
 	logrus.Info("creating databases and tables on target")
-	for tableName := range this.Ferry.Tables {
+	tablesToCreate := this.Ferry.Tables.GetTableListWithPriority(this.config.TableCreationOrder)
+	for _, tableName := range tablesToCreate {
```
shuhaowu: `... := range this.Ferry.Tables.OrderedByCreationPriority()`?

table_schema_cache.go
```diff
@@ -293,6 +293,31 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
 	return c[fullTableName(database, table)]
 }
 
+// Helper to sort a given map of tables with a second list giving a priority.
+// If an element is present in the input and the priority lists, the item will
+// appear first (in the order of the priority list), all other items appear in
+// the order given in the input
+func (c TableSchemaCache) GetTableListWithPriority(priorityList []string) (tableNames []string) {
+	// just a fast lookup if the list contains items already
+	contains := map[string]bool{}
```
shuhaowu: You can do `map[sting]struct{}`, as that's the standard way to express a set in Golang.

copydb/config.go
kolbitsch-lastline: ACK. Will fix once I get more feedback to avoid unnecessary patch-sets (or once I get the "all clear otherwise" :-) )

table_schema_cache.go
```diff
@@ -293,6 +293,31 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
...
+	if len(priorityList) >= 0 {
+		for _, tableName := range priorityList {
+			// ignore tables given in the priority list that we don't know
+			if c[tableName] != nil {
```
shuhaowu: `if _, found := c[tableName]; found {`

Otherwise the code might panic?

table_schema_cache.go
```diff
@@ -293,6 +293,31 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
...
+				contains[tableName] = true
```
shuhaowu: `contains[tableName] = struct{}{}`

This stores the key of the table with a value of 0 bytes, as recommended by Golang to express a set.

table_schema_cache.go
```diff
@@ -293,6 +293,31 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
...
+				tableNames = append(tableNames, tableName)
+			}
+		}
+	}
+	for tableName, _ := range c {
+		if contains[tableName] != true {
```
shuhaowu: `_, found := contains[tableName]; !found {`

shuhaowu:
This PR fixes #161  and #3

tufanbarisyildirim:
good to see progress on this, thanks

table_schema_cache.go
kolbitsch-lastline: happy to change, but in what situation will it panic?

from the golang language spec:

https://golang.org/doc/effective_go.html?h=An+attempt+to+fetch+a+map+value+with+a+key+that

> An attempt to fetch a map value with a key that is not present in the map will return the zero value for the type of the entries in the map. For instance, if the map contains integers, looking up a non-existent key will return 0. A set can be implemented as a map with value type bool. Set the map entry to true to put the value in the set, and then test it by simple indexing.

thus, we will receive a `nil` , which is the null-value for the pointer.

Not arguing against your suggestion - it's "the go way" and I'll definitely improve it, but want to make sure I'm not misunderstanding why you're suggesting it, as I don't see a path leading to panic

table_schema_cache.go
```diff
@@ -293,6 +293,31 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
 	return c[fullTableName(database, table)]
 }
 
+// Helper to sort a given map of tables with a second list giving a priority.
+// If an element is present in the input and the priority lists, the item will
+// appear first (in the order of the priority list), all other items appear in
+// the order given in the input
+func (c TableSchemaCache) GetTableListWithPriority(priorityList []string) (tableNames []string) {
```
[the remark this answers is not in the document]
kolbitsch-lastline: > I would remove the param from this method and make it implicitly use the configured TableCreationOrder as it doesn't appear to be necessary.

the method is defined on `TableSchemaCache` which is defined as

```
type TableSchemaCache map[string]*TableSchema
```

as such, it is self-contained and does not have access to the config. I think passing it in is required - unless I'm misunderstanding your suggestion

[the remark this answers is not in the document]
fjordan: Yes, sorry, I got ahead of myself and didn't fully explain. 

This would require moving the `TableCreationOrder` to the base configuration struct as well as the `TableSchema` struct instead of in `copydb/config.go` (similar to the `CompressedColumnsForVerification`). It can then be passed to the `LoadTables` func, although that method's params do seem to be getting a bit unwieldy. 

If you're in on board, it may be easier at this point to add an additional struct to be used simply for the `LoadTables` function - something like a `TableConfiguration` that contains all of the params as struct members.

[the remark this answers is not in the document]
kolbitsch-lastline: I'm not against it, but I'd like to hear other opinions. I personally feel it's more of a change than necessary for this specific change. I'm a strong proponent for minimal patch-sets whenever possible (as you may have seen from the 1734 issues I have opened and sent separate PRs for ;-) .

I **do** agree, however, that the number of parameters passed to `LoadTables` is insane :-) and a dedicated struct seems very reasonable. But I would say that it's orthogonal to the proposed change here

table_schema_cache.go
```diff
@@ -293,6 +293,31 @@ func (c TableSchemaCache) Get(database, table string) *TableSchema {
...
+// the order given in the input
+func (c TableSchemaCache) GetTableListWithPriority(priorityList []string) (prioritzedTableNames []string) {
+	// just a fast lookup if the list contains items already
+	contains := map[string]struct{}{}
+	if len(priorityList) >= 0 {
+		for _, tableName := range priorityList {
+			// ignore tables given in the priority list that we don't know
+			if _, found := c[tableName]; found {
+				contains[tableName] = struct{}{}
+				prioritzedTableNames = append(prioritzedTableNames, tableName)
```
shuhaowu: Typo: priorit`i`zed

kolbitsch-lastline: haha, you have no idea how much it would have bugged me to merge with this typo :laughing:

table_schema_cache.go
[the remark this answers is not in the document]
fjordan: That's fair. I don't feel strongly enough to block the PR, but was simply proposing an alternative that I felt would make more sense with a bit larger change.

kolbitsch-lastline:
> Let's fix the one typo and I'll merge it.

ack, thanks. BTW, I'm used to doing squashes before pushing a PR - but I think this is not the "github way". So I'll try to just do an update commit, I assume you'll merge after squashing, correct?

> Thanks for the contribution!

any time

shuhaowu:
Thanks for the contribution! Sorry it took a bit as we were a bit busy with some other work.

PR #167, opened 2020-04-06
Upgrade to go 1.14.1

fjordan:
Alternative to https://github.com/Shopify/ghostferry/pull/125. 

Upgrades to go **1.14.1** and introduces **go modules**.

The previous version was 1.10, and was still using `dep`. 

With the upgrade, this moves to go modules, but can still make the `vendor` dir if we'd like by using `go build -mod=vendor`.  The `vendor` directory was persisted using `go mod vendor` after running `go mod download` and `go mod tidy`.

/cc @Shopify/pods


shuhaowu:
Did you update the siddontang/go-mysql dependency? It's usually worth looking at a diff and a commit log so we can see if they changed something that may be important to us.

shuhaowu:
This is great work by the way. Thanks a tonne.

config.go
```diff
@@ -64,14 +65,14 @@ type DatabaseConfig struct {
...
-		Params:    c.Params,
-
-		MultiStatements: true,
+		User:                 c.User,
+		Passwd:               c.Pass,
+		Net:                  "tcp",
+		Addr:                 fmt.Sprintf("%s:%d", c.Host, c.Port),
+		Collation:            c.Collation,
+		Params:               c.Params,
+		AllowNativePasswords: true,
```
fjordan, line 74: This is used for the tests

Makefile
```diff
@@ -34,7 +34,7 @@ DEB_TARGET       = $(BUILD_DIR)/$(PROJECT_BIN)_$(VERSION_STR).deb
 
 $(PROJECTS): $(GOBIN)
 	$(eval proj := $@)
-	go build -i -ldflags "$(LDFLAGS)" -o $(BIN_TARGET) $(PROJECT_PKG)
+	go build -mod=vendor -i -ldflags "$(LDFLAGS)" -o $(BIN_TARGET) $(PROJECT_PKG)
```
fjordan, line 37: This is to ensure we're building with the dependencies in the `vendor/` directory

PR #168, opened 2020-04-08
Fix resumable binlog position after interrupt

fjordan:
Alternative to https://github.com/Shopify/ghostferry/pull/160 (and also makes use of the same added test with slight modifications).

As described in the other PR, attempting to resume Ghostferry from an invalid binlog position will result in an error as the preceding `TableMapEvent` is required to properly translate `RowsEvents`. 

This adds a resumable position and interface to the `DMLEvent` that the `InlineVerifier` and `BinlogWriter` persist during interruption to be reused during resume. 

The last resumable position is stored as a member of the `BinlogStreamer` and is periodically updated as the `BinlogStreamer` sees new resumable events (`GTIDEvents` and `XIDEvents`). This position is then sent along with the corresponding `DMLEvent`. The resumable position is made available on the `DMLEvents` using a new interface method - `ResumableBinlogPosition`.

/cc @Shopify/pods

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -5,8 +5,9 @@ import (
 	"crypto/tls"
 	sqlorig "database/sql"
 	"fmt"
-	sql "github.com/Shopify/ghostferry/sqlwrapper"
 	"time"
+
+	sql "github.com/Shopify/ghostferry/sqlwrapper"
 
 	"github.com/siddontang/go-mysql/mysql"
 	"github.com/siddontang/go-mysql/replication"
@@ -24,12 +25,13 @@ type BinlogStreamer struct {
 
 	TableSchema TableSchemaCache
 
-	binlogSyncer               *replication.BinlogSyncer
-	binlogStreamer             *replication.BinlogStreamer
-	lastStreamedBinlogPosition mysql.Position
-	targetBinlogPosition       mysql.Position
-	lastProcessedEventTime     time.Time
-	lastLagMetricEmittedTime   time.Time
+	binlogSyncer                *replication.BinlogSyncer
+	binlogStreamer              *replication.BinlogStreamer
+	lastStreamedBinlogPosition  mysql.Position
+	lastResumableBinlogPosition mysql.Position
+	targetBinlogPosition        mysql.Position
+	lastProcessedEventTime      time.Time
+	lastLagMetricEmittedTime    time.Time
 
 	stopRequested bool
 
@@ -122,7 +124,6 @@ func (s *BinlogStreamer) Run() {
 	}()
 
 	s.logger.Info("starting binlog streamer")
-
 	for !s.stopRequested || (s.stopRequested && s.lastStreamedBinlogPosition.Compare(s.targetBinlogPosition) < 0) {
 		var ev *replication.BinlogEvent
 		var timedOut bool
@@ -153,19 +154,22 @@ func (s *BinlogStreamer) Run() {
 		case *replication.RotateEvent:
 			// This event is needed because we need to update the last successful
 			// binlog position.
+			s.lastResumableBinlogPosition.Pos = uint32(e.Position)
+
 			s.lastStreamedBinlogPosition.Pos = uint32(e.Position)
 			s.lastStreamedBinlogPosition.Name = string(e.NextLogName)
 			s.logger.WithFields(logrus.Fields{
 				"pos":  s.lastStreamedBinlogPosition.Pos,
 				"file": s.lastStreamedBinlogPosition.Name,
 			}).Info("rotated binlog file")
+		case *replication.TableMapEvent:
+			continue
 		case *replication.RowsEvent:
 			err = s.handleRowsEvent(ev)
 			if err != nil {
 				s.logger.WithError(err).Error("failed to handle rows event")
 				s.ErrorHandler.Fatal("binlog_streamer", err)
 			}
-
 			s.updateLastStreamedPosAndTime(ev)
 		case *replication.FormatDescriptionEvent:
 			// This event has a LogPos = 0, presumably because this is the first
@@ -184,6 +188,7 @@ func (s *BinlogStreamer) Run() {
 			// so there's no way to handle this for us.
 			continue
 		default:
+			s.lastResumableBinlogPosition.Pos = uint32(ev.Header.LogPos)
 			s.updateLastStreamedPosAndTime(ev)
 		}
 	}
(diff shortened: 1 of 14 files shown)


binlog_streamer.go
```diff
@@ -184,6 +188,7 @@ func (s *BinlogStreamer) Run() {
 			// so there's no way to handle this for us.
 			continue
 		default:
+			s.lastResumableBinlogPosition.Pos = uint32(ev.Header.LogPos)
```
hkdsun, line 191: This is making some assumption about what's the `default:` case 🤔 IIRC, you said we wanna record XIDEvent..

should that be made explicit?

fjordan: Yes. 

Following up on our discussion, we see the `GTIDEvent` (if enabled) at the start of a transaction, and an `XIDEvent` at the end of a transaction. 

I've updated this block to reflect that and have also renamed the state tracker's methods to use `Resumable` in place of `Written` for the `BinlogWriter` and `InlineVerifier`'s resumable positions 👍

binlog_streamer.go
```diff
@@ -261,7 +268,8 @@ func (s *BinlogStreamer) handleRowsEvent(ev *replication.BinlogEvent) error {
 		return nil
 	}
 
-	dmlEvs, err := NewBinlogDMLEvents(table, ev, pos)
+	s.lastResumableBinlogPosition.Name = s.lastStreamedBinlogPosition.Name
```
hkdsun, line 271: Wait, what? I don't think this should be legal 😄 jokes aside, mutating any binlog position sounds like a recipe for bugs down the line.. it's a bit like randomizing the coordinates you put in a GPS.. you might end up in middle of nowhere!

 Would a test fail if I removed this line?

binlog_streamer.go
```diff
@@ -183,6 +187,9 @@ func (s *BinlogStreamer) Run() {
 			// with empty GenericEvent structs.
 			// so there's no way to handle this for us.
 			continue
+		case *replication.XIDEvent, *replication.GTIDEvent:
+			s.lastResumableBinlogPosition.Pos = uint32(ev.Header.LogPos)
```
hkdsun, line 191: Let's document in code comments what got us to this? 🙏

fjordan: can do 👍

binlog_streamer.go
fjordan: I wouldn't recommend removing it 😄. 

Instead of just duplicating all the places where `lastStreamedBinlogPosition` is set to also set this value, we just reuse the value like is already being done for the `pos` value in line 262. 

Do you think it makes more sense to put it closer to the `pos` assignment? I don't think it would make much different

hkdsun: Oh I see what's going on. This could use some cleanup, I'll PR something based off your branch.

binlog_streamer.go
```diff
@@ -188,6 +188,30 @@ func (s *BinlogStreamer) Run() {
...
+			// set to "ON".
+			//
+			// Second, there will be at least one (but potentially more) RowsEvents
+			// depending on the number of rows updated in the transaction.
+			//
+			// Lastly, GTIDEvents will only be available if they are enabled.
+			//
+			// As a result, the following case will set the last resumable position for
+			// interruption to EITHER the start (if using GTIDs) or the end of the
+			// last transaction
```
hkdsun, line 214: Succinct explanation. Thank you!

test/integration/interrupt_resume_test.rb
```diff
@@ -304,4 +303,61 @@ def test_interrupt_resume_inline_verifier_will_verify_additional_rows_changed_on
...
+          insert_sql += ", " if value_in_batch > 0
+          insert_sql += "('data#{row_id}')"
+        end
+        source_db.query(insert_sql)
+      end
+    end
+
+    ghostferry.on_status(Ghostferry::Status::AFTER_BINLOG_APPLY) do
+      # while we are emitting events in the loop above, try to inject a shutdown
+      # signal, hoping to interrupt between applying an INSERT and receiving the
```
shuhaowu, line 335: So this test will be known to racy? Will it make CI more flaky?

testhelpers/unit_test_suite.go
```diff
@@ -17,6 +17,9 @@ func SetupTest() {
 	var err error
 
 	logrus.SetLevel(logrus.DebugLevel)
+	if os.Getenv("CI") == "true" {
+		logrus.SetLevel(logrus.ErrorLevel)
```
shuhaowu, line 21: Can you remind me what this is for? I don't recall anymore.

binlog_streamer.go
```diff
@@ -153,19 +154,22 @@ func (s *BinlogStreamer) Run() {
 		case *replication.RotateEvent:
 			// This event is needed because we need to update the last successful
 			// binlog position.
+			s.lastResumableBinlogPosition.Pos = uint32(e.Position)
```
shuhaowu, line 157: Shouldn't we need to set the `.Name` here as well? This is a RotateEvent so the file should change, right? That would get rid of the slightly cryptic `s.lastResumableBinlogPosition.Name` line below?

binlog_streamer.go
fjordan: Ended up moving this assignment to the `RotateEvent` and in the initial connect further up. That allows us to remove it from this line here.

test/integration/interrupt_resume_test.rb
fjordan: It has been consistently passing in my experience and hasn’t been flaky

testhelpers/unit_test_suite.go
fjordan: The log output from the binlog events is seemingly causing CI to truncate the output. This just lowers the logging level unless needed for the tests

binlog_streamer.go
fjordan: Following up on our conversation, we decided to set the name here and in the `ConnectBinlogStreamerToMysqlFrom` above 👍

binlog_streamer.go
```diff
@@ -153,6 +155,9 @@ func (s *BinlogStreamer) Run() {
...
+			s.lastResumableBinlogPosition.Name = string(e.NextLogName)
```
hkdsun, line 159: Wait, I thought we agreed we can't resume from a RotateEvent's position?

shuhaowu: This is the only place where MySQL replication gives us the file name of the binlog. This value is therefore cached until the next Rotate event, which would give us the next file name. So we have to set this filename here?

hkdsun: well, why can't we use a separate `currentFilename` variable to cache the current filename? 

the current filename needs a clear separation from the resumable position.

PR #170, opened 2020-04-14
Clean up binlog streamer state tracking and reduce binlog file size to test binlog rotations

hkdsun:
Lot of small changes here:

* Reduce `max_binlog_size` in tests. This will exercise the binlog rotation codepath a lot more.
* Introduce a `currentBinlogFilename` variable and remove the implicit assumption that `lastStreamedPosition.Name` is the _current_ filename as well. This is a source of a lot of confusion.
* Consider fake events (some `RotateEvent`s and some `FormatDescriptionEvent`s) positions as invalid and don't advance our binlog position based on them
* Remove all mutations of all `mysql.Position` variables. Recreate the values from fresh state instead.
* Make the event handler `switch` statement completely imperative. Make it extra clear when `s.updateLastStreamedPosAndTime` is called.
* Fix a bug where we don't update the binlog position on `GenericEvent`s. This was causing tests to hang forever when the binlog is rotated and the target binlog position is in the _next_ file.

cc @Shopify/pods

### DIFF
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -2,12 +2,21 @@ package ghostferry
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	sql "github.com/Shopify/ghostferry/sqlwrapper"
 
 	"github.com/sirupsen/logrus"
 )
+
+var batchWriteRetryDelay = 5 * time.Millisecond
+
+func init() {
+	if os.Getenv("CI") == "true" {
+		batchWriteRetryDelay = 500 * time.Millisecond
+	}
+}
 
 type BatchWriterVerificationFailed struct {
 	mismatchedPaginationKeys []uint64
@@ -38,7 +47,7 @@ func (w *BatchWriter) Initialize() {
 }
 
 func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
-	return WithRetries(w.WriteRetries, 500*time.Millisecond, w.logger, "write batch to target", func() error {
+	return WithRetries(w.WriteRetries, batchWriteRetryDelay, w.logger, "write batch to target", func() error {
 		values := batch.Values()
 		if len(values) == 0 {
 			return nil
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -2,6 +2,8 @@ package ghostferry
 
 import (
 	"fmt"
+	"time"
+
 	sql "github.com/Shopify/ghostferry/sqlwrapper"
 
 	"github.com/sirupsen/logrus"
@@ -36,7 +38,7 @@ func (w *BatchWriter) Initialize() {
 }
 
 func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
-	return WithRetries(w.WriteRetries, 0, w.logger, "write batch to target", func() error {
+	return WithRetries(w.WriteRetries, 500*time.Millisecond, w.logger, "write batch to target", func() error {
 		values := batch.Values()
 		if len(values) == 0 {
 			return nil
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -124,7 +124,10 @@ func (s *BinlogStreamer) Run() {
 	s.ensureLogger()
 
 	defer func() {
-		s.logger.Info("exiting binlog streamer")
+		s.logger.WithFields(logrus.Fields{
+			"stopAtBinlogPosition":       s.stopAtBinlogPosition,
+			"lastStreamedBinlogPosition": s.lastStreamedBinlogPosition,
+		}).Info("exiting binlog streamer")
 		s.binlogSyncer.Close()
 	}()
 
@@ -155,10 +158,15 @@ func (s *BinlogStreamer) Run() {
 			continue
 		}
 
+		evPosition := mysql.Position{
+			Name: s.currentBinlogFilename,
+			Pos:  ev.Header.LogPos,
+		}
+
 		s.logger.WithFields(logrus.Fields{
-			"position":                   ev.Header.LogPos,
-			"file":                       s.currentBinlogFilename,
-			"type":                       fmt.Sprintf("%T", ev.Event),
+			"position": evPosition.Pos,
+			"file":     evPosition.Name,
+			"type":     fmt.Sprintf("%T", ev.Event),
 			"lastStreamedBinlogPosition": s.lastStreamedBinlogPosition,
 		}).Debug("reached position")
 
@@ -234,7 +242,7 @@ func (s *BinlogStreamer) Run() {
 		}
 
 		if isEventPositionValid {
-			s.updateLastStreamedPosAndTime(ev, isEventPositionResumable)
+			s.updateLastStreamedPosAndTime(ev, evPosition, isEventPositionResumable)
 		}
 	}
 }
@@ -272,16 +280,11 @@ func (s *BinlogStreamer) FlushAndStop() {
 	s.stopRequested = true
 }
 
-func (s *BinlogStreamer) updateLastStreamedPosAndTime(ev *replication.BinlogEvent, isResumablePosition bool) {
+func (s *BinlogStreamer) updateLastStreamedPosAndTime(ev *replication.BinlogEvent, pos mysql.Position, isResumablePosition bool) {
 	if (ev.Header.LogPos == 0) || ev.Header.Timestamp == 0 {
 		// This shouldn't happen, as the cases where it does happen are excluded and thus signal
 		// a programming error
 		s.logger.Panicf("logpos: %d %d %T", ev.Header.LogPos, ev.Header.Timestamp, ev.Event)
-	}
-
-	pos := mysql.Position{
-		Name: s.currentBinlogFilename,
-		Pos:  ev.Header.LogPos,
 	}
 
 	s.lastStreamedBinlogPosition = pos
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -194,9 +194,11 @@ func (s *BinlogStreamer) Run() {
 			}
 
 			s.logger.WithFields(logrus.Fields{
-				"pos":      evPosition.Pos,
-				"filename": evPosition.Name,
-			}).Info("binlog file rotated, advancing lastStreamedBinlogPosition")
+				"new_position":  evPosition.Pos,
+				"new_filename":  evPosition.Name,
+				"last_position": s.lastStreamedBinlogPosition.Pos,
+				"last_filename": s.lastStreamedBinlogPosition.Name,
+			}).Info("binlog file rotated")
 		case *replication.FormatDescriptionEvent:
 			// This event is sent:
 			//   1) when our replication client connects to mysql
@@ -243,7 +245,9 @@ func (s *BinlogStreamer) Run() {
 		}
 
 		if isEventPositionValid {
-			s.updateLastStreamedPosAndTime(ev, evPosition, isEventPositionResumable)
+			evType := fmt.Sprintf("%T", ev.Event)
+			evTimestamp := ev.Header.Timestamp
+			s.updateLastStreamedPosAndTime(evTimestamp, evPosition, evType, isEventPositionResumable)
 		}
 	}
 }
@@ -276,30 +280,30 @@ func (s *BinlogStreamer) FlushAndStop() {
 	if err != nil {
 		s.ErrorHandler.Fatal("binlog_streamer", err)
 	}
-	s.logger.WithField("target_position", s.stopAtBinlogPosition).Info("current stop binlog position was recorded")
+	s.logger.WithField("stop_at_position", s.stopAtBinlogPosition).Info("current stop binlog position was recorded")
 
 	s.stopRequested = true
 }
 
-func (s *BinlogStreamer) updateLastStreamedPosAndTime(ev *replication.BinlogEvent, pos mysql.Position, isResumablePosition bool) {
-	if pos.Pos == 0 {
+func (s *BinlogStreamer) updateLastStreamedPosAndTime(evTimestamp uint32, evPos mysql.Position, evType string, isResumablePosition bool) {
+	if evPos.Pos == 0 {
 		// This shouldn't happen, as the cases where it does happen are excluded and thus signal a programming error
-		s.logger.Panicf("tried to advance to a zero log position: %s %d %T", pos.Name, pos.Pos, ev.Event)
+		s.logger.Panicf("tried to advance to a zero log position: %s %d %T", evPos.Name, evPos.Pos, evType)
 	}
 
-	s.lastStreamedBinlogPosition = pos
+	s.lastStreamedBinlogPosition = evPos
 	if isResumablePosition {
-		s.lastResumableBinlogPosition = pos
+		s.lastResumableBinlogPosition = evPos
 	}
 
 	// The first couple of events when connecting the binlog syncer (RotateEvent
 	// and FormatDescriptionEvent) have a zero timestamp..  Ignore those for
 	// timing updates
-	if ev.Header.Timestamp == 0 {
+	if evTimestamp == 0 {
 		return
 	}
 
-	eventTime := time.Unix(int64(ev.Header.Timestamp), 0)
+	eventTime := time.Unix(int64(evTimestamp), 0)
 	s.lastProcessedEventTime = eventTime
 
 	if time.Since(s.lastLagMetricEmittedTime) >= time.Second {
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -178,23 +178,25 @@ func (s *BinlogStreamer) Run() {
 		switch e := ev.Event.(type) {
 		case *replication.RotateEvent:
 			// This event is used to keep the "current binlog filename" of the binlog streamer in sync.
+			nextFilename = string(e.NextLogName)
 
 			isFakeRotateEvent := ev.Header.LogPos == 0 && ev.Header.Timestamp == 0
 			if isFakeRotateEvent {
-				// Sometimes the RotateEvent is fake and not a real rotation. we want to ignore the log position for those events
+				// Sometimes the RotateEvent is fake and not a real rotation. we want to ignore the log position in the header for those events
 				// https://github.com/percona/percona-server/blob/3ff016a46ce2cde58d8007ec9834f958da53cbea/sql/rpl_binlog_sender.cc#L278-L287
 				// https://github.com/percona/percona-server/blob/3ff016a46ce2cde58d8007ec9834f958da53cbea/sql/rpl_binlog_sender.cc#L904-L907
-				isEventPositionValid = false
+
+				// However, we can always advance our lastStreamedBinlogPosition according to its data fields
+				evPosition = mysql.Position{
+					Name: string(e.NextLogName),
+					Pos:  uint32(e.Position),
+				}
 			}
 
-			nextFilename = string(e.NextLogName)
-
 			s.logger.WithFields(logrus.Fields{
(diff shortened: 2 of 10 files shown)


binlog_streamer.go
```diff
@@ -212,10 +221,11 @@ func (s *BinlogStreamer) Run() {
 			// As a result, the following case will set the last resumable position for
 			// interruption to EITHER the start (if using GTIDs) or the end of the
 			// last transaction
-			s.lastResumableBinlogPosition.Pos = uint32(ev.Header.LogPos)
-			s.updateLastStreamedPosAndTime(ev)
-		default:
```
fjordan, line 217: since we're removing the `default` case, are we are certain that we won't be missing events that will get us to the `stopAtBinlogPosition`?

binlog_streamer.go
```diff
@@ -241,26 +251,35 @@ func (s *BinlogStreamer) FlushAndStop() {
...
+	pos := mysql.Position{
+		Name: s.currentBinlogFilename,
+		Pos:  ev.Header.LogPos,
+	}
+
+	s.lastStreamedBinlogPosition = pos
+	if isResumablePosition {
+		s.lastResumableBinlogPosition = pos
+	}
+
```
fjordan, line 282: I think the cleanup here is nice 👍 Consolidating all of this in one place definitely makes is more clear.

binlog_streamer.go
```diff
@@ -212,10 +221,11 @@ func (s *BinlogStreamer) Run() {
...
 			// interruption to EITHER the start (if using GTIDs) or the end of the
 			// last transaction
-			s.lastResumableBinlogPosition.Pos = uint32(ev.Header.LogPos)
-			s.updateLastStreamedPosAndTime(ev)
-		default:
-			s.updateLastStreamedPosAndTime(ev)
+			isEventPositionResumable = true
+		}
+
+		if isEventPositionValid {
```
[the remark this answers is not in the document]
hkdsun: Good question.. I'll think about it

binlog_streamer.go
hkdsun: the default case is still the same.. all events are "valid" position by default (through the `isEventPositionValid` variable)

binlog_streamer.go
```diff
@@ -151,42 +155,47 @@ func (s *BinlogStreamer) Run() {
...
 		case *replication.RotateEvent:
-			// This event is needed because we need to update the last successful
-			// binlog position.
-			s.lastResumableBinlogPosition.Pos = uint32(e.Position)
-			s.lastResumableBinlogPosition.Name = string(e.NextLogName)
+			// On rotation, two instances of this event is emitted:
+			//   once as the very last event of the previous file
+			//   once as the very first event of a new file
+			// Their only purpose in Ghostferry is to keep track
+			// of the current fillename. Their position is not recorded
```
shuhaowu, line 174:
```suggestion
			// of the current filename. Their position is not recorded
```

binlog_streamer.go
shuhaowu: I have the same question. It may be possible that `stopAtBinlogPosition` points to one of the events where `isEventPositionValid == false`, thereby causing a hang, likely in test only.

docker-compose.yml
```diff
@@ -1,6 +1,17 @@
 mysql-1:
   image: percona:5.7
-  command: --server-id=1 --log-bin=mysql-bin --binlog-format=ROW --sync-binlog=1 --log-slave-updates=ON --gtid-mode=ON --enforce-gtid-consistency=ON --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --max-connections=1000 --read-only=OFF
+  command: --server-id=1
+    --log-bin=mysql-bin
+    --max-binlog-size=4096
```
shuhaowu, line 5: Would this change cause other problems in the tests? Maybe we can no longer replicate certain conditions if the binlog size is too small? Can we dynamically set this variable?

I don't really have an objection but I'm just speculating.

binlog_streamer.go
```diff
@@ -151,42 +155,47 @@ func (s *BinlogStreamer) Run() {
...
 
+		s.logger.WithFields(logrus.Fields{
+			"pos":                        ev.Header.LogPos,
+			"file":                       s.currentBinlogFilename,
+			"type":                       fmt.Sprintf("%T", ev.Event),
+			"lastStreamedBinlogPosition": s.lastStreamedBinlogPosition,
+		}).Debug("reached position")
+
+		isEventPositionResumable := false
+		isEventPositionValid := true
```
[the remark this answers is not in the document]
hkdsun: I have clarified this with references to the mysql source. See my latest changes.

Sorry for skipping out on that at first.

binlog_streamer.go
hkdsun: There is now only two explicit cases where `isEventPositionValid == false`. They're clearly documented with references to the mysql source. All other events are valid positions.

binlog_streamer.go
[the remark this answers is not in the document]
hkdsun: Should be clear now. If not still clear, let me know.

binlog_streamer.go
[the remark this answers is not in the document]
hkdsun: Repeating myself again lol.. The only "invalid" positions now have a `logPos == 0` (and sometimes additional) constraints.

test/integration/interrupt_resume_test.rb
```diff
@@ -315,8 +315,8 @@ def test_interrupt_resume_between_consecutive_rows_events
     # DML event.
     # Since we are racing between applying rows and sending the shutdown event,
     # we emit a whole bunch of them
-    num_batches = 20
-    num_values_per_batch = 1000
+    num_batches = 10
+    num_values_per_batch = 200
```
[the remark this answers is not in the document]
hkdsun: not with the reduced files size

docker-compose.yml
hkdsun: Well, it shouldn't, otherwise we have a bug, no?

This actually is uncovering a bug that we need to dig into more: https://travis-ci.com/github/Shopify/ghostferry/builds/160296941

hkdsun:
This is ready for another round of review.. Sorry for the plethora of changes all rolled into one PR, but it'd be hard to break it up at this point 😞

hkdsun:
Retract the last message. Not ready for review yet.

@jeremycole helped us make a little bit more sense of the mysql binlog sender code.. it turns out that [the fake event that is sent before sending a log file](https://github.com/mysql/mysql-server/blob/5.7/sql/rpl_binlog_sender.cc#L221) is a more trustworthy source of the current filename and we should use that. 

I'll make that change and re-ping for reviews

binlog_streamer.go
```diff
@@ -151,42 +160,60 @@ func (s *BinlogStreamer) Run() {
...
+			if isFakeRotateEvent {
+				// Sometimes the RotateEvent is fake and not a real rotation. we want to ignore the log position in the header for those events
+				// https://github.com/percona/percona-server/blob/3ff016a46ce2cde58d8007ec9834f958da53cbea/sql/rpl_binlog_sender.cc#L278-L287
+				// https://github.com/percona/percona-server/blob/3ff016a46ce2cde58d8007ec9834f958da53cbea/sql/rpl_binlog_sender.cc#L904-L907
+
+				// However, we can always advance our lastStreamedBinlogPosition according to its data fields
+				evPosition = mysql.Position{
+					Name: string(e.NextLogName),
+					Pos:  uint32(e.Position),
+				}
```
hkdsun, line 193: this essentially puts us back where we were before these changes 😅

binlog_streamer.go
```diff
@@ -241,26 +269,36 @@ func (s *BinlogStreamer) FlushAndStop() {
...
 
 	s.stopRequested = true
 }
 
-func (s *BinlogStreamer) updateLastStreamedPosAndTime(ev *replication.BinlogEvent) {
-	if ev.Header.LogPos == 0 || ev.Header.Timestamp == 0 {
-		// This shouldn't happen, as the cases where it does happen are excluded.
-		// However, I've not seen all the cases yet.
-		s.logger.Panicf("logpos: %d %d %T", ev.Header.LogPos, ev.Header.Timestamp, ev.Event)
+func (s *BinlogStreamer) updateLastStreamedPosAndTime(ev *replication.BinlogEvent, pos mysql.Position, isResumablePosition bool) {
```
fjordan, line 284: It's a bit confusing to see this function is receiving an event (that has a position in the header), but also a separate position along with it. I think we can simplify this by dropping either the `pos` or `ev` (pick one):

```suggestion
func (s *BinlogStreamer) updateLastStreamedPosAndTime(pos mysql.Position, isResumablePosition bool) {
```
and then we can consistently just refer to a the position as opposed to looking at both. Alternatively, just extracting the `Timestamp` from the `replication.BinlogEvent` as an `evTime` or something to make it clearer that's all this method is using.

Another thought is that one param seems to be used for actually updating the last streamed and resumable position, and the other for instrumenting binlog streamer lag? Worth breaking these into separate methods?

test/helpers/data_writer_helper.rb
```diff
@@ -59,9 +59,14 @@ def start(&on_write)
...
-          until @stop_requested do
+          begin
             connection = Mysql2::Client.new(@db_config)
-            write_data(connection, &on_write)
+
+            until @stop_requested do
+              write_data(connection, &on_write)
+            end
+          ensure
+            connection.close
```
fjordan, line 69: good catch 👍

test/helpers/data_writer_helper.rb
```diff
@@ -102,6 +107,7 @@ def insert_data(connection)
       table = @tables.sample
       insert_statement = connection.prepare("INSERT INTO #{table} (id, data) VALUES (?, ?)")
       insert_statement.execute(nil, DbHelper.rand_data)
+      insert_statement.close
```
fjordan, line 110: do we need `ensure` statements here and below as well just in case?

batch_writer.go
```diff
@@ -36,7 +38,7 @@ func (w *BatchWriter) Initialize() {
 }
 
 func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
-	return WithRetries(w.WriteRetries, 0, w.logger, "write batch to target", func() error {
+	return WithRetries(w.WriteRetries, 500*time.Millisecond, w.logger, "write batch to target", func() error {
```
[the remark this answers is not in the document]
shuhaowu: Having 0s retry is kinda problematic in production as well. That said until we have an exponential back off we should have this retry timeout to be small like what @fjordan suggests?

binlog_streamer.go
```diff
@@ -151,42 +160,60 @@ func (s *BinlogStreamer) Run() {
...
+			Name: currentFilename,
+			Pos:  ev.Header.LogPos,
+		}
+
+		s.logger.WithFields(logrus.Fields{
+			"position":                   evPosition.Pos,
+			"file":                       evPosition.Name,
+			"type":                       fmt.Sprintf("%T", ev.Event),
+			"lastStreamedBinlogPosition": s.lastStreamedBinlogPosition,
+		}).Debug("reached position")
```
shuhaowu, line 173: This is going to be very verbose, no?

shuhaowu: Also, the "first" event ghostferry receive will have `file` here as `""`, correct?

binlog_streamer.go
```diff
@@ -151,42 +160,60 @@ func (s *BinlogStreamer) Run() {
...
 
-			s.lastStreamedBinlogPosition.Pos = uint32(e.Position)
-			s.lastStreamedBinlogPosition.Name = string(e.NextLogName)
 			s.logger.WithFields(logrus.Fields{
-				"pos":  s.lastStreamedBinlogPosition.Pos,
-				"file": s.lastStreamedBinlogPosition.Name,
-			}).Info("rotated binlog file")
+				"pos":      evPosition.Pos,
+				"filename": evPosition.Name,
+			}).Info("binlog file rotated, advancing lastStreamedBinlogPosition")
```
shuhaowu, line 199: I don't quite understand this log line, specifically the "advancing lastSTreamedBinlogPosition" part. it's only logged if the rotate event is not fake, so shouldn't we just say rotated binlog file? Also possibly log the nextFilename here?

binlog_streamer.go
```diff
@@ -241,26 +269,36 @@ func (s *BinlogStreamer) FlushAndStop() {
...
-		s.targetBinlogPosition, err = ShowMasterStatusBinlogPosition(s.DB)
+		s.stopAtBinlogPosition, err = ShowMasterStatusBinlogPosition(s.DB)
 		return err
 	})
 
 	if err != nil {
 		s.ErrorHandler.Fatal("binlog_streamer", err)
 	}
-	s.logger.WithField("target_position", s.targetBinlogPosition).Info("current stop binlog position was recorded")
+	s.logger.WithField("target_position", s.stopAtBinlogPosition).Info("current stop binlog position was recorded")
```
shuhaowu, line 279:
```suggestion
	s.logger.WithField("stop_at_position", s.stopAtBinlogPosition).Info("current stop binlog position was recorded")
```

binlog_streamer.go
```diff
@@ -151,42 +160,60 @@ func (s *BinlogStreamer) Run() {
...
+			// This event is used to keep the "current binlog filename" of the binlog streamer in sync.
+			nextFilename = string(e.NextLogName)
+
+			isFakeRotateEvent := ev.Header.LogPos == 0 && ev.Header.Timestamp == 0
+			if isFakeRotateEvent {
+				// Sometimes the RotateEvent is fake and not a real rotation. we want to ignore the log position in the header for those events
+				// https://github.com/percona/percona-server/blob/3ff016a46ce2cde58d8007ec9834f958da53cbea/sql/rpl_binlog_sender.cc#L278-L287
+				// https://github.com/percona/percona-server/blob/3ff016a46ce2cde58d8007ec9834f958da53cbea/sql/rpl_binlog_sender.cc#L904-L907
+
+				// However, we can always advance our lastStreamedBinlogPosition according to its data fields
```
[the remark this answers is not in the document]
hkdsun: >Is it possible for the stopAtBinlogPosition to ever be the "fake" RotateEvent?

No, because AFAICT, similar to our replication, the core MySQL replication client also doesn't advance its "master position" (the "last streamed position") based on these fake events

binlog_streamer.go
hkdsun: Good idea to extract and pass the timestamp.. Not gonna go down the instrumentation refactor for now if that's okay

binlog_streamer.go
hkdsun: It's at `Debug` level. Our default logging level is `Info` AFAICT: https://github.com/Shopify/ghostferry/blob/eab1fec72d0081fabe8663676e702569e4671faa/vendor/github.com/sirupsen/logrus/logger.go#L26-L29

hkdsun: Here's an excerpt of the logs in the beginning:

```
D, [2020-04-16T13:19:11.875657 #84026] DEBUG -- : stderr: {"level":"info","msg":"starting data iterator run","tablesCount":1,"tag":"data_iterator","time":"2020-04-16T13:19:11-04:00"}
D, [2020-04-16T13:19:11.875922 #84026] DEBUG -- : stderr: {"level":"info","msg":"starting binlog streamer","tag":"binlog_streamer","time":"2020-04-16T13:19:11-04:00"}
D, [2020-04-16T13:19:11.875971 #84026] DEBUG -- : stderr: {"level":"info","msg":"starting periodic reverifier","tag":"inline-verifier","time":"2020-04-16T13:19:11-04:00"}
D, [2020-04-16T13:19:11.877298 #84026] DEBUG -- : stderr: {"file":"mysql-bin.004297","lastStreamedBinlogPosition":{"Name":"mysql-bin.004297","Pos":816},"level":"debug","msg":"reached position","position":0,"tag":"binlog_streamer","time":"2020-04-16T13:19:11-04:00","type":"*replication.RotateEvent"}
D, [2020-04-16T13:19:11.877467 #84026] DEBUG -- : stderr: {"filename":"mysql-bin.004297","level":"info","msg":"binlog file rotated, advancing lastStreamedBinlogPosition","pos":816,"tag":"binlog_streamer","time":"2020-04-16T13:19:11-04:00"}
D, [2020-04-16T13:19:11.877627 #84026] DEBUG -- : stderr: {"file":"mysql-bin.004297","lastStreamedBinlogPosition":{"Name":"mysql-bin.004297","Pos":816},"level":"debug","msg":"reached position","position":0,"tag":"binlog_streamer","time":"2020-04-16T13:19:11-04:00","type":"*replication.FormatDescriptionEvent"}
D, [2020-04-16T13:19:11.877772 #84026] DEBUG -- : stderr: {"file":"mysql-bin.004297","lastStreamedBinlogPosition":{"Name":"mysql-bin.004297","Pos":816},"level":"debug","msg":"reached position","position":881,"tag":"binlog_streamer","time":"2020-04-16T13:19:11-04:00","type":"*replication.GTIDEvent"}
D, [2020-04-16T13:19:11.883181 #84026] DEBUG -- : stderr: {"file":"mysql-bin.004297","lastStreamedBinlogPosition":{"Name":"mysql-bin.004297","Pos":881},"level":"debug","msg":"reached position","position":949,"tag":"binlog_streamer","time":"2020-04-16T13:19:11-04:00","type":"*replication.QueryEvent"}
```

the first `file` would be set correctly, as we start the binlog streamer with the master position set in `Init()`

binlog_streamer.go
hkdsun: I've updated this to make things more clear. Note that `evPosition` is the "next"/"new" filename and position!

batch_writer.go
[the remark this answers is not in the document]
hkdsun: This is now taken care of with a different retry delay value in CI vs non-CI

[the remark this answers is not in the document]
hkdsun: Note that having a longer retry made CI lot more stable.. especially for verification errors that  cause more panic

batch_writer.go
```diff
@@ -2,11 +2,22 @@ package ghostferry
...
+	"time"
+
 	sql "github.com/Shopify/ghostferry/sqlwrapper"
 
 	"github.com/sirupsen/logrus"
 )
 
+var batchWriteRetryDelay = 5 * time.Millisecond
+
+func init() {
```
shuhaowu, line 15: Can you only define one `init` per package? We can leave this code here and deal with it later, likely when we make this retry delay configurable.

binlog_streamer.go
```diff
@@ -241,27 +273,37 @@ func (s *BinlogStreamer) FlushAndStop() {
...
 
 	s.stopRequested = true
 }
 
-func (s *BinlogStreamer) updateLastStreamedPosAndTime(ev *replication.BinlogEvent) {
-	if ev.Header.LogPos == 0 || ev.Header.Timestamp == 0 {
-		// This shouldn't happen, as the cases where it does happen are excluded.
-		// However, I've not seen all the cases yet.
-		s.logger.Panicf("logpos: %d %d %T", ev.Header.LogPos, ev.Header.Timestamp, ev.Event)
+func (s *BinlogStreamer) updateLastStreamedPosAndTime(evTimestamp uint32, evPos mysql.Position, evType string, isResumablePosition bool) {
```
fjordan, line 288: 💯

batch_writer.go
hkdsun: >Can you only define one init per package?

You can define more than one `init()` and it's common to have multiple inits per package..

hkdsun:
Thanks you reviewers 😍 I'm very excited to merge this

batch_writer.go
shuhaowu: Cool. TIL.

PR #171, opened 2020-04-14
add default marginalia value

fjordan:
Follow up of https://github.com/Shopify/ghostferry/pull/155.

Sets a default marginalia value when not provided by the configuration.

Previous to this PR, if an annotation/marginalia was not provided, Ghostferry's DMLs would not have any annotation. After this change, if an annotation is not provided, Ghostferry will fallback to using the default annotation (`application:ghostferry`).

This is needed for an upcoming PR that will require use of these annotations to validate DMLs against the Target for the duration of the move (until cutover). 

/cc @Shopify/pods

### DIFF
--- a/config.go
+++ b/config.go
@@ -19,6 +19,8 @@ const (
 	VerifierTypeIterative      = "Iterative"
 	VerifierTypeInline         = "Inline"
 	VerifierTypeNoVerification = "NoVerification"
+
+	DefaultMarginalia = "application:ghostferry"
 )
 
 type TLSConfig struct {
@@ -56,11 +58,12 @@ type DatabaseConfig struct {
 	Pass      string
 	Collation string
 	Params    map[string]string
-	// SQL query comments to differentiate Ghostferry's binlog events
-	// Optional: defaults to empty string (no comments)
-	Marginalia string
+	TLS       *TLSConfig
 
-	TLS *TLSConfig
+	// SQL annotations used to differentiate Ghostferry's DMLs
+	// against other actor's. This will default to the defaultMarginalia
+	// constant above if not set.
+	Marginalia string
 }
 
 func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -19,6 +19,8 @@ const (
 	VerifierTypeIterative      = "Iterative"
 	VerifierTypeInline         = "Inline"
 	VerifierTypeNoVerification = "NoVerification"
+
+	DefaultMarginalia = "application:ghostferry"
```
fjordan, line 23: exported as it is used in the tests below

config.go
```diff
@@ -56,11 +58,12 @@ type DatabaseConfig struct {
 	Pass      string
 	Collation string
 	Params    map[string]string
-	// SQL query comments to differentiate Ghostferry's binlog events
-	// Optional: defaults to empty string (no comments)
-	Marginalia string
+	TLS       *TLSConfig
 
-	TLS *TLSConfig
+	// SQL annotations used to differentiate Ghostferry's DMLs
```
fjordan, line 63: Will be providing a comment with further context when adding the Target verification

sqlwrapper/ghostferry_db.go
```diff
@@ -90,5 +91,5 @@ func (tx Tx) QueryRow(query string, args ...interface{}) *sqlorig.Row {
 }
 
 func Annotate(query, marginalia string) string {
-	return marginalia + query
+	return fmt.Sprintf("/*%s*/ %s", marginalia, query)
```
hkdsun, line 94: This is a breaking change.. right? I'd avoid it I think

fjordan: Changed because we're not asking the user to provide the `/*` and `*/` in the configured annotation. 

It would only break something if the provided annotation happened to also have the above, ie: `/*myannotation*/` as opposed to just `myannotation`.

If we're really concerned I can add a check for a "valid" annotation in the `ValidateConfig` to make sure the user isn't providing a comment char...what do you think?

hkdsun: I'm not concerned about user input tbh.. we just have to remember to upgrade the shop mover side so the current config doesn't result in `/*/* stuff */*/`

hkdsun: it's a valid change.. lets keep it 👍

Issue #173, opened 2020-04-16
Target Verification against Data Corruption


fjordan:
Ghostferry does not provide any way to ensure that the Target's data is not modified (corrupted) during execution. Now that support for interruptibility (https://github.com/Shopify/ghostferry/issues/75) has shipped, the opportunity for something like this to occur has only increased.

To do this, the plan is to rely on annotations (marginalia), originally added in https://github.com/Shopify/ghostferry/pull/155 and modified to a use a default value in https://github.com/Shopify/ghostferry/pull/171. These annotations will allow Ghostferry to identify and distinguish its own DMLs against other actors. If the Target's data is detected to be corrupted, the move should fail.

Additionally, this feature should be configurable such that the user can choose to disable Target verification as there may be other known actors writing to the Target's data.

fjordan:
Closed with https://github.com/Shopify/ghostferry/pull/176

PR #174, opened 2020-04-16
Capture RowsQueryEvents; add annotations to DML events

fjordan:
Part of https://github.com/Shopify/ghostferry/issues/173.

Captures `RowsQueryEvent`s in the `BinlogStreamer` and passes them to the `DMLEvent`s. An added `Annotation` method is provided on the `DMLEvent` interface to expose annotations/marginalia in the `DMLEvent`s query.

This also includes a slight refactor of the `DMLEvent`s to accept a `DMLEventBase` directly instead of instantiating the struct directly inside each individual DML. 

Lastly, the `Annotate` method was moved into the `BinlogWriter` to support transactions and multiple statements as the `BinlogWriter` uses a `BEGIN; ...; COMMIT` syntax. Additional tests will be added with the follow-up PR in integration tests.

A follow up PR will be created to use these annotations to validate the `DMLEvent`s streamed from the `BinlogStreamer`.

/cc @Shopify/pods

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -129,6 +129,8 @@ func (s *BinlogStreamer) Run() {
 		s.binlogSyncer.Close()
 	}()
 
+	var query []byte
+
 	currentFilename := s.lastStreamedBinlogPosition.Name
 	nextFilename := s.lastStreamedBinlogPosition.Name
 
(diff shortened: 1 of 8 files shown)


hkdsun:
Continuing to dig further but initially looks like there's traces of how to handle "groups of events" (i.e. transactions) in the mysql source.. We should probably cross-check their logic against ours

https://github.com/percona/percona-server/blob/76d4e91a7668927eb866b034479fe3d34b7f4803/sql/log_event.h#L1173-L1188

fjordan:
> Annotating Batch Writer next?

The `BatchWriter` is using `stmt, err := w.stmtCache.StmtFor(w.DB, query)`, which is being annotated in the `ghostferry_db` wrapper:

```go
func (db DB) Prepare(query string) (*sqlorig.Stmt, error) {
	return db.DB.Prepare(AnnotateStmt(query, db.Marginalia))
}
```

In the follow-up PR i'll verify this further with an integration test that asserts all DMLs against the Target are annotated. Otherwise we would need to add either the `Query` and/or the `DMLEvent` to the `DataIterator`'s `RowBatch` to be asserted. Not sure if it's worth that added complexity just for the test.

jeremycole:
> The code here looks fine to me but I don't actually understand the intricacies of the binlog layout. Perhaps @jeremycole can chime in once again?

This looks fine to me!

fjordan:
Thanks everyone!

PR #176, opened 2020-04-21
Add Target verification against data corruption

fjordan:
Last part of https://github.com/Shopify/ghostferry/issues/173.

### Overview

This adds protection against Target data corruption for the process of the run until cutover is completed.

### How?

An additional `BinlogStreamer` is attached to the target and registers a `DMLEvent` callback that extracts the annotations to verify they originated from Ghostferry. If a `DMLEvent` does not have the configured annotation, the process fails with an error providing the corrupted paginationKey. 

#### Interruptibility Support
The last inspected binlog position of the Target is persisted on the `StateTracker` to ensure interrupted processes can be correctly resumed and the target verification can continue where it left off previously.

#### Configurability

It is possible to disable target verification as it is added as an option in the configuration (`SkipTargetVerification`), but the verification is enabled by default.

This option also required the use of `binlog_rows_query_log_events` with row-based replication. A runtime check for this configuration option was added to the `Ferry` that will not allow the process to start unless this option is properly configured.

/cc @Shopify/pods


ferry.go
```diff
@@ -482,24 +488,38 @@ func (f *Ferry) Start() error {
...
 	// DataIterator determines the range of IDs to copy and the time that
 	// the starting binlog coordinates are determined.
-	var pos siddontangmysql.Position
+	var sourcePos siddontangmysql.Position
+	var targetPos siddontangmysql.Position
+
 	var err error
 	if f.StateToResumeFrom != nil {
-		pos, err = f.BinlogStreamer.ConnectBinlogStreamerToMysqlFrom(f.StateToResumeFrom.MinBinlogPosition())
+		sourcePos, err = f.SourceBinlogStreamer.ConnectBinlogStreamerToMysqlFrom(f.StateToResumeFrom.MinBinlogPosition())
```
hkdsun, line 512: unchecked error.. 😥

ferry.go
```diff
@@ -583,7 +607,12 @@ func (f *Ferry) Run() {
 	}
 
 	binlogWg := &sync.WaitGroup{}
-	binlogWg.Add(2)
+
+	waitNum := 2
+	if !f.Config.SkipTargetVerification {
+		waitNum = 3
+	}
+	binlogWg.Add(waitNum)
```
hkdsun, line 615: nit: can you `Add(1)` (right before each `go func...`) instead of having this conditional?

ferry.go
```diff
@@ -482,24 +488,38 @@ func (f *Ferry) Start() error {
 	// Registering the builtin event listeners in Start allows the consumer
 	// of the library to register event listeners that gets called before
 	// and after the data gets written to the target database.
-	f.BinlogStreamer.AddEventListener(f.BinlogWriter.BufferBinlogEvents)
+	f.SourceBinlogStreamer.AddEventListener(f.BinlogWriter.BufferBinlogEvents)
 	f.DataIterator.AddBatchListener(f.BatchWriter.WriteRowBatch)
 
 	if f.inlineVerifier != nil {
```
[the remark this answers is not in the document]
fjordan: it's a result of adding it to the `InlineVerifier`. It wouldn't make sense to have the `BinlogStreamer` attach if there wasn't an `InlineVerifier` in this approach.

I'll push up an alternative that makes it a separate component

binlog_streamer.go
```diff
@@ -101,12 +101,22 @@ func (s *BinlogStreamer) ConnectBinlogStreamerToMysqlFrom(startFromBinlogPositio
 		return mysql.Position{}, err
 	}
 
+	if startFromBinlogPosition.Pos == 0 && startFromBinlogPosition.Name == "" {
```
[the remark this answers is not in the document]
fjordan: It's for added safety as we only check the presence of a `StateToResumeFrom` before making the assumption that the state has the `LastStoredTargetBinlogPositionForInlineVerifier`. Keeping the logic here was nicer then checking that field explicitly in the state before connecting. 

I'll look at changing it to one or the other

fjordan:
> I don't see why the target verifier's stopping condition is tied to the inline verifier? How did you make that decision?

The decision was primarily based on the `TargetVerifier` living within the `InlineVerifier`. If we do decide to move it to its own component that is attached directly the `Ferry` then we can set the atomic boolean directly before we call `CutoverUnlock` 👍

inline_verifier.go
```diff
@@ -569,8 +572,46 @@ func (v *InlineVerifier) binlogEventListener(evs []DMLEvent) error {
...
+		annotations, err := ev.Annotations()
+		if err != nil {
+			return err
+		}
+
+		foundAnnotation := false
+		for _, annotation := range annotations {
+			if strings.Contains(annotation, v.TargetDB.Marginalia) {
+				foundAnnotation = true
+				continue
```
shuhaowu, line 600: Is this supposed to be `break`?

inline_verifier.go
```diff
@@ -569,8 +572,46 @@ func (v *InlineVerifier) binlogEventListener(evs []DMLEvent) error {
...
+			}
+		}
+
+		if !foundAnnotation {
+			paginationKey, err := ev.PaginationKey()
+			if err != nil {
+				return err
+			}
+			return fmt.Errorf("row data with paginationKey %d on `%s`.`%s` has been corrupted", paginationKey, ev.Database(), ev.Table())
```
shuhaowu, line 609: "corrupted by a change directly in the target database at binlog position x, y"?

ferry.go
```diff
@@ -482,35 +488,63 @@ func (f *Ferry) Start() error {
...
 	// DataIterator determines the range of IDs to copy and the time that
 	// the starting binlog coordinates are determined.
-	var pos siddontangmysql.Position
+	var sourcePos siddontangmysql.Position
+	var targetPos siddontangmysql.Position
+
 	var err error
 	if f.StateToResumeFrom != nil {
-		pos, err = f.BinlogStreamer.ConnectBinlogStreamerToMysqlFrom(f.StateToResumeFrom.MinBinlogPosition())
+		sourcePos, err = f.SourceBinlogStreamer.ConnectBinlogStreamerToMysqlFrom(f.StateToResumeFrom.MinBinlogPosition())
```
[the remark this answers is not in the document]
fjordan: this method being `MinBinlogPosition()`, right? Yeah, that makes sense.

inline_verifier.go
```diff
@@ -357,6 +358,8 @@ func (v *InlineVerifier) VerifyBeforeCutover() error {
 
 func (v *InlineVerifier) VerifyDuringCutover() (VerificationResult, error) {
 	v.verifyDuringCutoverStarted.Set(true)
+	defer v.cutoverCompleted.Set(true)
```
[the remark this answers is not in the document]
fjordan: I moved this atomic boolean to the `TargetVerifier`, and we're now setting it directly before we make the callback to `/unlock` after cutover. I'm not sure if we can get much closer than that?

inline_verifier.go
fjordan: do we want the binlog file and position instead of the pagination key? I guess we could have both?

config.go
```diff
@@ -490,6 +494,21 @@ type Config struct {
...
+	// This does so by inspecting the annotations (configured as Marginalia in the DatabaseConfig above)
+	// and will fail the move unless all applicable DMLs (as identified by the sharding key) sent to the
+	// Target were sent from Ghostferry.
+	//
+	// NOTE:
+	// The Target database must be configured with binlog_rows_query_log_events
+	// set to "ON" for this to function properly. Ghostferry not allow the move
+	// process to begin if this is enabled and the above option is set to "OFF".
+	//
+	// Required: defaults to false
```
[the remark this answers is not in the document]
fjordan: slight confusing, but the PR says target verification is _enabled_ by default. This option _disables_ it. A bit of a double-negative, is it worth changing?

binlog_streamer.go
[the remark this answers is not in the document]
fjordan: removed this and added a check in the ferry to ensure that the position is not empty before attempting to resume

ferry.go
```diff
@@ -721,7 +765,14 @@ func (f *Ferry) FlushBinlogAndStopStreaming() {
...
 
-	f.BinlogStreamer.FlushAndStop()
+	f.SourceBinlogStreamer.FlushAndStop()
+}
+
+func (f *Ferry) Stop() {
+	if !f.Config.SkipTargetVerification {
+		f.TargetBinlogStreamer.FlushAndStop()
+		f.targetVerifierWg.Wait()
+	}
```
hkdsun, line 775: Not sure about the purpose of this. Why do we need both this and the calls to `r.Ferry.TargetVerifier.CutoverCompleted.Set(true)`

Can they all be rolled up into one thing?

target_verifier.go
```diff
@@ -0,0 +1,55 @@
...
+	sql "github.com/Shopify/ghostferry/sqlwrapper"
+	"github.com/sirupsen/logrus"
+)
+
+type TargetVerifier struct {
+	logger           *logrus.Entry
+	DB               *sql.DB
+	StateTracker     *StateTracker
+	CutoverCompleted AtomicBoolean
+}
```
hkdsun, line 15: nice 😃 thank you

target_verifier.go
```diff
@@ -0,0 +1,55 @@
...
+
+func NewTargetVerifier(targetDB *sql.DB, stateTracker *StateTracker, binlogStreamer *BinlogStreamer) (*TargetVerifier, error) {
```
hkdsun, line 17: `binlogStreamer` is unused

ferry.go
```diff
@@ -508,9 +534,13 @@ func (f *Ferry) Start() error {
...
 	// is terminated with some rows copied but no binlog events are written.
 	// This guarentees that we are able to restart from a valid location.
-	f.StateTracker.UpdateLastResumableBinlogPosition(pos)
+	f.StateTracker.UpdateLastResumableSourceBinlogPosition(sourcePos)
 	if f.inlineVerifier != nil {
-		f.StateTracker.UpdateLastResumableBinlogPositionForInlineVerifier(pos)
+		f.StateTracker.UpdateLastResumableSourceBinlogPositionForInlineVerifier(sourcePos)
+
+		if !f.Config.SkipTargetVerification {
+			f.StateTracker.UpdateLastResumableTargetBinlogPositionForInlineVerifier(targetPos)
```
hkdsun, line 542: `ForInlineVerifier`? 😕

fjordan: Oops 🤦

config.go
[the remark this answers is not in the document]
shuhaowu: I still think we should default target verification to be disabled in copydb and enabled in sharding, but we can do this another time.

ferry.go
```diff
@@ -508,9 +534,13 @@ func (f *Ferry) Start() error {
 	// If we don't set this now, there is a race condition where Ghostferry
 	// is terminated with some rows copied but no binlog events are written.
 	// This guarentees that we are able to restart from a valid location.
-	f.StateTracker.UpdateLastResumableBinlogPosition(pos)
+	f.StateTracker.UpdateLastResumableSourceBinlogPosition(sourcePos)
 	if f.inlineVerifier != nil {
-		f.StateTracker.UpdateLastResumableBinlogPositionForInlineVerifier(pos)
+		f.StateTracker.UpdateLastResumableSourceBinlogPositionForInlineVerifier(sourcePos)
+
+		if !f.Config.SkipTargetVerification {
```
shuhaowu, line 541: This is saying that TargetVerifier only runs if InlineVerifier runs?

sharding/sharding.go
```diff
@@ -172,6 +172,8 @@ func (r *ShardingFerry) Run() {
 
 	r.Ferry.Throttler.SetDisabled(false)
 
+	r.Ferry.StopTargetVerifier()
```
shuhaowu, line 175: This new call is essential for all Ghostferry based application now? It's not super intuitive imo. We should update or docs then.

target_verifier.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type TargetVerifier struct {
+	logger         *logrus.Entry
+	DB             *sql.DB
+	BinlogStreamer *BinlogStreamer
+	StateTracker   *StateTracker
+}
+
+func NewTargetVerifier(targetDB *sql.DB, stateTracker *StateTracker, binlogStreamer *BinlogStreamer) (*TargetVerifier, error) {
```
shuhaowu, line 17: Traditionally I've used `Ferry` as a sort of factory, so you don't have to pass it all these variables (or worse, pass in the wrong variables). It's also somewhat natural to think of Ferry to be a factory for all the components of Ghostferry. Should this component not follow that convention? Do we not like the other convention?

ferry.go
fjordan: good catch - will fix

sharding/sharding.go
fjordan: Planning to update the docs, yes. Unfortunately, there isn't really a better option as we need this to continue running even after `Ferry.Run()` completes. We need to extend the API of the Ferry somehow to stop the TargetVerification.

target_verifier.go
fjordan: Do we want to keep that convention? It also makes some implicit decisions about what we want to use for the struct members that are built there. I can move it if so, but I thought I would take this opportunity to introduce a different pattern typically seen in go projects.

fjordan: Will keep as-is for now, and we can move it into the Ferry later if we decide we want to keep that pattern 👍

PR #177, opened 2020-04-23
Allow column rewrites in RowBatch

tiwilliam:


row_batch.go
```diff
@@ -60,6 +62,19 @@ func (e *RowBatch) AsSQLQuery(schemaName, tableName string) (string, []interface
...
+	if renamedColumns == nil {
+		return
+	}
+
+	for i, column := range e.table.Columns {
+		if newName := renamedColumns[column.Name]; newName != "" {
+			e.table.Columns[i].Name = newName
+		}
+	}
+}
```
hkdsun, line 76: Interesting PR; I think we'd wanna support renames in `dml_events.go` as well if this was to be merged.

Can you share a bit about your usecase? mostly out of curiosity

tiwilliam: Yeah, there are definitely some more work needed to get this merge ready.

There is basically one use case I had in mind when playing with this:

You start sharding your application on domain level and move out one table. You update your application to start writing to the new storage but continue to read older history from the old storage to save time when in hurry for more write capacity. If you at a later point decide to backfill history data to your new database you have a new AUTO_INCREMENT and have to re-ID your history (e.g by renaming the original `id` column to `old_id` and add a UNIQUE index to it in your target).

tiwilliam: Hm, when thinking some more, I think you can do the above with two instances of ghostferry instead if we would start to allow copying to an already existing table. First process copying from new DB to a new table in new DB, the other process copying from the old DB to the new table in the new DB.

PR #179, opened 2020-04-27
Ignore termination signals during cutover

xliang6:
Since we are so close to being done during the cutover state, it makes sense to ignore the termination signals and complete the database replication/move.

cc @Shopify/pods

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -570,8 +570,8 @@ func (f *Ferry) Run() {
 					f.ErrorHandler.Fatal("user_interrupt", fmt.Errorf("signal received: %v", s.String()))
 				} else {
 					// Log and ignore the signal during cutover
-					f.logger.Warnf("Received signal: %s during cutover. " + 
-								   "Refusing to interrupt and will attempt to complete the run.", s.String());
+					f.logger.Warnf("Received signal: %s during cutover. "+
+						"Refusing to interrupt and will attempt to complete the run.", s.String())
 				}
 			} else {
 				// shutdown() has been called and Ghostferry is done.
--- a/ferry.go
+++ b/ferry.go
@@ -11,6 +11,7 @@ import (
 	"os/signal"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -81,7 +82,7 @@ type Ferry struct {
 
 	StartTime    time.Time
 	DoneTime     time.Time
-	OverallState string
+	OverallState atomic.Value
 
 	logger *logrus.Entry
 
@@ -289,7 +290,7 @@ func (f *Ferry) NewIterativeVerifier() (*IterativeVerifier, error) {
 // Initialize all the components of Ghostferry and connect to the Database
 func (f *Ferry) Initialize() (err error) {
 	f.StartTime = time.Now().Truncate(time.Second)
-	f.OverallState = StateStarting
+	f.OverallState.Store(StateStarting)
 
 	f.logger = logrus.WithField("tag", "ferry")
 	f.rowCopyCompleteCh = make(chan struct{})
@@ -520,7 +521,7 @@ func (f *Ferry) Start() error {
 // Wait for the background tasks to finish.
 func (f *Ferry) Run() {
 	f.logger.Info("starting ferry run")
-	f.OverallState = StateCopying
+	f.OverallState.Store(StateCopying)
 
 	ctx, shutdown := context.WithCancel(context.Background())
 
@@ -564,7 +565,14 @@ func (f *Ferry) Run() {
 			s := <-c
 			if ctx.Err() == nil {
 				// Ghostferry is still running
-				f.ErrorHandler.Fatal("user_interrupt", fmt.Errorf("signal received: %v", s.String()))
+				if f.OverallState.Load() != StateCutover {
+					// Save state dump and exit if not during the cutover stage
+					f.ErrorHandler.Fatal("user_interrupt", fmt.Errorf("signal received: %v", s.String()))
+				} else {
+					// Log and ignore the signal during cutover
+					f.logger.Warnf("Received signal: %s during cutover. " + 
+								   "Refusing to interrupt and will attempt to complete the run.", s.String());
+				}
 			} else {
 				// shutdown() has been called and Ghostferry is done.
 				os.Exit(0)
@@ -614,7 +622,7 @@ func (f *Ferry) Run() {
 
 	if f.Verifier != nil {
 		f.logger.Info("calling VerifyBeforeCutover")
-		f.OverallState = StateVerifyBeforeCutover
+		f.OverallState.Store(StateVerifyBeforeCutover)
 
 		metrics.Measure("VerifyBeforeCutover", nil, 1.0, func() {
 			err := f.Verifier.VerifyBeforeCutover()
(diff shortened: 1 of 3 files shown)


xliang6:
cc @Shopify/pods

ferry.go
```diff
@@ -50,6 +50,23 @@ func MaskedDSN(c *mysql.Config) string {
...
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	return s.state
+}
+
+func (s *SafeState) Update(state string) {
+	s.mutex.Lock()
+	s.state = state
+	s.mutex.Unlock()
+}
```
hkdsun, line 68: 👌 

You can use the `atomic` package (`atomic.Value`) if you wanted to simplify this

test/integration/interrupt_resume_test.rb
```diff
@@ -77,8 +77,8 @@ def test_interrupt_resume_with_writes_to_source
     assert_test_table_is_identical
   end
 
-  def test_interrupt_resume_when_table_has_completed
-    # Start a run of Ghostferry expecting to be interrupted
+  def test_interrupt_ignored_when_table_has_completed
```
hkdsun, line 80: > Ghostferry::Status::ROW_COPY_COMPLETED

I assume that state maps to `StateCutover` in the main Ferry code?

ferry.go
```diff
@@ -564,7 +581,13 @@ func (f *Ferry) Run() {
 			s := <-c
 			if ctx.Err() == nil {
 				// Ghostferry is still running
-				f.ErrorHandler.Fatal("user_interrupt", fmt.Errorf("signal received: %v", s.String()))
+				if f.OverallState.Value() != StateCutover {
+					// Save state dump and exit if not during the cutover stage
+					f.ErrorHandler.Fatal("user_interrupt", fmt.Errorf("signal received: %v", s.String()))
+				} else {
+					// Log and ignore the signal during cutover
+					f.logger.Warn("Signal received: %s during cutover, will ignore and complete the move", s.String())
```
hkdsun, line 589:
```suggestion
					f.logger.Warnf("Signal received: %s during cutover, will ignore and complete the move", s.String())
```

To correctly format this

ferry.go
xliang6: I don't see the `sync/atomic` package dealing with strings?

ferry.go
hkdsun: Could you add an assertion for this log line in test?

Similar to https://github.com/Shopify/ghostferry/blob/5ec4ab591e3257f8f21bfa5b4d083746d688c66a/test/integration/trivial_test.rb#L33-L39

ferry.go
hkdsun: `atomic.Value` deals with any type, just need to `Load` and `Store` a consistent type across calls

test/integration/interrupt_resume_test.rb
xliang6: Yes, ROW_COPY_COMPLETED is received after data copy is done and ferry state has just been set to `StateCutover`: https://github.com/Shopify/ghostferry/blob/d2df94173c6deea7d99d463cb9bc17b98c183129/ferry.go#L634

ferry.go
xliang6: Good catch, I'll fix it.

ferry.go
xliang6: I see, let me take a look.

xliang6:
> Awesome work. LGTM if you can add the log assertion

Thanks for the review @hkdsun. I have addressed all your comments.

ferry.go
```diff
@@ -564,7 +565,13 @@ func (f *Ferry) Run() {
 			s := <-c
 			if ctx.Err() == nil {
 				// Ghostferry is still running
-				f.ErrorHandler.Fatal("user_interrupt", fmt.Errorf("signal received: %v", s.String()))
+				if f.OverallState.Load() != StateCutover {
+					// Save state dump and exit if not during the cutover stage
+					f.ErrorHandler.Fatal("user_interrupt", fmt.Errorf("signal received: %v", s.String()))
+				} else {
+					// Log and ignore the signal during cutover
+					f.logger.Warnf("Signal received: %s during cutover, will ignore and complete the move", s.String())
```
fjordan, line 573:
```suggestion
					f.logger.Warnf("Received signal: %s during cutover. Refusing to interrupt and will attempt to complete the run. ", s.String())
```

ferry.go
```diff
@@ -646,7 +653,7 @@ func (f *Ferry) Run() {
 	binlogWg.Wait()
 
 	f.logger.Info("ghostferry run is complete, shutting down auxiliary services")
-	f.OverallState = StateDone
+	f.OverallState.Store(StateDone)
```
fjordan, line 656: Not particularly related to your PR, but I noticed we are setting `StateDone` here, but it isn't actually until well after that we actually post the cutover unlock (in `sharding`, not in `copydb`).

ferry.go
```diff
@@ -832,7 +839,7 @@ func (f *Ferry) ensureInitialized() {
 	// Note: the constructor shouldn't have a large amount of positional argument
 	//       so it is more readable.
 	//
-	if f.OverallState == "" {
+	if f.OverallState.Load() == "" {
```
fjordan, line 842: If this isn't set, it is still an interface, and should be compared against `nil`, right? 

```go
package main

import (
	"sync/atomic"
)

func main() {
	var m atomic.Value
	
	if m.Load() == "" {
		panic("not initialized")
	}
	
	if m.Load() == nil {
		println("m is nil")
	}
}

```
the above evaluates to:
```
m is nil
```

and does not panic.

test/integration/interrupt_resume_test.rb
```diff
@@ -89,19 +89,18 @@ def test_interrupt_resume_when_table_has_completed
...
+    assert_test_table_is_identical
 
-    start_datawriter_with_ghostferry(datawriter, ghostferry)
-    stop_datawriter_during_cutover(datawriter, ghostferry)
+    assert ghostferry.logrus_lines["ferry"].length > 0
 
-    ghostferry.run(dumped_state)
+    ghostferry.logrus_lines["ferry"].each do |line|
+      if line["msg"].start_with?("Signal ")
+        assert line["msg"].match?("Signal received: terminated during cutover, will ignore and complete the move")
```
fjordan, line 100: Nice 👍

ferry.go
xliang6: Good catch!

ferry.go
xliang6: `StateDone` signals that ghostferry is done, and shop mover cutover unlock signals the shop move is done?

test/integration/interrupt_resume_test.rb
```diff
@@ -89,19 +89,19 @@ def test_interrupt_resume_when_table_has_completed
...
+    assert_test_table_is_identical
 
-    start_datawriter_with_ghostferry(datawriter, ghostferry)
-    stop_datawriter_during_cutover(datawriter, ghostferry)
+    assert ghostferry.logrus_lines["ferry"].length > 0
 
-    ghostferry.run(dumped_state)
+    ghostferry.logrus_lines["ferry"].each do |line|
+      if line["msg"].start_with?("Received signal: ")
+        assert line["msg"].match?("Received signal: terminated during cutover. " \
```
hkdsun, line 100: and now we need an assertion that the assertion ran 😛 

(because there's no guarantee that any of the array elements `start_with?("Received signal")`)

ferry.go
xliang6: > This is my bad probably for suggesting using `atomic.Value` 🤦 sorry Xiaoli

No worries. ❤️

test/integration/interrupt_resume_test.rb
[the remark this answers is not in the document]
xliang6: Sure, I'll add that. Also I'm adding the following block around the asserts. I just found out that during CI, ghostferry.logrus_lines["ferry"] is null.

```
    with_env('CI', nil) do
      assert ghostferry.logrus_lines["ferry"].length > 0
      ...
    end
```

ferry.go
```diff
@@ -832,7 +840,7 @@ func (f *Ferry) ensureInitialized() {
 	// Note: the constructor shouldn't have a large amount of positional argument
 	//       so it is more readable.
 	//
-	if f.OverallState == "" {
+	if f.OverallState.Load() == nil || f.OverallState.Load() == "" {
```
fjordan, line 843: 👍

PR #180, opened 2020-04-28
Allow target table to already exist in CopydbFerry

tiwilliam:
This will make it configurable to allow an already existing table in target. Default will be to not allow it, just like today.

There are a few use-cases where this could make sense, for example if you want to make changes to a large table before you fill it with data (eg. new defaults, new composite primary key or new indices). Another use case is if you already have a table with data that you want to pad with data from your source database.

### DIFF
--- a/copydb/config.go
+++ b/copydb/config.go
@@ -72,7 +72,8 @@ type Config struct {
 	WaitForReplicationTimeout string
 
 	// Ghostferry will by default create tables on your target and fail if these already exist.
-	// This allows pre-existing tables on your target, schemas won't be validated.
+	// This allows pre-existing tables on your target, schemas and data compatibility won't be validated.
+	// Use at your own risk.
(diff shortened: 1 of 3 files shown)


copydb/config.go
```diff
@@ -70,6 +70,10 @@ type Config struct {
 
 	// The duration to wait for the replication to catchup before aborting. Only use if RunFerryFromReplica is true.
 	WaitForReplicationTimeout string
+
+	// Ghostferry will by default create tables on your target and fail if these already exist.
+	// This allows pre-existing tables on your target, schemas won't be validated.
```
hkdsun, line 75:
```suggestion
	// This allows pre-existing tables on your target, schemas and data compatibility won't be validated.
	// Use at your own risk
```

I feel like this should have a harder disclaimer 😄

tiwilliam:
Thanks!

PR #185, opened 2020-05-12
Fix resuming state without inline verifier

### LINKED ISSUE #184: Resuming ghostferry may fail due to inactive verifier
The inline verifier uses a "resume position" for keeping track of the last event that it has seen. Similar to the binlog writer's "resume position", `ghostferry` uses this verifier resume position to continue its job after a restart of the tool.

`ghostferry` uses the smaller value of _inline verifier_ and _binlog writer_ resume position to decide what's the position on the source DB from which to stream binlogs.

If the inline-verifier is not enabled enabled at all, its resume position can grow stale. In some cases it points to such an old position that resuming from it fails (if the source has already deleted such old replication logs).

While having a inline verifier that cannot keep up with the event frequency is a problem that can't really be solved. But we can tackle the trivial case where the verifier is not enabled at all (and we inevitably will get to a point where the initial position will become invalid)

kolbitsch-lastline:
If the inline-verifier is not enabled (as is the case for various
production uses of ghostferry), its binlog position can grow stale. In
some cases it points to such an old position that resuming from it
fails (if the source has already deleted such old replication logs).

This commit fixes this by relying solely on the binlog writer resume
position if the inline verifier is not enabled.

We still fail if the inline verifier *is* enabled and the position is
stale, but there is nothing one can do about that. If verification is
enabled, one must ensure that it's able to keep up with migration.

This fixes #184

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -508,14 +508,25 @@ func (f *Ferry) Start() error {
 	// miss some records that are inserted between the time the
 	// DataIterator determines the range of IDs to copy and the time that
 	// the starting binlog coordinates are determined.
+	//
+	// NOTE: If we don't use the inline verifier, we don't consider its last
+	// position for the resume position. We could be migrating for a long time,
+	// and the inline verifier position may grow outdated to the point that it
+	// is no-longer a valid position (the logs could have been deleted). Since
+	// the inline verifier position is not updated if it's not enabled and we
+	// use the oldest position in `MinBinlogPosition()`, resume may fail for
+	// the minimum position.
+	// In this case, using the last written position is the better state to use
 	var sourcePos siddontangmysql.Position
 	var targetPos siddontangmysql.Position
 
 	var err error
-	if f.StateToResumeFrom != nil {
+	if f.StateToResumeFrom == nil {
+		sourcePos, err = f.BinlogStreamer.ConnectBinlogStreamerToMysql()
+	} else if f.inlineVerifier != nil {
 		sourcePos, err = f.BinlogStreamer.ConnectBinlogStreamerToMysqlFrom(f.StateToResumeFrom.MinSourceBinlogPosition())
 	} else {
-		sourcePos, err = f.BinlogStreamer.ConnectBinlogStreamerToMysql()
+		sourcePos, err = f.BinlogStreamer.ConnectBinlogStreamerToMysqlFrom(f.StateToResumeFrom.LastWrittenBinlogPosition)
 	}
 	if err != nil {
 		return err


shuhaowu:
Thanks for the PR! This feels like a problem somewhere else tho, since we have a guard for this condition in both of these places: https://github.com/Shopify/ghostferry/blob/04dda8500d9ae764306e94b5115988a1986a41b6/state_tracker.go#L51-L53 and https://github.com/Shopify/ghostferry/blob/50a7760c383407e09fe0efcded4ac27ebf3e4b8f/ferry.go#L539-L541

kolbitsch-lastline:
thanks for the quick feedback!

yeah, I also considered that approach, but I felt that adding more checks in multiple places isn't the cleanest solution.

On start, you can't really fix the problem (here https://github.com/Shopify/ghostferry/blob/50a7760c383407e09fe0efcded4ac27ebf3e4b8f/ferry.go#L539-L541), as we need to _continuously_ update that counter, but if the inline verifier is not running, that code doesn't even get event notifications.

I considered moving it into `MinSourceBinlogPosition` but that felt like propagating a flag into a method that doesn't need to care about it (hence I moved the `if` into the call).

We could also register a "dummy event listener" that simply sets the last-seen binlog position for the inline verifier. This way one could resume the verifier at any point. Not sure if that's cleaner than the simple `if` though.

hkdsun:
From what I'm reading, if you start the ferry with the inline verifier disabled, we never update the state tracker's `lastStoredBinlogPositionForInlineVerifier`.. The only possible way to get here is then to start a ferry with the verifier enabled, interrupt, and then resume with it disabled.

Is there failing test or repro steps you can demonstrate?

hkdsun:
I think that is what happened... based on #184 .

How about instead of this change, we clear the state from the SerializedState?

That is, if across invocations you disable the verifier, we clear all its previous state.

And this won't reach into the `MinSourceBinlogPosition` implementation.

Something like:

```
	if f.StateToResumeFrom != nil {
		if f.inlineVerifier == nil {
			f.StateToResumeFrom.LastStoredBinlogPositionForInlineVerifier = siddontangmysql.Position{}
		}
		sourcePos, err = f.BinlogStreamer.ConnectBinlogStreamerToMysqlFrom(f.StateToResumeFrom.MinSourceBinlogPosition())
	} else {
```

kolbitsch-lastline:
> How about instead of this change, we clear the state from the SerializedState?
>
> That is, if across invocations you disable the verifier, we clear all its previous state.

that a good suggestion, but there is a reason I did not follow this approach: conceptually, it is very nice to have the **possibility** to **temporarily** disable the verifier. If we were to clear the state, that means we cannot pick back up where we had left off.

Furthermore, in many cases, there is actually no need to clear the state, because resuming later is perfectly safe. The problem arises only once the source DB can no-longer serve binlog data starting at the resume point. Sure, sooner or later this will always happen, but you don't disable the verifier for (typically) weeks, there is no harm.

Thus, the current implementation seems more generic and safer, but I don't feel strongly about this at all and am happy to change how this happens

PR #187, opened 2020-05-19
Bump go-mysql dependency

tiwilliam:
Current go-mysql vendor is from 2018 and don't include support for UNIX sockets. This was fixed 15 commits after our frozen commit. This pull will bring in the latest go-mysql.

We also set source and target port to 0 when using UNIX sockets since this is how go-mysql distinguish a UNIX config:
https://github.com/siddontang/go-mysql/blob/master/replication/binlogsyncer.go#L829


go.mod
```diff
@@ -23,11 +23,11 @@ require (
 	github.com/ooyala/go-dogstatsd v0.0.0-20140922214459-23f2a1659b02 // indirect
 	github.com/pingcap/check v0.0.0-20200212061837-5e12011dc712 // indirect
 	github.com/satori/go.uuid v1.2.0 // indirect
-	github.com/shopspring/decimal v0.0.0-20171108225254-9ca7f51822d2
+	github.com/shopspring/decimal v0.0.0-20180709203117-cd690d0c9e24
 	github.com/siddontang/go v0.0.0-20180604090527-bdc77568d726 // indirect
 	github.com/siddontang/go-log v0.0.0-20180807004314-8d05993dda07
-	github.com/siddontang/go-mysql v0.0.0-20180802024903-58848a70cf1a
-	github.com/sirupsen/logrus v1.0.4
+	github.com/siddontang/go-mysql v0.0.0-20200424072754-803944a6e4ea
```
shuhaowu, line 29: We need to audit the changes from the version we pin until this new version as this is a high risk change.

tiwilliam: Yeah, kind of understood that, 89 commits, 4,656 additions and 18,017 deletions.

https://github.com/siddontang/go-mysql/compare/58848a70cf1a...803944a6e4ea

shuhaowu: I'll look through that list and pull out the highlights in a comment below.

shuhaowu:
#### go-mysql changes

- https://github.com/siddontang/go-mysql/commit/8939dbe3092047486e3c5161c1d9ac0470394cec (PR: https://github.com/siddontang/go-mysql/pull/305): A bug fix with respect to `TableMapEvent`?
- https://github.com/siddontang/go-mysql/commit/0f2e7f60ec933995a557757ca7445fc7a9cb10a3 (PR: https://github.com/siddontang/go-mysql/pull/319): some bugfixes when handling binlogs with SET, ENUM, BIT types. It looks like previously the decoded binlog was incorrect?
  - https://github.com/siddontang/go-mysql/commit/83f8eaeb3515872afb53949cf11dd54d691cac60 (PR: https://github.com/siddontang/go-mysql/pull/320): More ENUM test cases.
- https://github.com/siddontang/go-mysql/commit/76c059e36c620bb2e5f8471092f9d9b22920e052 (PR: https://github.com/siddontang/go-mysql/pull/332): `schema.TableColumn` now supports DECIMAL.
  - I recall that we had to support DECIMAL at some point, this commit adds DECIMAL support to `canal` (not something we use). We should see if we can use the new DECIMAL option on `TableColumn`.
- https://github.com/siddontang/go-mysql/commit/8b4711353793b5410c8dee9f44377bb11f9aa728 (PR: https://github.com/siddontang/go-mysql/pull/312): Added support to MySQL 8, but this mostly seems like changes with the go-mysql/client code. Replication code looks minimally changed.
  - A bit while later we see that the table map event gets the table schema? https://github.com/siddontang/go-mysql/commit/514d075349668f938db1cba08de7e611b8de1f25 (PR: https://github.com/siddontang/go-mysql/pull/468).
    - This could be extremely helpful once we move to MySQL 8, but the code is hard to review quickly now as it's 800+ lines as well as a lot of MySQL internals.
  - More MySQL 8 binlog changes:
    - PR: https://github.com/siddontang/go-mysql/pull/468 (multiple commits)
- https://github.com/siddontang/go-mysql/commit/7fc3b28d61044c8d3ee4062c3b96ed4a1307b862 (PR: https://github.com/siddontang/go-mysql/pull/372): More JSON fixes that works around MySQL bugs that I don't quite understand.
- https://github.com/siddontang/go-mysql/commit/8b9c05ee162eed5710b25325eaaa978f00470f91 (PR: https://github.com/siddontang/go-mysql/pull/399): Handling unsigned MEDIUMINT seemed to require some nudges in canal, which implies that it may require some code changes here in Ghostferry.
  - MEDIUMINT is 3 byte which can cause problems when casting.
- https://github.com/siddontang/go-mysql/commit/046188b858f9a4b47cdedb11068eb868463fcb75 (PR: https://github.com/siddontang/go-mysql/pull/397): They're ignoring fake rotate event in canal. We should cross reference our fix for this with theirs.
  - [Original bug report](https://github.com/siddontang/go-mysql/issues/396) for this is in Chinese, which appears to describe a situtation where Canal can be restarted from a binlog position after a table map event (in their case it seems to stem from a delete table, which generates a large amount of binlog entries). Some explanation on how to fix is in English. 
  - More related commits: https://github.com/siddontang/go-mysql/commit/cb33eb8bb30f5728cdfe53b031e882e60fe85cec (PR: https://github.com/siddontang/go-mysql/pull/428)
    - This appears to be address how binlog file names are handled in Canal, which we've also independently worked on.
  - We should evaluate if our fix for resumable binlog position matches theirs, cc: @fjordan 
- https://github.com/siddontang/go-mysql/commit/803944a6e4ea9655a4912ad6af3febd5bf04a254 (PR: https://github.com/siddontang/go-mysql/commit/803944a6e4ea9655a4912ad6af3febd5bf04a254) Binary column fix by our friend @kolbitsch-lastline
  - I don't recall the status of how we handled the BINARY column patch. Maybe a good idea to revisit it now. 

Nothing looks too crazy. MySQL 8 changes is great, but pretty new. I haven't had a chance to review all the code in PR 468 and if it would have any impact on us as we don't use MySQL 8. Some bug fixes in Canal caught my eye as we independently did this. Also saw a lot of work on GTID stuff that we don't use.

shuhaowu:
Also interesting to note that once we merge this theoretically we can support MySQL 8... but that requires extensive validation...

tiwilliam:
Changes look good to me. I've been tested this branch using copydb on a dry-run attempt to move 1 TB of data via UNIX sockets. Everything went smooth.

PR #189, opened 2020-06-16
Cast OverallState to string in StatusDeprecated

tiwilliam:
Fix crash when rendering web UI introduced in Shopify/ghostferry#179.
> template: index.html:82:24: executing "index.html" at : error calling eq: incompatible types for comparison

### DIFF
--- a/status_deprecated.go
+++ b/status_deprecated.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"math"
 	"sort"
-	"sync/atomic"
 	"time"
 
 	"github.com/siddontang/go-mysql/mysql"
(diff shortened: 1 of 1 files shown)


status_deprecated.go
```diff
@@ -64,7 +63,7 @@ func FetchStatusDeprecated(f *Ferry, v Verifier) *StatusDeprecated {
 	status.SourceHostPort = fmt.Sprintf("%s:%d", f.Source.Host, f.Source.Port)
 	status.TargetHostPort = fmt.Sprintf("%s:%d", f.Target.Host, f.Target.Port)
 
-	status.OverallState.Store(f.OverallState.Load())
+	status.OverallState = fmt.Sprintf("%s", f.OverallState.Load())
```
xliang6, line 66: Alternatively, I haven't verified it myself but I think something like this one-line change will work too: https://github.com/Shopify/ghostferry/blob/1ad39c7ab99f58c2d5858e01f0ab734e789ddeff/ferry.go#L798.

PR #192, opened 2020-06-24
Enhancements to Go tests

tiwilliam:
Switch to `gotestsum` for easier to read test output from `go test`. It will only expand stdout on failure and show time consumed for each test. I've also added a slow test output for tests slower than 5 seconds which we could focus to fix.

I've also tweaked the seed sizes a bit and a Go test run is now 2x faster than before.

Example output:
```
...
PASS sharding/test.TestShardedTableFilterRejectsIgnoredTables (0.00s)
PASS sharding/test.TestShardedTableFilterSelectsTablesWithShardingKey (0.00s)
PASS sharding/test.TestShardedTableFilterSelectsJoinedTables (0.00s)
PASS sharding/test.TestSelectiveCopyDataWithoutAnyWritesToSource (0.69s)
PASS sharding/test.TestSelectiveCopyDataWithInsertLoadOnOtherTenants (1.20s)
PASS sharding/test.TestSelectiveCopyDataWithInsertLoadOnAllTenants (1.21s)
PASS sharding/test.TestErrorsIfShardingKeyChanged (1.21s)
PASS sharding/test
DONE 229 tests in 113.127s
Slow Go tests:
github.com/Shopify/ghostferry/test/go TestIterativeVerifierCollationTestSuite 19.58s
github.com/Shopify/ghostferry/test/go TestIterativeVerifierTestSuite 18.36s
github.com/Shopify/ghostferry/sharding/test TestCallbacksTestSuite 14.56s
github.com/Shopify/ghostferry/sharding/test TestPrimaryKeyTableTestSuite 6.46s
```


shuhaowu:
As a note: one of my long term goals is to remove all Golang integration tests and move all integration tests into Ruby. This means eventually the go tests should only have unit tests and be very fast.

tiwilliam:
Yeah, saw that somewhere. We should reduce the slow test warning threshold when that happens and nag more about it to keep them fast.

tiwilliam:
I've now also split Ruby and Go tests to two separate jobs on Travis.

Makefile
```diff
@@ -50,10 +53,21 @@ $(PROJECT_DEBS): reset-deb-dir
...
-test:
+test-go:
 	@go version
-	go test ./test/go ./copydb/test ./sharding/test -p 1 -v -count 1
-	bundle install && bundle exec rake test DEBUG=1 TESTOPTS="-v"
+	ulimit -n 1024
+
+	@if [ ! -f $(GOBIN)/gotestsum ]; then \
+		curl -sL $(GOTESTSUM_URL) | tar -xz -C $(GOBIN) gotestsum; \
+	fi
```
shuhaowu, line 62: Installing a random binary to people's GOPATH/bin as a part of the side effects of the test seems a bit sketchy.. especially since this is a public project.

Maybe we can temporarily install this to `./bin/`? At some point in the near-ish future I can convert this project using `nix-shell` (and dev.yml with nix).

Makefile
```diff
@@ -30,6 +30,9 @@ PROJECT_BIN      = ghostferry-$(proj)
 BIN_TARGET       = $(GOBIN)/$(PROJECT_BIN)
 DEB_TARGET       = $(BUILD_DIR)/$(PROJECT_BIN)_$(VERSION_STR).deb
 
+PLATFORM        := $(shell uname -s | tr A-Z a-z)
+GOTESTSUM_URL   := "https://github.com/gotestyourself/gotestsum/releases/download/v0.5.1/gotestsum_0.5.1_$(PLATFORM)_amd64.tar.gz"
+
 .PHONY: test clean reset-deb-dir $(PROJECTS) $(PROJECT_DEBS)
```
shuhaowu, line 36: Need to add `test-go` and `test-ruby` here.

Makefile
tiwilliam: You are right, want to avoid adding the dependencies to the project. Changing it to `./bin` as you suggest.

PR #193, opened 2020-06-24
Enabled InlineVerifier to work with ControlServer

shuhaowu:
The InlineVerifier did not implement interface needed by the ControlServer to start VerifyDuringCutover when cutover is occuring. As a result, using the InlineVerifier with copydb will result in the binlogs not verified.

This commit implements the aforementioned interface by adapting the code from the IterativeVerifier.

FYI: I'm still setting up a local test setup to test this as we don't have test coverage for this :(

### DIFF
--- a/inline_verifier.go
+++ b/inline_verifier.go
@@ -3,6 +3,7 @@ package ghostferry
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


inline_verifier.go
```diff
@@ -229,20 +230,49 @@ type InlineVerifier struct {
...
+	v.logger.Info("starting InlineVerifier.VerifyDuringCutover in the background")
+
+	v.backgroundVerificationWg.Add(1)
+	go func() {
+		defer func() {
+			v.backgroundVerificationResultAndStatus.DoneTime = time.Now()
+			v.backgroundVerificationWg.Done()
+		}()
+
+		v.backgroundVerificationResultAndStatus.VerificationResult, v.backgroundVerificationErr = v.VerifyDuringCutover()
```
fjordan, line 264:
```suggestion
		v.backgroundVerificationResultAndStatus.VerificationResult, v.backgroundVerificationErr = v.VerifyDuringCutover()
		v.verificationResultAndStatus.DoneTime = time.Now()
```

Do we not need to record the `DoneTime`? (I'm looking at the IterativeVerifier as a reference):

https://github.com/Shopify/ghostferry/blob/3113d9c4da8ac6e01b7ff75c9e852f11dcd58584/iterative_verifier.go#L274

inline_verifier.go
```diff
@@ -229,20 +230,49 @@ type InlineVerifier struct {
...
 
+// This is called from the control server, which is triggered by pushing Run
+// Verification during cutover.
+// This step is necessary to ensure the binlogs are verified in Ghostferry.
 func (v *InlineVerifier) StartInBackground() error {
-	// not needed?
+	if v.verifyDuringCutoverStarted.Get() {
+		return errors.New("verification during cutover has already been started")
+	}
+
```
fjordan, line 247: Also looks like we may be missing some other checks that exist in the `IterativeVerifier`?

https://github.com/Shopify/ghostferry/blob/3113d9c4da8ac6e01b7ff75c9e852f11dcd58584/iterative_verifier.go#L245-L251

inline_verifier.go
fjordan: turns out this is set above in the `defer` 👍

inline_verifier.go
shuhaowu: Done.

PR #196, opened 2020-06-25
Display additional verifier info in progress

shuhaowu:
This way the InlineVerifier can report how many rows is in its reverify queue, which may help sanity check in a move. Example screen shot:

Screenshot_2020-06-25 Ghostferry 127 0 0 1 29291 - 127 0 0 1 29292

I also added this to `ReportProgress`, which means any consumer of Ghostferry's API callbacks can also use this information.

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -828,6 +828,11 @@ func (f *Ferry) Progress() *Progress {
 		return true
 	})
 
+	// Verifier information
+	if f.Verifier != nil {
+		s.VerifierMessage = f.Verifier.Message()
+	}
+
 	tables := f.Tables.AsSlice()
 
 	for _, table := range tables {
--- a/inline_verifier.go
+++ b/inline_verifier.go
@@ -205,6 +205,10 @@ func (s *BinlogVerifyStore) Batches(batchsize int) []BinlogVerifyBatch {
 	return batches
 }
 
+func (s *BinlogVerifyStore) CurrentRowCount() uint64 {
+	return s.currentRowCount
+}
+
 func (s *BinlogVerifyStore) Serialize() BinlogVerifySerializedStore {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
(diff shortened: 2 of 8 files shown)


ferry.go
```diff
@@ -828,6 +828,11 @@ func (f *Ferry) Progress() *Progress {
 		return true
 	})
 
+	// Verifier information
+	if f.Verifier != nil {
```
fjordan, line 832: At some point I think we should consider removing these conditionals in favor of a single runtime check that prevents operation without a verifier. 

What do you think?

inline_verifier.go
```diff
@@ -275,6 +279,10 @@ func (v *InlineVerifier) Wait() {
 	v.backgroundVerificationWg.Wait()
 }
 
+func (v *InlineVerifier) Message() string {
+	return fmt.Sprintf("BinlogVerifyStore.currentRowCount = %d", v.reverifyStore.CurrentRowCount())
```
fjordan, line 283: Does it matter this is not _Binlog**Re**verifyStore_?

ferry.go
shuhaowu: I still think we should allow operations without a verifier. It's a bit unsafe if done carelessly thus we don't really recommend it officially. However, I can see cases where we would want to run with the verifier off (if people have performed validation runs with Ghostferry and are confident that the dataset will be OK).

IMO given that this is an open source project, we should keep the flexibility of Ghostferry intact, if it is not too much work.

inline_verifier.go
shuhaowu: The actual data structure is called BinlogVerifyStore for the InlineVerifier. We keep calling it the reverify store because that's what the IterativeVerifier does, but it's not as accurate of a name given that the InlineVerifier keeps that store small (where as the IterativeVerifier did not and let it grow basically unbounded).

shuhaowu:
Thanks for the approval.

PR #197, opened 2020-06-26
Batch tables in DataIterator

tiwilliam:
When Ghostferry transfers data using the `DataIterator`, it does so by queuing all tables that have any data. These queued tables are then processed in N goroutines which batch rows for transfer.

The problem with this approach is that when a database doesn't have evenly sized tables, the smaller tables will transfer fast in parallel and then you will be waiting for a larger which is only row batched in a single process. We have seen examples where these larger tables might take several days to complete instead of a couple of hours if all processes and bandwidth were to be used.

This greatly affects the speed of shop mover and when moving databases more manually (e.g. migrations to KateSQL).

Tables are today loaded on initializing a new `Ferry` and passed to the `DataIterator`. Tables are passed as slices of `TableSchemas`. These tables are then passed to a channel for processing in first available processes. Instead of passing tables to the channel, we could pass a fixed set of rows of each table, size of this set would be determined by the number of rows through the number of processors.

Resuming from state would simply load up the same batch sizes no matter if concurrency has changed.

Even though this will be a improvement for many cases, it still suffers from unevenly distributed PKs. This issue can be partly fixed by splitting batches in a higher factor than cores available.

### Scope
* Databases with one or a couple large tables must use all available processes at all time in DataIterator.
* Small tables must not be batched unnecessarily.
* Concurrency must remain possible to re-configure.
* Stop and resume must work (transfer progress must be saved to state).
* UI should show progress correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ test-go:
 		curl -sL $(GOTESTSUM_URL) | tar -xz -C ./bin gotestsum; \
 	fi
 
-	ulimit -n 1024 && ./bin/gotestsum --format short-verbose ./test/go ./copydb/test ./sharding/test -count 1 -p 1 -failfast
+	ulimit -n 1024 && ./bin/gotestsum --format short-verbose ./test/go ./copydb/test ./sharding/test -count 1 -p 1
 
 test-ruby:
 	bundle install
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ test-go:
 		curl -sL $(GOTESTSUM_URL) | tar -xz -C ./bin gotestsum; \
 	fi
 
-	ulimit -n 1024 && ./bin/gotestsum --format short-verbose ./test/go ./copydb/test ./sharding/test -count 1 -p 1
+	ulimit -n 1024 && ./bin/gotestsum --format short-verbose ./test/go ./copydb/test ./sharding/test -count 1 -p 1 -failfast
 
 test-ruby:
 	bundle install
--- a/data_iterator.go
+++ b/data_iterator.go
@@ -173,8 +173,8 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 			tableEndPaginationKey := batch.paginationKeys.MaxPaginationKey
 
 			// Number of batches are set to number of processes, unless each batch becomes smaller than the cursor size
-			keyInterval := tableEndPaginationKey - tableStartPaginationKey
-			concurrencyBatchSize := math.Ceil(float64(keyInterval) / float64(d.Concurrency))
+			numKeys := tableEndPaginationKey - tableStartPaginationKey
+			concurrencyBatchSize := math.Ceil(float64(numKeys) / float64(d.Concurrency))
 			batchSize := uint64(math.Max(concurrencyBatchSize, float64(d.CursorConfig.BatchSize)))
 
 			d.logger.WithFields(logrus.Fields{
@@ -182,7 +182,7 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 				"batchSize":          batchSize,
 				"endPaginationKey":   tableEndPaginationKey,
 				"startPaginationKey": tableStartPaginationKey,
-			}).Debugf("queueing %d batches", (keyInterval/batchSize)+1)
+			}).Debugf("queueing %d batches", (numKeys/batchSize)+1)
 
 			for batchStartPaginationKey := tableStartPaginationKey; batchStartPaginationKey < tableEndPaginationKey; batchStartPaginationKey += batchSize {
 				batchEndPaginationKey := batchStartPaginationKey + batchSize
--- a/data_iterator.go
+++ b/data_iterator.go
@@ -1,13 +1,20 @@
 package ghostferry
 
 import (
-	"fmt"
 	sql "github.com/Shopify/ghostferry/sqlwrapper"
 	"math"
-	"sync"
 
 	"github.com/sirupsen/logrus"
 )
+
+type DataIteratorBatch struct {
+	table          *TableSchema
+	paginationKeys *MinMaxKeys
+}
+
+func (p *DataIteratorBatch) BatchID() uint64 {
+	return p.paginationKeys.MinPaginationKey
+}
 
 type DataIterator struct {
 	DB                *sql.DB
@@ -18,25 +25,24 @@ type DataIterator struct {
 	CursorConfig *CursorConfig
 	StateTracker *StateTracker
 
-	targetPaginationKeys *sync.Map
-	batchListeners       []func(*RowBatch) error
-	doneListeners        []func() error
-	logger               *logrus.Entry
+	batches        map[string]*DataIteratorBatch
+	batchListeners []func(*RowBatch) error
+	doneListeners  []func() error
+	logger         *logrus.Entry
 }
 
 func (d *DataIterator) Run(tables []*TableSchema) {
 	d.logger = logrus.WithField("tag", "data_iterator")
-	d.targetPaginationKeys = &sync.Map{}
 
 	// If a state tracker is not provided, then the caller doesn't care about
 	// tracking state. However, some methods are still useful so we initialize
 	// a minimal local instance.
 	if d.StateTracker == nil {
-		d.StateTracker = NewStateTracker(0)
+		d.StateTracker = NewStateTracker()
 	}
 
 	d.logger.WithField("tablesCount", len(tables)).Info("starting data iterator run")
-	tablesWithData, emptyTables, err := MaxPaginationKeys(d.DB, tables, d.logger)
+	tablesWithData, emptyTables, err := MinMaxPaginationKeys(d.DB, tables, d.logger)
 	if err != nil {
 		d.ErrorHandler.Fatal("data_iterator", err)
 	}
@@ -45,61 +51,55 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 		d.StateTracker.MarkTableAsCompleted(table.String())
 	}
 
-	for table, maxPaginationKey := range tablesWithData {
+	d.batches = make(map[string]*DataIteratorBatch, len(tablesWithData))
+
+	for table, keys := range tablesWithData {
 		tableName := table.String()
 		if d.StateTracker.IsTableComplete(tableName) {
 			// In a previous run, the table may have been completed.
 			// We don't need to reiterate those tables as it has already been done.
-			delete(tablesWithData, table)
-		} else {
-			d.targetPaginationKeys.Store(table.String(), maxPaginationKey)
+			d.logger.WithField("table", tableName).Warn("table already completed, skipping")
+			continue
 		}
+		d.batches[tableName] = &DataIteratorBatch{table: table, paginationKeys: keys}
 	}
 
-	tablesQueue := make(chan *TableSchema)
-	wg := &sync.WaitGroup{}
-	wg.Add(d.Concurrency)
+	batchQueue := make(chan *DataIteratorBatch)
 
 	for i := 0; i < d.Concurrency; i++ {
 		go func() {
-			defer wg.Done()
-
 			for {
-				table, ok := <-tablesQueue
+				batch, ok := <-batchQueue
 				if !ok {
 					break
 				}
 
-				logger := d.logger.WithField("table", table.String())
+				batchLogger := d.logger.WithFields(logrus.Fields{
+					"table":              batch.table.String(),
+					"startPaginationKey": batch.paginationKeys.MinPaginationKey,
+					"endPaginationKey":   batch.paginationKeys.MaxPaginationKey,
+					"batchID":            batch.BatchID(),
+				})
 
-				targetPaginationKeyInterface, found := d.targetPaginationKeys.Load(table.String())
-				if !found {
-					err := fmt.Errorf("%s not found in targetPaginationKeys, this is likely a programmer error", table.String())
-					logger.WithError(err).Error("this is definitely a bug")
-					d.ErrorHandler.Fatal("data_iterator", err)
-					return
-				}
-
-				startPaginationKey := d.StateTracker.LastSuccessfulPaginationKey(table.String())
-				if startPaginationKey == math.MaxUint64 {
-					err := fmt.Errorf("%v has been marked as completed but a table iterator has been spawned, this is likely a programmer error which resulted in the inconsistent starting state", table.String())
-					logger.WithError(err).Error("this is definitely a bug")
-					d.ErrorHandler.Fatal("data_iterator", err)
-					return
-				}
-
-				cursor := d.CursorConfig.NewCursor(table, startPaginationKey, targetPaginationKeyInterface.(uint64))
+				batchLogger.Debug("setting up new batch cursor")
+				cursor := d.CursorConfig.NewCursor(batch.table, batch.paginationKeys.MinPaginationKey, batch.paginationKeys.MaxPaginationKey)
 				if d.SelectFingerprint {
 					if len(cursor.ColumnsToSelect) == 0 {
 						cursor.ColumnsToSelect = []string{"*"}
 					}
 
-					cursor.ColumnsToSelect = append(cursor.ColumnsToSelect, table.RowMd5Query())
+					cursor.ColumnsToSelect = append(cursor.ColumnsToSelect, batch.table.RowMd5Query())
 				}
 
 				err := cursor.Each(func(batch *RowBatch) error {
+					batchValues := batch.Values()
+					paginationKeyIndex := batch.PaginationKeyIndex()
+
+					batch.batchID = batch.BatchID()
+
+					batchLogger.WithField("size", batch.Size()).Debug("row event")
 					metrics.Count("RowEvent", int64(batch.Size()), []MetricTag{
-						MetricTag{"table", table.Name},
+						MetricTag{"table", batch.table.Name},
 						MetricTag{"source", "table"},
 					}, 1.0)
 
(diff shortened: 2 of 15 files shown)


tiwilliam:
Demonstration of the problem we are trying to solve, example below is using 12 threads and transfers a database with 30 tables that are vastly different in size.

Screenshot 2020-06-27 at 11 21 02

tiwilliam:
After bouncing the idea some more we brainstormed the challenge with state and resume using this chunk method. A solution to this could be to change the chunking to happen outside DataIterator and be a fixed partition map, saved to state as almost "mini tables".

data_iterator.go
```diff
@@ -107,81 +127,133 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
 
-	for table, _ := range tablesWithData {
-		tablesQueue <- table
-		i++
-		if i%loggingIncrement == 0 {
-			d.logger.WithField("table", table.String()).Infof("queued table for processing (%d/%d)", i, len(tablesWithData))
+				d.StateTracker.RegisterBatch(tableName, paginationKey.batchIndex, paginationKey.startPaginationKey, paginationKey.endPaginationKey)
+				batchQueue <- paginationKey
+
+				// Protect against uint64 overflow, this might happen if the table is full
```
shuhaowu, line 231: Not sure if I quite follow this. I think I can figure it out with some effort, but maybe we can just illustrate this via an example in the comment as opposed to making the reader figure it out?

data_iterator.go
```diff
@@ -1,14 +1,26 @@
...
+func (e GracefulShutdownError) Error() string {
+	return "graceful shutdown"
+}
+
+type PaginationKey struct {
+	table              *TableSchema
+	batchIndex         uint64
+	startPaginationKey uint64
+	endPaginationKey   uint64
+}
```
shuhaowu, line 22: This struct is confusingly named for me. Can we just use the `MinMaxKeys` struct already defined (possibly extending that with the `table` variable here), especially since we initialize variables of this type with the MinMaxKey data?

shuhaowu (commented):
I went through this as a first pass. I have a few questions, especially relating to `batchIndex` and how we apparently need a graceful shutdown of the data iterator.

Makefile
```diff
@@ -60,7 +60,7 @@ test-go:
 		curl -sL $(GOTESTSUM_URL) | tar -xz -C ./bin gotestsum; \
 	fi
 
-	ulimit -n 1024 && ./bin/gotestsum --format short-verbose ./test/go ./copydb/test ./sharding/test -count 1 -p 1 -failfast
+	ulimit -n 1024 && ./bin/gotestsum --format short-verbose ./test/go ./copydb/test ./sharding/test -count 1 -p 1
```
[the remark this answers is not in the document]
tiwilliam: The flag will make tests fail as soon as it hits a failure, making it hard to get an overview of how broken things are at current state.

ferry.go
```diff
@@ -596,7 +595,7 @@ func (f *Ferry) Run() {
 			if ctx.Err() == nil {
 				// Ghostferry is still running
 				if f.OverallState.Load() != StateCutover {
-					// Save state dump and exit if not during the cutover stage
+					f.DataIterator.WaitForGracefulShutdown()
```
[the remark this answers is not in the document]
tiwilliam: The graceful shutdown of iterators was introduced to guarantee that the shutdown state dump contain all copied rows. Most tests also rely on shutting down the process after one data iteration, which will trigger when one goroutine is finished, making it difficult to assert states when another goroutine is half-way.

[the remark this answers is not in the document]
shuhaowu: Are these golang tests? I think in the ruby tests we may be able to block the data iterator.

Also: we might want to introduce ways to force per-table-parallelism to be 1, not only for tests, but perhaps also for actual runs. This might help with some sort of phased rollout of this code.

table_schema_cache.go
```diff
@@ -116,14 +121,14 @@ func QuotedTableNameFromString(database, table string) string {
...
-	tablesWithData := make(map[*TableSchema]uint64)
+func MinMaxPaginationKeys(db *sql.DB, tables []*TableSchema, logger *logrus.Entry) (map[*TableSchema]*MinMaxKeys, []*TableSchema, error) {
+	tablesWithData := make(map[*TableSchema]*MinMaxKeys)
 	emptyTables := make([]*TableSchema, 0, len(tables))
 
 	for _, table := range tables {
 		logger := logger.WithField("table", table.String())
 
-		maxPaginationKey, maxPaginationKeyExists, err := maxPaginationKey(db, table)
+		maxPaginationKey, maxPaginationKeyExists, err := getPaginationKeyValue(db, table, false)
```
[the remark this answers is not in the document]
tiwilliam: We probably could, but I didn't deem it worth it for performance nor code clarity.

data_iterator.go
tiwilliam: Agree. Should find a new name or way to store this.

ferry.go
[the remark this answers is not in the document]
tiwilliam: When thinking about this again, I think we should get rid of the graceful shutdown and let rows be copied again on resume. And as you say, fix tests and code to allow parallelism to be turned off easily.

data_iterator.go
```diff
@@ -107,81 +130,135 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
+					batchEndPaginationKey = tableEndPaginationKey
+				}
+
+				paginationKey := &PaginationKey{
+					table:              key.table,
+					startPaginationKey: batchStartPaginationKey,
+					endPaginationKey:   batchEndPaginationKey,
+				}
+
+				d.StateTracker.RegisterBatch(tableName, paginationKey.BatchId(), paginationKey.startPaginationKey, paginationKey.endPaginationKey)
```
fjordan, line 228: would be great to see a log after the batch iteration regarding how many batches are created per-table

state_tracker.go
```diff
@@ -34,12 +35,15 @@ type SerializableState struct {
...
-	CompletedTables                           map[string]bool
+	CompletedTables map[string]bool
+	BatchProgress   map[string]map[uint64]*BatchProgress
+
 	LastWrittenBinlogPosition                 mysql.Position
 	BinlogVerifyStore                         BinlogVerifySerializedStore
 	LastStoredBinlogPositionForInlineVerifier mysql.Position
 	LastStoredBinlogPositionForTargetVerifier mysql.Position
+
+	estimatedPaginationKeysPerSecond float64
```
fjordan, line 46: Do we know this is correct with the batching?

test/integration/interrupt_resume_test.rb
```diff
@@ -54,12 +52,8 @@ def test_interrupt_resume_with_writes_to_source
 
     start_datawriter_with_ghostferry(datawriter, ghostferry)
 
-    batches_written = 0
     ghostferry.on_status(Ghostferry::Status::AFTER_ROW_COPY) do
-      batches_written += 1
-      if batches_written >= 2
-        ghostferry.send_signal("TERM")
-      end
```
fjordan, line 62: why were all of these removed?

data_iterator.go
```diff
@@ -107,81 +130,135 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
+			}
+		} else {
+			// Set start to minus one since cursor is searching for greater values
+			tableStartPaginationKey := key.startPaginationKey - 1
+			tableEndPaginationKey := key.endPaginationKey
+
+			// Number of batches are set to number of processes, unless each batch becomes smaller than the cursor size
+			keyInterval := tableEndPaginationKey - tableStartPaginationKey
+			concurrencyBatchSize := math.Ceil(float64(keyInterval) / float64(d.Concurrency))
+			batchSize := uint64(math.Max(concurrencyBatchSize, float64(d.CursorConfig.BatchSize)))
```
fjordan, line 205: Would like to see some tests for this

fjordan (commented):
Thanks for prototyping this.

Overall, I really like the approach and think it will really have a massive impact on the speed at which we're able to perform migrations, especially with large tables. 

However, this looks like a really rough draft and needs some cleanup and more tests before I think it's ready. I also don't think we need the graceful shutdown as discussed earlier.

tiwilliam:
Thanks for your 👀 @fjordan. I totally agree, it is rough currently. It's also a bit large to just merge as is even after fixing review comments. But if we like the approach I think this would be doable to tidy up in a couple of PRs. My approach to get this concept into master would be to break out what is possible to break out and launch this in a couple of steps. Maybe as a separate iterator we can switch on and off.

data_iterator.go
```diff
@@ -45,50 +62,50 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
 		if d.StateTracker.IsTableComplete(tableName) {
 			// In a previous run, the table may have been completed.
 			// We don't need to reiterate those tables as it has already been done.
-			delete(tablesWithData, table)
-		} else {
-			d.targetPaginationKeys.Store(table.String(), maxPaginationKey)
+			d.logger.WithField("table", tableName).Warn("table already completed, skipping")
+			continue
+		}
+		d.paginationKeys[tableName] = &PaginationKey{
```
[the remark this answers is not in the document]
tiwilliam: Good feedback, thanks. This is a rest from the past and I think as you say would be quite easy to get rid of and do when calculating batches instead.

state_tracker.go
```diff
@@ -137,31 +176,93 @@ func (s *StateTracker) UpdateLastResumableBinlogPositionForTargetVerifier(pos my
...
+	s.updateSpeedLog(newLatestPaginationKey - batch.LatestPaginationKey)
+
+	// Set latest key to EndPaginationKey if we are ahead due to a fragmented PK
+	batch.LatestPaginationKey = newLatestPaginationKey
+
+	if batch.EndPaginationKey == batch.LatestPaginationKey {
+		logger.Debug("marking batch as completed")
+		batch.Completed = true
+		batch.completedPercentage = 100
+		s.CopyRWMutex.Unlock()
```
floriecai, line 222: Can we move the `s.CopyRWMutext.Unlock` in both the `if` and `else` case and use `defer` instead to keep it consistent and in case of dangling locks?

state_tracker.go
```diff
@@ -137,31 +176,93 @@ func (s *StateTracker) UpdateLastResumableBinlogPositionForTargetVerifier(pos my
...
 
-	s.updateSpeedLog(deltaPaginationKey)
+	return batch.Completed
 }
 
-func (s *StateTracker) LastSuccessfulPaginationKey(table string) uint64 {
-	s.CopyRWMutex.RLock()
-	defer s.CopyRWMutex.RUnlock()
+func (s *StateTracker) areAllBatchesCompleted(table string) bool {
+	s.CopyRWMutex.Lock()
```
floriecai, line 251: why the change to `Lock`? Aren't we only reading in this method?

table_schema_cache.go
```diff
@@ -24,6 +24,11 @@ type TableIdentifier struct {
 	TableName  string
 }
 
+type MinMaxKeys struct {
```
floriecai, line 27: This looks almost the same as `PaginationKey` - could we collapse these 2 structs?

floriecai (commented):
In the description, could you also mention that the performance gains would be less if the keys are evenly distributed in these batches. As you mentioned in another comment, these batches are essentially "mini tables" (i.e. face the same issues of some go-routines finishing earlier and idling). It might be easier for people coming across this PR to understand.

state_tracker.go
tiwilliam: You are correct, it wrote some commits ago but forgot to change it back. Fixing.

data_iterator.go
```diff
@@ -107,81 +130,135 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
 						d.ErrorHandler.Fatal("inline_verifier", err)
 					default:
-						logger.WithError(err).Error("failed to iterate table")
+						d.logger.WithError(err).Error("failed to iterate table")
 						d.ErrorHandler.Fatal("data_iterator", err)
 					}
 
 				}
-
-				logger.Debug("table iteration completed")
```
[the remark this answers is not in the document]
tiwilliam: We kind of log this already from `UpdateBatchPosition` in `StateTracker` but I will add another row here, it's cheap.

state_tracker.go
tiwilliam: I have indeed updated `EstimatedPaginationKeysPerSecond` but please review this part an extra time.

data_iterator.go
tiwilliam: Do you mean more verbose than https://github.com/Shopify/ghostferry/blame/batch-data-iterator/data_iterator.go#L191?

data_iterator.go
tiwilliam: I actually went and reverted all test changes from this branch for now. Will go back and figure out how to test this properly since it need some changes to how we assert and halt on iteration loops etc.

state_tracker.go
```diff
@@ -137,31 +176,93 @@ func (s *StateTracker) UpdateLastResumableBinlogPositionForTargetVerifier(pos my
...
+		return
+	}
+
+	newLatestPaginationKey := uint64(math.Min(float64(latestPaginationKey), float64(batch.EndPaginationKey)))
+	s.updateSpeedLog(newLatestPaginationKey - batch.LatestPaginationKey)
+
+	// Set latest key to EndPaginationKey if we are ahead due to a fragmented PK
+	batch.LatestPaginationKey = newLatestPaginationKey
+
+	if batch.EndPaginationKey == batch.LatestPaginationKey {
```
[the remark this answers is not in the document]
tiwilliam: It's not a valid state but doesn't hurt to change this to `>=`. Will update.

data_iterator.go
```diff
@@ -107,81 +130,135 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
+					endPaginationKey:   batch.EndPaginationKey,
+				}
+			}
+		} else {
+			// Set start to minus one since cursor is searching for greater values
+			tableStartPaginationKey := key.startPaginationKey - 1
+			tableEndPaginationKey := key.endPaginationKey
+
+			// Number of batches are set to number of processes, unless each batch becomes smaller than the cursor size
+			keyInterval := tableEndPaginationKey - tableStartPaginationKey
```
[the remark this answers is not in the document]
tiwilliam: I was refraining from use `num` since it's not always true due to fragmentation.

data_iterator.go
```diff
@@ -107,81 +130,135 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
+				"batchSize":          batchSize,
+				"endPaginationKey":   tableEndPaginationKey,
+				"startPaginationKey": tableStartPaginationKey,
+			}).Debugf("queueing %d batches", (keyInterval/batchSize)+1)
+
+			for batchStartPaginationKey := tableStartPaginationKey; batchStartPaginationKey < tableEndPaginationKey; batchStartPaginationKey += batchSize {
+				batchEndPaginationKey := batchStartPaginationKey + batchSize
+
+				// Set batchEndPaginationKey to endPaginationKey if out of bounds
+				if batchEndPaginationKey > tableEndPaginationKey {
```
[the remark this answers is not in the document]
tiwilliam: `batchEndPaginationKey` is naive and set to `+ batchSize` and this clause is how we correct to proper batch ends. So this is not really an error, it's rather how we calculate the span we want to iterate. Will update the comment to make this clearer.

data_iterator.go
```diff
@@ -1,14 +1,29 @@
...
 )
 
+type GracefulShutdownError struct{}
+
+func (e GracefulShutdownError) Error() string {
+	return "graceful shutdown"
+}
+
+type PaginationKey struct {
+	table              *TableSchema
```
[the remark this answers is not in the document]
tiwilliam: The TableSchema is indeed used for setting up `NewCursor` and `SelectFingerprint` in the goroutine.

table_schema_cache.go
tiwilliam: Agree, fixed.

state_tracker.go
tiwilliam: We can't since we want to release and read in this method. But I can remove one of them and add `defer` to address your dangling point.

state_tracker.go
```diff
@@ -137,31 +176,93 @@ func (s *StateTracker) UpdateLastResumableBinlogPositionForTargetVerifier(pos my
...
+	if !batchFound {
+		logger.Error("tried to mark non-existing batch as completed")
+		return
+	}
+
+	newLatestPaginationKey := uint64(math.Min(float64(latestPaginationKey), float64(batch.EndPaginationKey)))
+	s.updateSpeedLog(newLatestPaginationKey - batch.LatestPaginationKey)
+
+	// Set latest key to EndPaginationKey if we are ahead due to a fragmented PK
+	batch.LatestPaginationKey = newLatestPaginationKey
```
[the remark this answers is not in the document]
tiwilliam: I do not follow here. This is running when batch writer succeeds, we set this to the reported value from the batch writer if it's larger then the previously reported value. The struct for holding these values are per table and batch, so no concurrency is happening at batch writer here.

data_iterator.go
```diff
@@ -107,81 +130,135 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
+
+				batchQueue <- &PaginationKey{
+					table:              key.table,
+					startPaginationKey: batch.LatestPaginationKey,
+					endPaginationKey:   batch.EndPaginationKey,
+				}
+			}
+		} else {
+			// Set start to minus one since cursor is searching for greater values
+			tableStartPaginationKey := key.startPaginationKey - 1
```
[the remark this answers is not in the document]
tiwilliam: Sure, here: https://github.com/Shopify/ghostferry/blob/master/cursor.go#L288

data_iterator.go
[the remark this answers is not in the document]
tiwilliam: Fixed in 1ab058ac44b49615142e91751f9159d188c7f64f. Got rid of `d.paginationKeys` all together and merged the two into one.

tiwilliam:
I've now addressed most feedback and reduced this PR to the minimum change set. I think this is now a greater prototype we can use to figure out how to break into smaller releases.

sharding/test/trivial_integration_test.go
```diff
@@ -106,7 +106,7 @@ func TestSelectiveCopyDataWithInsertLoadOnAllTenants(t *testing.T) {
 	assert.Equal(t, 0, count)
 
 	rows := testcase.AssertQueriesHaveEqualResult("SELECT * FROM gftest.table1 WHERE tenant_id = 2")
-	assert.True(t, len(rows) > 33)
+	assert.True(t, len(rows) >= 33)
```
tiwilliam, line 109: Flaky test.

data_iterator.go
```diff
@@ -45,140 +50,198 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 		d.StateTracker.MarkTableAsCompleted(table.String())
 	}
 
-	for table, maxPaginationKey := range tablesWithData {
+	batchQueue := make(chan *DataIteratorBatch)
```
coltiebaby, line 53: `batchQueue := make(chan *DataIteratorBatch, d.Concurrency)`

Your for loop is going to get blocked a few times if you don't.

tiwilliam:
Closing this since the PoC point is made and implementation must be incremental if decided on. Branch and PR will be around for history.

Issue #200, opened 2020-07-22
Ghostferry Idempotence


fjordan:
While there is some form of implicit assurance (in the form of `INSERT IGNORE` and the added predicates for all generated SQL DMLs) that Ghostferry is idempotent in nature, there is no explicit or formal validation either in tests or the the TLA+ specification. 

Explicitly verifying this feature of Ghostferry's algorithm is something that should be pursued as it would allow users of Ghostferry to replay moves from positions in the TableIterator and/or BinlogWriter that may have previously been applied. 

For example, if a run completes to position _p2_ and is then interrupted, but it then resumed from point _p1_, or even _p0_, it must be proven that re-running ghostferry from these previous positions will not corrupt any data and that source/target equality will be successfully achieved. 

![image](https://user-images.githubusercontent.com/390811/88227683-8f6e7280-cc33-11ea-86d3-e2dfea204c65.png)

Following this verification, it may be very helpful in the case that hundreds, or thousands, or more Ghostferry processes are taking place and would like to support interruption. Instead of requiring each of these moves to capture their state at this exact point-in-time at interrupt, we could then periodically persist the state of the ghostferry run to be resumed in the event of interruption and bypass a potential thundering herd of writes.

/cc @Shopify/pods

milanatshopify:
Being able to periodically save the (future resume) state would also simplify dealing with different interrupt scenarios, as well as cover the "the k8s pod just died" situations we frequently encounter.

fjordan:
Closed by https://github.com/Shopify/ghostferry/pull/207

PR #201, opened 2020-07-22
Ghostferry idempotence

fjordan:
## Overview
This PR adds both integration tests and TLA+ spec modifications that explicitly verify the idempotence of ghostferry's algorithm.

### Integration Tests
Several tests were added to the ruby integration test suite that perform combinations of interrupting, resuming, and interrupting or failing again before running finally to completion. These tests utilize assertions that the source and target are equal and some assertions that inspect the logs to ensure it indeed ran the intended number of times.

### TLA+ Spec
Additions to the TLA+ specification (via Pluscal) were added and checked with TLC to verify the correctness of these changes. 

To model the idempotence of the `TableIterator`, an `either...or` clause in the `tblit_upkey` label was added that may or may not increment the `PK` of the `TableIterator` as it is iterating the data of the source table. The effect of this is that the current row may be applied more than one time as the data is being copied.

Likewise, a `label` was added to the `BinlogWriter` (`reapply_binlog_write`) that duplicates the first entry to the `BinlogWriteQueue`. The effect of this is that any binlog entry may be applied more than once, but while also preserving the order of entries in this queue.

Following the other simplifications and proofs by induction in the spec, we make the simplification that replaying the current row in the `TableIterator` and current entry in the `BinlogStreamer` is the same as replaying the last _N_ rows or operations (in order).

Also note that while making these changes, I was able to violate the `SourceTargetEquality` condition in a TLC run by copying the current first element of the `BinlogWriteQueue` to the _end_ of the queue instead of placing it again at the _beginning_. This proves further that, as we have modeled in the implementation, the ordering of these events must remain the same and that we cannot apply them arbitrarily, and as long as they are applied in order, idempotence does indeed hold. 

### Why?
See https://github.com/Shopify/ghostferry/issues/200 for details.

/cc @Shopify/pods

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -329,7 +329,6 @@ func (s *BinlogStreamer) updateLastStreamedPosAndTime(evTimestamp uint32, evPos 
 }
 
 func (s *BinlogStreamer) handleRowsEvent(ev *replication.BinlogEvent, query []byte) error {
-	eventTime := time.Unix(int64(ev.Header.Timestamp), 0)
 	rowsEvent := ev.Event.(*replication.RowsEvent)
 
 	if ev.Header.LogPos == 0 {
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -369,11 +369,6 @@ func (s *BinlogStreamer) handleRowsEvent(ev *replication.BinlogEvent, query []by
 		}
 
 		events = append(events, dmlEv)
-		s.logger.WithFields(logrus.Fields{
-			"database": dmlEv.Database(),
-			"table":    dmlEv.Table(),
-			"position": pos,
-		}).Debugf("received event %T at %v", dmlEv, eventTime)
 
 		metrics.Count("RowEvent", 1, []MetricTag{
 			MetricTag{"table", dmlEv.Table()},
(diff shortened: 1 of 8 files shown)


binlog_streamer.go
```diff
@@ -369,11 +368,6 @@ func (s *BinlogStreamer) handleRowsEvent(ev *replication.BinlogEvent, query []by
 		}
 
 		events = append(events, dmlEv)
-		s.logger.WithFields(logrus.Fields{
```
fjordan, line 372: This was removed as it was spamming the logs and preventing the travis tests from completing. I don't think it adds much value

fjordan:
After discussions with @shuhaowu, I've decided to rework the TLA+ changes proposed in this PR to model the idempotence of the algorithm.

milanatshopify:
> After discussions with @shuhaowu, I've decided to rework the TLA+ changes proposed in this PR to model the idempotence of the algorithm.

Can we do "both"?  Or is that just a waste of computers?

shuhaowu:
We're still going to keep the tests here right? They looked decent as far as I remember.

fjordan:
> Can we do "both"? Or is that just a waste of computers?

We want to make sure we don't move the spec further away from the actual implementation, which is what the additions here were unfortunately doing. We could keep them, because yes, they do model idempotence in one way. I think we'd prefer the alternative with something like multiple initial starting states.

> We're still going to keep the tests here right? They looked decent as far as I remember.

Yes. I'll break them out into a separate PR.

PR #205, opened 2020-09-02
Add periodic state callback

fjordan:
Part of https://github.com/Shopify/ghostferry/issues/200

This PR adds a periodic callback of the run's state, very similar to how the progress callback is performed. Tests were added that both verify the callback is performed correctly and also that the state sent in the callback can be used to successfully resume an interrupted/failed move.

While we're still yet to complete formal verification of Ghostferry's idempotence with TLA+, we have verified the correctness with [integration tests](https://github.com/Shopify/ghostferry/pull/202), and have enough confidence to move forward with this PR.

[This branch](https://github.com/Shopify/ghostferry/tree/idempotence-tla-shuhao) contains the latest TLA+ proving the DataIterator's idempotence (the BinlogStreamer is next). We're targeting to wrap up the TLA+ changes in the next couple of weeks.

/cc @Shopify/pods

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -912,14 +912,13 @@ func (f *Ferry) ReportState() {
 	callback := f.Config.StateCallback
 	state, err := f.SerializeStateToJSON()
 	if err != nil {
-		f.logger.WithError(err).Error("failed to dump state to JSON")
-		return
+		f.logger.Panicf("failed to serialize state to JSON: %s", err)
 	}
 
 	callback.Payload = string(state)
 	err = callback.Post(&http.Client{})
 	if err != nil {
-		f.logger.WithError(err).Errorf("failed to post state to callback: %s", callback)
+		f.logger.Panicf("failed to post state to callback: %s with err: %s", callback, err)
 	}
 }
 
--- a/ferry.go
+++ b/ferry.go
@@ -587,6 +587,24 @@ func (f *Ferry) Run() {
 		}()
 	}
 
+	if f.Config.StateCallback.URI != "" {
+		supportingServicesWg.Add(1)
+		go func() {
+			defer supportingServicesWg.Done()
+
+			frequency := time.Duration(f.Config.StateReportFrequency) * time.Millisecond
+
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(frequency):
+					f.ReportState()
+				}
+			}
+		}()
+	}
+
 	if f.DumpStateOnSignal {
 		go func() {
 			c := make(chan os.Signal, 1)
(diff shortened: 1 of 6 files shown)


ferry.go
```diff
@@ -890,6 +908,21 @@ func (f *Ferry) ReportProgress() {
 	}
 }
 
+func (f *Ferry) ReportState() {
+	callback := f.Config.StateCallback
+	state, err := f.SerializeStateToJSON()
+	if err != nil {
+		f.logger.WithError(err).Error("failed to dump state to JSON")
```
fjordan, line 915: I made this an `Error` instead of `Warn` given the difference from simply the progress and how it may be used.

ferry.go
```diff
@@ -890,6 +908,21 @@ func (f *Ferry) ReportProgress() {
...
+		return
+	}
+
+	callback.Payload = string(state)
+	err = callback.Post(&http.Client{})
+	if err != nil {
+		f.logger.WithError(err).Errorf("failed to post state to callback: %s", callback)
```
fjordan, line 922: I made this an `Error` instead of `Warn` (like above) considering the importance and how it will be used versus simply a progress report.

ferry.go
```diff
@@ -890,6 +908,21 @@ func (f *Ferry) ReportProgress() {
 	}
 }
 
+func (f *Ferry) ReportState() {
+	callback := f.Config.StateCallback
+	state, err := f.SerializeStateToJSON()
```
shuhaowu, line 913: I think this is fine, but maybe it's worth mentioning that this method will lock the StateTracker which will in turn block the BinlogWriter and BatchWriter from proceeding. I'm pretty sure this method is quite fast, but just as a FYI.

ferry.go
```diff
@@ -890,6 +908,21 @@ func (f *Ferry) ReportProgress() {
...
+	if err != nil {
+		f.logger.WithError(err).Error("failed to dump state to JSON")
+		return
```
shuhaowu, line 916: Should this panic here via the ErrorHandler?

ferry.go
shuhaowu: I don't think the `.Error` method actually does anything other than emit to stderr. Maybe you want to panic via the `ErrorHandler`?

test/integration/callbacks_test.rb
```diff
@@ -33,4 +33,19 @@ def test_progress_callback
...
+    ghostferry = new_ghostferry(MINIMAL_GHOSTFERRY, config: { verifier_type: "Inline" })
+    state = []
+    ghostferry.on_callback("state") do |state_data|
+      state << state_data
+    end
+
+    ghostferry.run
+
+    assert state.length >= 1
+    assert_basic_fields_exist_in_dumped_state(state.last)
```
shuhaowu, line 49: Maybe assert all of them?

```
states.each do |state|
assert..
```

shuhaowu (approved):
Generally looks good, a few minor comments.

ferry.go
fjordan: Good catch. I can add that note. Would you rather see it here or in the config where the interval is configured?

shuhaowu: Maybe here is fine. I don't if/how that lock will actually impact the run.

ferry.go
fjordan: Hmm, I think no but don't feel strongly and could be convinced otherwise. 

This would _not_ prevent the run from completing, and worst case the run would need to restart from the beginning. 

What do you think?

ferry.go
fjordan: Same comment below. I did want to raise it in the logs as an error instead of just a warning. If we do want this to panic and fail the run we can, although it isn't really necessary. Again, if you feel differently I could be convinced.

ferry.go
shuhaowu: Failing to dump the JSON seems like a weird condition that we would probably panic?

Issue #194, opened 2020-06-24
Ghostferry should queue tables in the DataIterator based on estimated table size


shuhaowu:
Since Ghostferry's data copy process is parallelized by using one goroutine per table, it may eventually be bottle-necked by the largest table. 

A simple optimization is to queue the table that's estimated to be the biggest one first. For most situations, we can get an estimate via the TargetPK - StartingPK.

PR #206, opened 2020-09-10
Run DataIterator over tables in descending order of pagination key

floriecai:
Closes https://github.com/Shopify/ghostferry/issues/194

Estimate the size of the table to be copied by counting the number of pagination keys of that table. Start copying the largest tables first in order to avoid large table from being the bottleneck near the end of copying tables.

### DIFF
--- a/data_iterator.go
+++ b/data_iterator.go
@@ -5,6 +5,7 @@ import (
 	sql "github.com/Shopify/ghostferry/sqlwrapper"
 	"math"
 	"sync"
+	"sort"
 
 	"github.com/sirupsen/logrus"
 )
@@ -52,7 +53,7 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 			// We don't need to reiterate those tables as it has already been done.
 			delete(tablesWithData, table)
 		} else {
-			d.targetPaginationKeys.Store(table.String(), maxPaginationKey)
+			d.targetPaginationKeys.Store(tableName, maxPaginationKey)
 		}
 	}
 
@@ -165,11 +166,25 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 		loggingIncrement = 1
 	}
 
-	for table, _ := range tablesWithData {
-		tablesQueue <- table
+	type kv struct {
+		Table *TableSchema
+		MaxPaginationKey uint64
+	}
+
+	sortedTableData := make([]kv, 0, len(tablesWithData))
+	for k, v := range tablesWithData {
+		sortedTableData = append(sortedTableData, kv{k, v})
+	}
+
+	sort.Slice(sortedTableData, func(i, j int) bool {
+		return sortedTableData[i].MaxPaginationKey > sortedTableData[j].MaxPaginationKey
+	})
+
+	for _, tableData := range sortedTableData {
+		tablesQueue <- tableData.Table
 		i++
 		if i%loggingIncrement == 0 {
-			d.logger.WithField("table", table.String()).Infof("queued table for processing (%d/%d)", i, len(tablesWithData))
+			d.logger.WithField("table", tableData.Table.String()).Infof("queued table for processing (%d/%d)", i, len(sortedTableData))
 		}
 	}
 
--- a/vendor/github.com/sirupsen/logrus/go.mod
+++ b/vendor/github.com/sirupsen/logrus/go.mod
@@ -1,5 +1,7 @@
 module github.com/sirupsen/logrus
 
+go 1.14
+
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/konsorten/go-windows-terminal-sequences v1.0.1


vendor/github.com/sirupsen/logrus/go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/sirupsen/logrus
 
+go 1.14
```
shuhaowu, line 3: Any reason why this is changed?

data_iterator.go
```diff
@@ -165,11 +166,25 @@ func (d *DataIterator) Run(tables []*TableSchema) {
...
+		MaxPaginationKey uint64
+	}
+
+	sortedTableData := make([]kv, 0, len(tablesWithData))
+	for k, v := range tablesWithData {
+		sortedTableData = append(sortedTableData, kv{k, v})
+	}
+
+	sort.Slice(sortedTableData, func(i, j int) bool {
+		return sortedTableData[i].MaxPaginationKey > sortedTableData[j].MaxPaginationKey
```
shuhaowu, line 180: This is fine for now, but I'm wondering if we can get a bit more accurate via `information_schema`. Specifically, we can estimate a table size via something like:

```
SELECT 
    table_name AS `Table`, 
    round(((data_length + index_length) / 1024 / 1024), 2) `Size in MB` 
FROM information_schema.TABLES 
WHERE table_schema = "$DB_NAME"
    AND table_name = "$TABLE_NAME";
```

In our case, I think we only need to know about the `data_length` instead of the `data_length + index_length`. However, this could still be a relatively inaccurate estimate if one uses ghostferry-sharding to select a subset of data.

Ultimately, I think a good idea would be to have an interface in Ghostferry that allows application to implement arbitrary table sorting algorithms, so the code could look like something like this:

```
// bad name, but just an example
// Alternatively, this method could be in the TableFilter struct 
type TableSorterForDataIterator interface { 
  Sort(tables []*TableSchema) []*TableSchema // Maybe a better name here..
}
```

A custom table sorter implementing that interface is passed to Ferry then passed to DataIterator, which is then used here. We can implement a few simple algorithm within Ghostferry (based on both information_schema and naively via max_primary_key) so people can choose from that if they don't want to implement their own.

This would be in line with how we do things with Filters, Thottlers, and Verifier.

shuhaowu (approved):

vendor/github.com/sirupsen/logrus/go.mod
floriecai: No idea why this is even picking up as a change since this was changed in April. 

image

data_iterator.go
```diff
@@ -52,7 +59,7 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 			// We don't need to reiterate those tables as it has already been done.
 			delete(tablesWithData, table)
 		} else {
-			d.targetPaginationKeys.Store(table.String(), maxPaginationKey)
+			d.targetPaginationKeys.Store(tableName, maxPaginationKey)
```
fjordan, line 62: 👍

data_iterator.go
```diff
@@ -189,3 +204,62 @@ func (d *DataIterator) AddBatchListener(listener func(*RowBatch) error) {
 func (d *DataIterator) AddDoneListener(listener func() error) {
 	d.doneListeners = append(d.doneListeners, listener)
 }
+
+func GetOrderedTables(unorderedTables map[*TableSchema]uint64, db *sql.DB) ([]TableKeyInfo, error) {
```
fjordan, line 208: probably worth adding a test for all of this?

data_iterator.go
```diff
@@ -189,3 +204,62 @@ func (d *DataIterator) AddBatchListener(listener func(*RowBatch) error) {
...
+
+	databaseSchemas := make([]string, len(databaseSchemasMap))
+
+	i = 0
+	for schema := range databaseSchemasMap {
+		databaseSchemas[i] = schema
+	}
+
+	query := fmt.Sprintf("SELECT TABLE_NAME, TABLE_SCHEMA, DATA_LENGTH FROM information_schema.tables WHERE TABLE_SCHEMA IN (\"%s\") AND TABLE_NAME IN (\"%s\") ORDER BY DATA_LENGTH DESC", strings.Join(databaseSchemas, `", "`), strings.Join(tableNames, `", "`))
+	rows, err := db.Query(query)
```
shuhaowu, line 235: Alternatively, to make this code simpler (at the code of performance), you can just issue one of these SELECT queries per table in the list?

data_iterator.go
```diff
@@ -159,17 +166,26 @@ func (d *DataIterator) Run(tables []*TableSchema) {
 		}()
 	}
 
-	i := 0
+	sorter := MaxTableSizeSorter{DataIterator: d}
+	sortedTableData, err := sorter.Sort(tablesWithData)
```
floriecai, line 170: I'm hardcoding the type of sorter we use for now and we can pass this in the config in another PR.

data_iterator_sorter.go
```diff
@@ -0,0 +1,92 @@
...
+}
+// MaxTableSizeSorter uses `information_schema.tables` to estimate the size of the DB and sorts tables in DESC order
+type MaxTableSizeSorter struct {
+	DataIterator *DataIterator
+}
+
+func (s *MaxTableSizeSorter) Sort(unorderedTables map[*TableSchema]uint64) ([]TableMaxPaginationKey, error) {
+	orderedTables := []TableMaxPaginationKey{}
+	tableNames := []string{}
+	databaseSchemasMap := map[string]int{}
```
shuhaowu, line 40: Minor nitpick, the typical ways of doing a set in golang is:

```suggestion
	databaseSchemasSet := map[string]struct{}{}
```

data_iterator_sorter.go
```diff
@@ -0,0 +1,92 @@
...
+	databaseSchemas := []string{}
+
+	for tableSchema, maxPK := range unorderedTables {
+		orderedTables = append(orderedTables, TableMaxPaginationKey{tableSchema, maxPK})
+		tableNames = append(tableNames, tableSchema.Name)
+
+		if _, exists := databaseSchemasMap[tableSchema.Schema]; !exists {
+			databaseSchemasMap[tableSchema.Schema] = 0
```
shuhaowu, line 48:
```suggestion
			databaseSchemasSet[tableSchema.Schema] = struct{}{}
```

shuhaowu (approved):
Very minor nitpick

PR #207, opened 2020-09-16
Ghostferry interrupt resume idempotence TLA+ modifications

shuhaowu:
Huge thanks to @fjordan for helping me with all of these. We tackled this problem via the multiple starting state feature of TLC as follows:

## TLA+ for DataIterator interrupt/resume

Instead of interrupting and then resuming in TLA as that would be messy and difficult to check, we generate starting states for the SourceTable and TargetTable such that it looks like it was previously interrupted. This is a multiple-starting state approach that significantly increases the search space of the TLC run.

We also generate a valid "LastSuccessfulPK" that will be valid given the table states such that some of the initial states will cause the Ghostferry algorithm to re-copy previously copied rows. There's no mathematical proof that as long as we can re-copy one row correctly, we can re-copy n rows correctly, but intuition says this is indeed the case.

## TLA+ for BinlogStreamer interrupt/resume

Taking a similar approach with the TLA+ for the DataIterator, the idea here is we generate a possible SourceBinlog sequence at the initial starting state.

This time, we generate a SourceBinlog that's either an empty sequence (simulating a fresh Ghostferry run) or contains a single binlog entry that has already been previously copied over to the TargetTable. The TLA+ specification is constructed such that when Ghostferry starts, it will attempt to copy that binlog over to the target again. Effectively, this simulates when Ghostferry is resumed from a binlog coordinate that is "outdated": i.e. if a binlog is already copied but the last successful binlog position is not updated, it is safe to resume from an outdated position.

Once again, this is only a single binlog entry and we do not have a mathematical proof that if it works with a single binlog entry, it works with n binlog entries. However, strong intuitions suggests that this is the case so this proof is omitted for now.

## Other notes

Checking this with TLC on My Ryzen 7 3700X machine took about 17min. On a 2020 MBP with a 6core/12thread processor it took about 38min. 

Most of the code changes are generated by TLA+ Toolbox. The actual changes is relatively small.

### DIFF
--- a/tlaplus/ghostferry.tla
+++ b/tlaplus/ghostferry.tla
@@ -349,7 +349,7 @@ PossibleSourceBinlogs == [{1} -> PossibleBinlogEntries]
       \* TODO: model this with TLA+ and validate its correctness.
       tblit_loop:  while (LastSuccessfulPK < MaxPrimaryKey) {
       tblit_rw:      currentRow := SourceTable[LastSuccessfulPK + 1];
-                     if (currentRow # NoRecordHere /\ TargetTable[LastSuccessfulPK + 1] = NoRecordHere) { \* TODO (NOW): This should fail because the second part of the database is not correctly constructed?
+                     if (currentRow # NoRecordHere /\ TargetTable[LastSuccessfulPK + 1] = NoRecordHere) {
                        TargetTable[LastSuccessfulPK + 1] := currentRow;
                      };
       tblit_upkey:   LastSuccessfulPK := LastSuccessfulPK + 1;
@@ -470,7 +470,7 @@ PossibleSourceBinlogs == [{1} -> PossibleBinlogEntries]
 }
 
  ***************************************************************************)
-\* BEGIN TRANSLATION - the hash of the PCal code: PCal-0fb2191b7ea3f2d9d2ddacddfb52ede1 (chksum(pcal) \in STRING /\ chksum(tla) \in STRING) (chksum(pcal) = "cad34be2" /\ chksum(tla) = "7a595238") (chksum(pcal) = "5f2f1443" /\ chksum(tla) = "7ebd37e2") (chksum(pcal) = "59302a14" /\ chksum(tla) = "7ebd37e2") (chksum(pcal) = "605b0be6" /\ chksum(tla) = "9b1e66de") (chksum(pcal) = "cccf8dab" /\ chksum(tla) = "fb6ea81b") (chksum(pcal) = "ddab7822" /\ chksum(tla) = "8c2fdca7") (chksum(pcal) = "dd200a77" /\ chksum(tla) = "108353c1") (chksum(pcal) = "dd200a77" /\ chksum(tla) = "5f47d505") PCal-88d6f264d1db8757d257b45fcc829f8a
+\* BEGIN TRANSLATION - the hash of the PCal code: PCal-88d6f264d1db8757d257b45fcc829f8a
 CONSTANT defaultInitValue
 VARIABLES MaxPrimaryKey, CurrentMaxPrimaryKey, LastSuccessfulPK, 
           ActualCopiedPK, SourceTable, TargetTable, SourceBinlog, 
@@ -781,5 +781,5 @@ BinlogSizeActionConstraint == Len(SourceBinlog) <= MaxBinlogSize
 
 =============================================================================
 \* Modification History
-\* Last modified Wed Sep 16 15:27:21 EDT 2020 by shuhao
+\* Last modified Wed Sep 16 16:07:45 EDT 2020 by shuhao
 \* Created Thu Jan 18 11:35:40 EST 2018 by shuhao
--- a/tlaplus/ghostferry.tla
+++ b/tlaplus/ghostferry.tla
@@ -33,7 +33,7 @@
  This whole process is modeled as five separate processes:
  1. TableIterator: Performs the job of the DataIterator, but only on one table
                    (see Data Model for justification).
- 2. BinlogStreamer: Tails and queues the binlog to be written by the 
+ 2. BinlogStreamer: Tails and queues the binlog to be written by the
                     BinlogWriter.
  3. BinlogWriter: Writes binlogs in binlog writer queue.
  4. Application: Perform INSERT/UPDATE/DELETE on the source database and record
@@ -81,9 +81,9 @@
       is changed.
     - To ensure the binlog does not grow to infinity and checking with TLC stays
       feasible, a maximum size is constrained on it.
- 5. The BinlogWriter writes one binlog at a time as opposed to writing them in 
+ 5. The BinlogWriter writes one binlog at a time as opposed to writing them in
     batches atomically like it does in the implementation via transactions.
-    - The reasoning is that writing one binlog event at a time is a "worse" 
+    - The reasoning is that writing one binlog event at a time is a "worse"
       case than writing more than one binlog event atomically at a time. As long
       as they are ordered correctly, it should be okay.
 
@@ -182,20 +182,6 @@ SetMin(S) == CHOOSE i \in S: \A j \in S : i <= j
 \* Constant Declarations
 \* =====================
 
-(***************************************************************************
-This defines the initial layout of the source table. The source table should
-be defined as a sequence with a bunch of Records. It is possible to be
-NoRecordHere as that is the placeholder for an non-existent row.
-
-Example: InitialTable = <<r0, r1, NoRecordHere>>
-         MaxPrimaryKey = 2
-
-The example table will also have a TableCapacity of 3 and a CurrentMaxPrimaryKey
-of 2. The CurrentMaxPrimaryKey can be increased when the third element is
-updated to an element \in Records.
- ***************************************************************************)
-CONSTANT InitialTable
-CONSTANT MaxPrimaryKey
 
 (***************************************************************************
 This defines the set of possible records that can be written to the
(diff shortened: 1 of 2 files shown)


tlaplus/ghostferry.tla
```diff
@@ -407,33 +469,47 @@ PossibleRecords == Records \cup {NoRecordHere}
 }
 
  ***************************************************************************)
-\* BEGIN TRANSLATION
+\* BEGIN TRANSLATION - the hash of the PCal code: PCal-0fb2191b7ea3f2d9d2ddacddfb52ede1 (chksum(pcal) \in STRING /\ chksum(tla) \in STRING) (chksum(pcal) = "cad34be2" /\ chksum(tla) = "7a595238") (chksum(pcal) = "5f2f1443" /\ chksum(tla) = "7ebd37e2") (chksum(pcal) = "59302a14" /\ chksum(tla) = "7ebd37e2") (chksum(pcal) = "605b0be6" /\ chksum(tla) = "9b1e66de") (chksum(pcal) = "cccf8dab" /\ chksum(tla) = "fb6ea81b") (chksum(pcal) = "ddab7822" /\ chksum(tla) = "8c2fdca7") (chksum(pcal) = "dd200a77" /\ chksum(tla) = "108353c1") (chksum(pcal) = "dd200a77" /\ chksum(tla) = "5f47d505")
```
shuhaowu, line 472: Everything below this line is automatically generated.

tlaplus/ghostferry.toolbox/ghostferry___GhostferryPrimaryModel.launch
```diff
@@ -1,60 +1,64 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
shuhaowu, line 1: This entire file was automatically updated as TLA+ toolbox was updated.

tlaplus/ghostferry.tla
```diff
@@ -284,12 +347,12 @@ PossibleRecords == Records \cup {NoRecordHere}
       \* It may be possible to perform some sort of locking via the Application,
       \* but this seems cumbersome and prone to implementation level error.
       \* TODO: model this with TLA+ and validate its correctness.
-      tblit_loop:  while (lastSuccessfulPK < MaxPrimaryKey) {
-      tblit_rw:      currentRow := SourceTable[lastSuccessfulPK + 1];
-                     if (currentRow # NoRecordHere) {
-                       TargetTable[lastSuccessfulPK + 1] := currentRow;
+      tblit_loop:  while (LastSuccessfulPK < MaxPrimaryKey) {
+      tblit_rw:      currentRow := SourceTable[LastSuccessfulPK + 1];
+                     if (currentRow # NoRecordHere /\ TargetTable[LastSuccessfulPK + 1] = NoRecordHere) {
```
shuhaowu, line 352: A lot of drama came with this line of change. A very detailed explanation is available at the commit message of: https://github.com/Shopify/ghostferry/pull/207/commits/1daf5fdd393f5bdd240e90ad11100e9148ad4767

xliang6: This took me a while to understand. I think this is only okay because we do `SELECT FOR UPDATE` and `REPLACE INTO` or `INSERT IGNORE` for the data copy as an atomic operation.

xliang6 (approved):
Thanks for completing the TLA+ verification for interrupt and resume as thorough and yet practical as possible! This has been more of a learning opportunity for me than a review, specially not knowing all od PlusCal syntax. I have a couple of questions/clarifications, but overall looks really good.

tlaplus/ghostferry.tla
```diff
@@ -224,25 +222,92 @@ NoRecordHere == CHOOSE r : r \notin Records
...
 PossibleRecords == Records \cup {NoRecordHere}
 
+\* These are all the possible states of the table, but they may not be valid.
+PossibleTableStates == [1..TableCapacity -> PossibleRecords]
+
+\* This is a set of all possible binlog structs in a partial form: it is missing the pk.
+PossibleOldNewRowStates == {s \in [{"oldr", "newr"} -> Records] : s.oldr # s.newr}
+
+\* This is an operator that adds the pk back into a partial binlog struct in PossibleOldNewRowStates and adds back the PK.
+ConstructBinlogFromOldNewRowStates(k, s) == [pk |-> k] @@ s
```
fjordan, line 232: nice use of the function merging operator!

tlaplus/ghostferry.tla
```diff
@@ -224,25 +222,92 @@ NoRecordHere == CHOOSE r : r \notin Records
...
+    \* We generate a possible SourceBinlog before Ghostferry starts running from two possibilities:
+    \* 
+    \* 1. There are no binlog entries yet. This is the "standard" Ghostferry start.
+    \* 2. There's a single binlog entry. This binlog entry has already been copied to the 
+    \*    target database during a previous Ghostferry run. This simulates a resume.
+    SourceBinlog \in (
+        {<<>>}
+        \cup
+        {binlogs \in PossibleSourceBinlogs : binlogs[1].newr = TargetTable[binlogs[1].pk]}
+    ),
```
fjordan, line 310: I like how simple this turned out 💯

tlaplus/ghostferry.tla
```diff
@@ -224,25 +222,92 @@ NoRecordHere == CHOOSE r : r \notin Records
...
+PossibleOldNewRowStates == {s \in [{"oldr", "newr"} -> Records] : s.oldr # s.newr}
+
+\* This is an operator that adds the pk back into a partial binlog struct in PossibleOldNewRowStates and adds back the PK.
+ConstructBinlogFromOldNewRowStates(k, s) == [pk |-> k] @@ s
+
+\* This generates a set of possible binlog entries with all possible PKs.
+PossibleBinlogEntries == UNION {{ConstructBinlogFromOldNewRowStates(k, s) : k \in PrimaryKeys} : s \in PossibleOldNewRowStates}
+
+\* This generates a set of possible SourceBinlog, which is a sequence.
+\* For simplicity and TLC feasibility, we only generate a single binlog entry in the initial possible SourceBinlog.
```
fjordan, line 238: Hmm, would it make sense to generate the number of entries up to the `LastSuccessfulPK ` instead of just a single entry? I do know we are simplifying 1 == n here and in other places, so maybe that doesn't make sense.

tlaplus/ghostferry.tla
```diff
@@ -407,33 +446,40 @@ PossibleRecords == Records \cup {NoRecordHere}
...
 
 Init == (* Global variables *)
+        /\ MaxPrimaryKey \in 1..TableCapacity
         /\ CurrentMaxPrimaryKey = MaxPrimaryKey
-        /\ SourceTable = InitialTable
-        /\ TargetTable = [k \in PrimaryKeys |-> NoRecordHere]
+        /\ LastSuccessfulPK \in 0..MaxPrimaryKey
+        /\ SourceTable \in                 {
+                               table \in PossibleTableStates : /\ \A k \in MaxPrimaryKey+1..TableCapacity : table[k] = NoRecordHere
+                                                               /\ ~(\A k \in 1..TableCapacity : table[k] = NoRecordHere)
```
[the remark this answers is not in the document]
shuhaowu: Yes that is correct.

tlaplus/ghostferry.tla
fjordan: I think your explanation above is correct. As @xliang6 stated above, this is predicated on the fact that the DataIterator operates atomically and is not subject to a race condition with the BinlogStreamer.

fjordan (approved):
This is great. Thanks @shuhaowu 👏

tlaplus/ghostferry.tla
```diff
@@ -407,33 +446,40 @@ PossibleRecords == Records \cup {NoRecordHere}
...
+                           }
+        /\ TargetTable \in                 {
+                               table \in PossibleTableStates : /\ \A k \in 1..LastSuccessfulPK: table[k] = SourceTable[k]
```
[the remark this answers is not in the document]
shuhaowu: I think you are looking at an outdated diff. Not sure how this came to be, but the actual clause in the most up to date spec is:
```
    TargetTable \in {
        table \in PossibleTableStates : /\ \A k \in 1..ActualCopiedPK: table[k] = SourceTable[k]
                                        /\ \A k \in (ActualCopiedPK+1)..TableCapacity : table[k] = NoRecordHere
    },
```

where `ActualCopiedPK \in LastSuccessfulPK..MaxPrimaryKey,`

So basically for the target table, from 1..LastSuccessfulPK, it will definitely contain entries from the source table. But from LastSuccessfulPK..TableCapacity, it _could_ have entries from the source table, which mean ghostferry will recopy it.

tlaplus/ghostferry.tla
shuhaowu: It would work I think, but it would explode the initial states. Currently already taking 20min to check on ryzen 3700x so if we introduce more states we will likely need to state check in a much more powerful machine...

tlaplus/ghostferry.tla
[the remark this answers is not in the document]
shuhaowu: What happened is that this comment is attached to the generated TLA+ code, and I at one point had a generated TLA+ spec that is outdated.

tlaplus/ghostferry.tla
```diff
@@ -407,33 +446,40 @@ PossibleRecords == Records \cup {NoRecordHere}
...
+                                                               /\ \A k \in (LastSuccessfulPK+1)..TableCapacity : table[k] \in (IF k = 1 \/ table[k-1] = SourceTable[k] THEN {NoRecordHere, SourceTable[k]} ELSE {NoRecordHere})
```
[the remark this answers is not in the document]
shuhaowu: Same here, this is outdated. The latest code doesn't have this `k-1` and is correct.

PR #212, opened 2020-09-23
Compare the properties of the verifiers in a table.

shuhaowu:
This transcribes an internal document we have comparing the properties of the different verifiers. The intricacies of the verifiers should probably be documented better, but this is a start.

### DIFF
--- a/docs/source/verifiers.rst
+++ b/docs/source/verifiers.rst
@@ -6,29 +6,29 @@ Verifiers
 
 Verifiers in Ghostferry is designed to ensure that Ghostferry did not
 corrupt/miss data. There are three different verifiers: the
-``ChecksumTableVerifier``, the ``IterativeVerifier``, and the ``InlineVerifier``. A comparison of them
-are given below:
+``ChecksumTableVerifier`` and the ``InlineVerifier``. A comparison of them are
+given below:
 
-+-----------------------+-----------------------+-----------------------------+-----------------------------+
-|                       | ChecksumTableVerifier | IterativeVerifier           | InlineVerifier              |
-+-----------------------+-----------------------+-----------------------------+-----------------------------+
-|Mechanism              | ``CHECKSUM TABLE``    | Verify row before cutover;  | Verify row after insert;    |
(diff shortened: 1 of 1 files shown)


docs/source/verifiers.rst
```diff
@@ -39,14 +39,67 @@ are given below:
...
+       https://bugs.mysql.com/bug.php?id=87847. This applies to every row in
+       this table.
+
+.. [4] If the rows modified by the the rogue application is modified again on
+       the source after Ghostferry starts, the InlineVerifier's binlog tailer
+       should pick up that row and attempt to reverify it.
+
+.. [5] If the rows missed after resume is modified again on the source after
+       Ghostferry starts, the InlineVerifier's binlog tailer should pick up
+       that row and attempt to reverify it.
```
fjordan, line 92: good points 👍

docs/source/verifiers.rst
```diff
@@ -39,14 +39,67 @@ are given below:
...
+| binlog coordinates.                               |               |               |                 |
++---------------------------------------------------+---------------+---------------+-----------------+
+| Data inconsistency if Ghostferry's Binlog writing | Yes           | Probably not  | Probably not    |
+| implementation is incorrect and modified the      |               | [6]_          | [6]_            |
+| wrong row on the target (example, an UPDATE is    |               |               |                 |
+| supposed to go to id = 1 but Ghostferry instead   |               |               |                 |
+| issued a query for id = 2). This is an unrealistic|               |               |                 |
+| scenario, but is included for illustrative        |               |               |                 |
+| purposes.                                         |               |               |                 |
++---------------------------------------------------+---------------+---------------+-----------------+
```
fjordan, line 77: interesting case to think about 🤔  this is if the binlog writer happens to compose and then write an incorrect row due to an incorrect pagination key?

fjordan (approved):
Thanks for the update. A few questions.

docs/source/verifiers.rst
shuhaowu: Yes, this would be difficult/impossible to catch with anything other than the ChecksumTable verifier.

PR #213, opened 2020-10-02
Invoke a custom script in the web UI

shuhaowu:
With this PR we can run a custom script with the push of a button within the web ui as follows:

Screenshot_2020-10-02 Ghostferry 127 0 0 1 29291 - 127 0 0 1 29292

The button is "Run Thing Performer" (The `Thing Performer` part is customizable as per the config).

Some problems with this PR:

- The control server is kind of unmaintained and needs to be refactored. I don't think this design was done in like a day when I originally wrote the code and didn't change since.
- We cannot stream stdout nor stderr. This is a problem for long running scripts.
- We cannot kill the script easily.

Use cases:

- For example, we could run mysqldump while Ghostferry is in cutover for small tables not supported by Ghostferry (such as those without an ordered, numeric unique column).


shuhaowu:
This is not quite ready for review... Need some additional work to make it usable...

PR #215, opened 2020-10-05
Min reporting frequency?

shuhaowu:


milanatshopify:
Other than more testing, I think this is a reasonable and quick way to make things better. May want to consider a higher minimum?

Issue #149, opened 2020-01-20
Corruption error due to InlineVerification in BatchWriter that goes away after being retried.


shuhaowu:
# Symptom 

InlineVerification in BatchWriter could log an error message such as the following:

```
level=error msg="failed to write batch to target, 1 of 5 max retries" error="row fingerprints for pks [XXXXXX] on schema.table do not match" tag=batch_writer
```

This error may show up multiple times and resolves itself. The move continues on as normal and passes the final verification.

# Possible cause

This could be caused by a harmless race condition between the BinlogWriter and the InlineVerifier. Since the InlineVerifier is not covered by TLA+, this race condition is not discovered until after implementation.

To understand the race condition, it is important to understand the operations that occur within the BatchWriter in pseudocode:

```
for i = 1..retries:
  (On source) BEGIN
  (On source) SELECT ... FOR UPDATE
  for j = 1..retries:
    (On target) BEGIN
    (On target) INSERT IGNORE ...
    (On target) SELECT MD5(... # Checksum occurs here
    (On target) COMMIT/ROLLBACK
  (On source) COMMIT/ROLLBACK
```

The error above is emitted at the `SELECT MD5(...` line, which causes the inner loop to retry. The second time it encounters `SELECT MD5(` line, the verification succeeds, causing the code to break out both the inner loop and outer loop and move on to the next batch of rows.

To understand the race, we can take a look at the state matrix of the source and target database row as the steps in the above pseudo code are executed. Specifically, we need to look at how the actions are serialized. We denote the database row values with *v1* and *v2*, where *v1* represents one set of values and *v2* represents a different set of values. We define a "payload" as either the argument an action is given (e.g. INSERT v1 INTO ...) or the results an action receives (e.g. SELECT FROM ... -> v1). Let's also assume that the error occured once and resolved itself after:

| Step |    Actor     |      Action       | Payload | Source | Target |
| ---- | -----------  | ----------------- | ------- | ------ | ------ |
|    1 | DataIterator | BEGIN             | N/A     | ??     | ??     |
|    2 | DataIterator | SELECT FOR UPDATE | **v2**  | ??     | ??     |
|    3 | BatchWriter  | BEGIN             | N/A     | ??     | ??     |
|    4 | BatchWriter  | INSERT IGNORE     | **v2**  | ??     | ??     |
|    5 | BatchWriter  | SELECT MD5()      | **v2**  | ??     | ??     |
[...]

PR #217, opened 2020-10-09
False positive InlineVerifier errors

shuhaowu:
This commit fixes a false positive problem with the InlineVerifier, which is documented in #149. That bug report documented a relatively complex and uncommon race condition that could occur and result in a false positive inline verifier error. However, with the presence of interrupt resume with periodic snapshots in the past, this problem becomes a lot more prevalent. An example case is given in the test, and is also given below for reference:

| Step | Action             | Source | Target | Notes |
|------|--------------------|--------|--------|-------|
| 0    | Initial State      | v1     | nil    |       |
| 1    | State Dump         | v1     | nil    |       |
| 2    | Copy row           | v1     | v1     |       |
| 3    | Interrupt          | v1     | v1     |       |
| 4    | App UPDATE         | v2     | v1     |       |
| 5    | Resume from step 1 | v2     | v1     |       |
| 6    | Copy row           | v2     | v1     | NO-OP due to insert ignore. |
| 7    | CHECKSUM           | v2     | v1     | InlineVerifier error here.  |
| 8    | Process halted     | v2     | v1     |       |

This means that if a move fail like this, it might never resume, as the binlog streamer could be quite lagged and therefore the error will repeat itself on every resume. Suppose if we do not error immediately on step 7 and allow the process to continue, the binlog streamer will eventually catch up and correct that row before cutover. This checksum table error could be queued in the background reverify store and periodically reverified until either the source and target matches, or until cutover, when it would fail if the discrepency is not resolved.

The following example is a run after this commit:

| Step | Action             | Source | Target | Notes |
|------|--------------------|--------|--------|-------|
| 0    | Initial State      | v1     | nil    |       |
| 1    | State Dump         | v1     | nil    |       |
| 2    | Copy row           | v1     | v1     |       |
| 3    | Interrupt          | v1     | v1     |       |
| 4    | App UPDATE         | v2     | v1     |       |
| 5    | Resume from step 1 | v2     | v1     |       |
| 6    | Copy row           | v2     | v1     | NO-OP due to insert ignore. |
| 7    | CHECKSUM           | v2     | v1     | Row added to reverify store.|
| 8    | Binlog apply       | v2     | v2     |       |
| 9    | Reverify           | v2     | v2     | Verification successful, row is dropped from reverify store |

If step 9 is not successful here because step 8 failed to apply correctly, resulting in a data corruption, the row will not be dropped
from the reverify store. During cutover, a final verification will be performed, and the Ghostferry process will error due to a verification error.

Additionally, we have some consumers of the BatchWriter that must error immediately without a background verification component. In these cases, EnforceInlineVerification is available for the BatchWriter, which will cause BatchWriter to error if mismatches are detected. This is used for the delta copier and the primary key table copier during cutover for ghostferry-sharding.

### DIFF
--- a/inline_verifier.go
+++ b/inline_verifier.go
@@ -287,14 +287,14 @@ func (v *InlineVerifier) Result() (VerificationResultAndStatus, error) {
 	return v.backgroundVerificationResultAndStatus, v.backgroundVerificationErr
 }
 
-func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetSchema, targetTable string, sourceBatch *RowBatch) ([]uint64, error) {
+func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetSchema, targetTable string, sourceBatch *RowBatch) error {
 	table := sourceBatch.TableSchema()
 
 	paginationKeys := make([]uint64, len(sourceBatch.Values()))
 	for i, row := range sourceBatch.Values() {
 		paginationKey, err := row.GetUint64(sourceBatch.PaginationKeyIndex())
 		if err != nil {
-			return nil, err
+			return err
 		}
 
 		paginationKeys[i] = paginationKey
@@ -303,7 +303,7 @@ func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetSchema, target
 	// Fetch target data
 	targetFingerprints, targetDecompressedData, err := v.getFingerprintDataFromTargetDb(targetSchema, targetTable, tx, table, paginationKeys)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	// Fetch source data
@@ -313,7 +313,7 @@ func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetSchema, target
 	for _, rowData := range sourceBatch.Values() {
 		paginationKey, err := rowData.GetUint64(sourceBatch.PaginationKeyIndex())
 		if err != nil {
-			return nil, err
+			return err
 		}
 
 		sourceDecompressedData[paginationKey] = make(map[string][]byte)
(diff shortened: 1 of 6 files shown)


inline_verifier.go
```diff
@@ -326,14 +326,24 @@ func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetSchema, target
...
+
+	for _, mismatchedPk := range mismatches {
+		v.reverifyStore.Add(table, mismatchedPk)
+	}
+
+	if len(mismatches) > 0 {
+		v.logger.WithField("mismatches", mismatches).Info("inline verification during data copy noticed mismatched pk, which is okay (issue 149)")
+	}
+
+	return nil
```
shuhaowu, line 346: This should end the main code changes. I added a lot of tests and had to change a lot of them.

inline_verifier.go
```diff
@@ -333,7 +333,17 @@ func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetSchema, target
...
+
+	for _, mismatchedPk := range mismatches {
+		v.reverifyStore.Add(table, mismatchedPk)
+	}
+
+	if len(mismatches) > 0 {
+		v.logger.WithField("mismatches", mismatches).Info("inline verification during data copy noticed mismatched pk, which may be okay (issue 149, unless overwritten due to EnforceInlineVerification)")
+	}
+
+	return mismatches, nil
```
shuhaowu, line 346: Everything below this line are tests.

tiwilliam (approved):

inline_verifier.go
```diff
@@ -333,7 +333,17 @@ func (v *InlineVerifier) CheckFingerprintInline(tx *sql.Tx, targetSchema, target
 		}
 	}
 
-	return v.compareHashesAndData(sourceFingerprints, targetFingerprints, sourceDecompressedData, targetDecompressedData), nil
+	mismatches := v.compareHashesAndData(sourceFingerprints, targetFingerprints, sourceDecompressedData, targetDecompressedData)
+
+	for _, mismatchedPk := range mismatches {
+		v.reverifyStore.Add(table, mismatchedPk)
+	}
```
xliang6, line 340: Should we check if `EnforceInlineVerification` is true and not add the mismatches to the`reverifyStore` in that case?

xliang6 (approved):

PR #220, opened 2020-10-13
Fixed initialization of SourceReplicationMaster

shuhaowu:
Since we have tcp/unix net settings now, these DatabaseConfigs have to be Validated and Initialized. They should have been validated all along, but it wasn't done.

The Ferry initialization and config validation logic should probably be refactored at some point to make simpler..


tiwilliam:
Nice. We should look into why `RunFerryFromReplica` tests aren't catching these cases.

copydb/config.go
```diff
@@ -101,6 +101,12 @@ func (c *Config) InitializeAndValidateConfig() error {
 		}
 	}
 
+	if c.SourceReplicationMaster != nil {
```
xliang6, line 104:
```suggestion
	if c.RunFerryFromReplica && c.SourceReplicationMaster != nil {
```

sharding/config.go
```diff
@@ -43,5 +43,11 @@ func (c *Config) ValidateConfig() error {
 		c.CutoverRetryWaitSeconds = 1
 	}
 
+	if c.SourceReplicationMaster != nil {
```
xliang6, line 46:
```suggestion
	if c.RunFerryFromReplica && c.SourceReplicationMaster != nil {
```

xliang6 (approved):
Thanks for fixing this quickly! 💯 

My real question is why we duplicate the following parameters for copydb and sharding configurations? Should we refactor this code so it is in the top level config and can be validated with the source and target at the top level too? If so, we should create an issue to track this and also create more test cases to catch this in the future.

```
	SourceReplicationMaster       *ghostferry.DatabaseConfig
	ReplicatedMasterPositionQuery string
	RunFerryFromReplica           bool
```

shuhaowu:
> My real question is why we duplicate the following parameters for copydb and sharding configurations?

At the time, the feature was new and we didn't want to add it directly into the base Ferry. We probably should refactor it...

xliang6:
cc @Shopify/db-engineering

PR #222, opened 2020-10-19
Make MySQL read and write timeout configurable

tiwilliam:
We have had a couple of 4 hour slow queries during a migration. These could have been more easily found if the MySQL client in Ghostferry would have timed out faster and retried or crashed after 4 retries if persistent. I've set the time to 30 seconds for both reads and writes here, maybe we want this configurable, but also don't feel for adding more complexity to config.

Ps. Using a very large chunk size might mean it takes more than 30 seconds, this case isn't really cared for in this PR. We could increase the time to 60s or so, everything is better than 4 hours here.

**Update:** I've made it configurable and set fallback to 60s.

### DIFF
--- a/config.go
+++ b/config.go
@@ -62,6 +62,14 @@ type DatabaseConfig struct {
 	Params    map[string]string
 	TLS       *TLSConfig
 
+	// ReadTimeout is used to configure the MySQL client timeout for waiting for data from server.
+	// Timeout is in seconds. Defaults to unlimited.
+	ReadTimeout uint64
+
+	// WriteTimeout is used to configure the MySQL client timeout for writing data to server.
+	// Timeout is in seconds. Defaults to unlimited.
+	WriteTimeout uint64
+
 	// SQL annotations used to differentiate Ghostferry's DMLs
 	// against other actor's. This will default to the defaultMarginalia
 	// constant above if not set.
@@ -90,6 +98,14 @@ func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
 		Params:               c.Params,
 		AllowNativePasswords: true,
 		MultiStatements:      true,
+	}
+
+	if c.ReadTimeout != 0 {
+		cfg.ReadTimeout = time.Duration(c.ReadTimeout) * time.Second
+	}
+
+	if c.WriteTimeout != 0 {
+		cfg.WriteTimeout = time.Duration(c.WriteTimeout) * time.Second
 	}
 
 	if c.TLS != nil {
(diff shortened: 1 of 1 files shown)


tiwilliam:
Really no way to catch this timeout clean from upstream libraries, this is what it will look like in logs:
```
[mysql] 2020/10/19 17:15:17 packets.go:36: read tcp 127.0.0.1:64280->127.0.0.1:29291: i/o timeout
[mysql] 2020/10/19 17:15:17 packets.go:36: read tcp 127.0.0.1:64281->127.0.0.1:29291: i/o timeout
[mysql] 2020/10/19 17:15:17 packets.go:36: read tcp 127.0.0.1:64282->127.0.0.1:29291: i/o timeout
time="2020-10-19T17:15:17+02:00" level=error msg="source connection checking failed" error="driver: bad connection" tag=ferry
```

tiwilliam:
Changed my mind, made them configurable now.

shuhaowu:
Hmm the way the code is written now means we cannot disable the timeout like before?

tiwilliam:
Yeah, we can default to nil if we want to keep current behaviour by default.

shuhaowu:
I would prefer to leave the current behavior default for now, so we dont impact the usage of ghostferry-sharding for now

tiwilliam:
I hear you, have updated now so we default to current behaviour.

config.go
```diff
@@ -538,6 +543,14 @@ type Config struct {
 	//
 	// Required: defaults to false
 	SkipTargetVerification bool
+
+	// ReadTimeout is used to configure the MySQL client timeout for waiting for data from server.
+	// Timeout is in seconds. Defaults to unlimited.
+	ReadTimeout uint64
```
shuhaowu, line 549: If these are already in DatabaseConfig, do we still need to have it on the top level here?

shuhaowu (approved):
This is fine for now. That said I am thinking that our config situation is too complex as there are a lot of variables in various levels of the config. At some point we should simplify it all...

config.go
tiwilliam: Oh didn't think about DatabaseConfig being exposed. That is indeed neat but JSON load lacks Duration support and we need to move the casting.

config.go
```diff
@@ -62,6 +62,14 @@ type DatabaseConfig struct {
 	Params    map[string]string
 	TLS       *TLSConfig
 
+	// ReadTimeout is used to configure the MySQL client timeout for waiting for data from server.
+	// Timeout is in seconds. Defaults to unlimited.
+	ReadTimeout uint64
+
+	// WriteTimeout is used to configure the MySQL client timeout for writing data to server.
+	// Timeout is in seconds. Defaults to unlimited.
```
shuhaowu, line 70: Minor nit: maybe  "defaults to 0, which means no timeout (see https://github.com/go-sql-driver/mysql#timeout)"?

shuhaowu (approved):

PR #225, opened 2021-01-05
Update the list of package cloud ubuntu dists

casperisfine:
All is in the title. I simply need the same package pushed in for other ubuntu bases.


DougEdey (approved):
oOoh we can kill trusty soon! April 2022! /s

ChrisBr (approved):

Issue #227, opened 2021-01-05
Refactor ControlServer to use Progress instead of StatusDeprecated


shuhaowu:
Right now the ControlServer uses a struct called `StatusDeprecated` for rendering. This struct is very old and it fetches data directly from the `Ferry` structs, which is not ideal. We have a `Progress` struct (introduced in https://github.com/Shopify/ghostferry/pull/118) that integrates much better with Ferry and it would be good if we used that struct instead of `StatusDeprecated` so we can remove the latter.

PR #228, opened 2021-01-06
Progress now account for more things

### LINKED ISSUE #226: Expand the Progress struct to contain more detailed information
To evaluate performance, we need a few more data points in the `Progress` struct of Ghostferry. Right now, the candidates are:

1. Amount of time spent while throttled.
2. The current number of actively copying DataIterator goroutines.
3. The number of rows copied per second (both per table and the sum).
4. The cumulative number of rows copied thus far for a particular table.
5. The number of bytes copied per second (both per table and the sum).
6. The cumulative number of bytes copied thus far for a particular table.

To avoid a very large PR, we should break this into several PRs:

- [ ] Export data for 1 and 2 (easiest).
- [ ] Export data for 3 and 4.
- [ ] Export data for 5 and 6.

For exporting table specific metrics (3-6), we can alter the `TableProgress` struct.

shuhaowu:
Related: #226 

Added the current number of active data iterators as well as the number of rows copied per table into the progress struct. This will allow us to track the performance of a Ghostferry move with a bit more precision.

The code is a bit crappy right now as we get this functionality working and measuring in real moves. The reason it's crappy is because the StateTracker now has two roles: tracking the state which is vital for interrupt and resume, as well as measuring some optional performance metrics. These two objectives ideally would be refactored so they're not as closely intertwined.


progress.go
```diff
@@ -32,6 +33,9 @@ type Progress struct {
 	BinlogStreamerLag       float64 // seconds
 	Throttled               bool
 
+	// The number of data iterators currently active.
+	ActiveDataIterators int
```
tiwilliam, line 37: Would be nice to show especially this on in the UI.

tiwilliam (approved):

progress.go
shuhaowu: Yeah I think a ControlServer refactor is coming after #227 cc @Manan007224

fjordan (approved):
Overall looks good, but I think we should just remove all the bytes written work until we implement it.

test/integration/callbacks_test.rb
```diff
@@ -20,6 +20,8 @@ def test_progress_callback
     assert_equal 1111, progress.last["Tables"]["gftest.test_table_1"]["TargetPaginationKey"]
     assert_equal "completed", progress.last["Tables"]["gftest.test_table_1"]["CurrentAction"]
 
+    assert_equal 0, progress.last["ActiveDataIterators"]
+
```
xliang6, line 29: Is it possible that we receive more than one `progress` reports during the ghostferry run? If so, can you add a check that the `ActiveDataIterators` is `>0` in one of the non-last progresses?

xliang6:
I agree with Forrest that we can hold off the code for `EstimatedBytes` till we know and have time to implement it. This PR looks good otherwise.

test/integration/callbacks_test.rb
shuhaowu: We do sometimes, so it will introduce flakiness. I've manually verified these feature for now.

we really a better integration testing environment...

PR #230, opened 2021-01-13
Add estimated row batch size in bytes to state

floriecai:
Related: https://github.com/Shopify/ghostferry/issues/226 

Similar PR: https://github.com/Shopify/ghostferry/pull/228 

Add an estimate of number of bytes for each row batch and sending it with the `progress` callback

### DIFF
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -40,6 +40,8 @@ type BatchWriter struct {
 
 	stmtCache *StmtCache
 	logger    *logrus.Entry
+
+	enableRowBatchSize bool
 }
 
 func (w *BatchWriter) Initialize() {
@@ -123,7 +125,12 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 		// Note that the state tracker expects us the track based on the original
 		// database and table names as opposed to the target ones.
 		if w.StateTracker != nil {
-			w.StateTracker.UpdateLastSuccessfulPaginationKey(batch.TableSchema().String(), endPaginationKeypos, uint64(batch.Size()), batch.EstimateByteSize())
+			bytesWrittenForThisBatch := uint64(0)
+			if w.enableRowBatchSize {
+				bytesWrittenForThisBatch = batch.EstimateByteSize()
+			}
+			w.StateTracker.UpdateLastSuccessfulPaginationKey(batch.TableSchema().String(), endPaginationKeypos,
+				RowStats{NumBytes: bytesWrittenForThisBatch, NumRows: uint64(batch.Size())})
 		}
 
 		return nil
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -107,7 +107,7 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 			if w.EnforceInlineVerification {
 				// This code should only be active if the InlineVerifier background
 				// reverification is not occuring. An example of this would be when you
-				// run the BatchWriter as a part of copying the primary table or delta
+				// run the BatchWriter as a part of copying the primary table or the delta
 				// copying the joined table.
 				if len(mismatches) > 0 {
 					tx.Rollback()
--- a/batch_writer.go
+++ b/batch_writer.go
@@ -123,7 +123,7 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 		// Note that the state tracker expects us the track based on the original
 		// database and table names as opposed to the target ones.
 		if w.StateTracker != nil {
-			w.StateTracker.UpdateLastSuccessfulPaginationKey(batch.TableSchema().String(), endPaginationKeypos, uint64(batch.Size()))
+			w.StateTracker.UpdateLastSuccessfulPaginationKey(batch.TableSchema().String(), endPaginationKeypos, uint64(batch.Size()), batch.EstimateByteSize())
 		}
 
 		return nil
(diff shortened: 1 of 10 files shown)


cursor.go
```diff
@@ -142,7 +141,6 @@ func (c *Cursor) Each(f func(*RowBatch) error) error {
 		tx.Rollback()
 
 		c.lastSuccessfulPaginationKey = paginationKeypos
-		c.rowsExamined += uint64(batch.Size())
```
shuhaowu, line 145: oh yeah this wasn't used... :whistling:..

row_batch.go
```diff
@@ -23,6 +24,19 @@ func (e *RowBatch) Values() []RowData {
...
 }
 
+func (e *RowBatch) EstimateByteSize() uint64 {
+	var total int
+	for _, v := range e.values {
+		size, err := json.Marshal(v)
+		if err != nil {
+			continue
+		}
+		total += len(size)
```
shuhaowu, line 34: This will be an overestimate, no? Any `int64` nominally will consume 8 bytes but if turned into a string can be a lot higher, no?

test/go/row_batch_test.go
```diff
@@ -108,6 +109,26 @@ func (this *RowBatchTestSuite) TestRowBatchNoPaginationKeyIndex() {
...
+		ghostferry.RowData{"hello goodbye"},
+		ghostferry.RowData{100},
+		ghostferry.RowData{t},
+	}
+
+	batch := ghostferry.NewRowBatch(this.sourceTable, vals, 0)
+
+	s := batch.EstimateByteSize()
+
+	fmt.Printf("%d", s)
```
shuhaowu, line 129: Maybe we should actually assert something?

row_batch.go
shuhaowu: Also how will this interact with binary data stored in `blob` types?

floriecai: Just checked with `MEDIUMBLOB` and a txt file read as binary data locally and it works fine. Yes there's some overestimate but i think since we're "consistently" overestimating it should be fine?

row_batch.go
```diff
@@ -23,6 +24,19 @@ func (e *RowBatch) Values() []RowData {
 	return e.values
 }
 
+func (e *RowBatch) EstimateByteSize() uint64 {
+	var total int
+	for _, v := range e.values {
+		size, err := json.Marshal(v)
```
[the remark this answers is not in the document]
floriecai: `unsafe.Sizeof` (and `reflect` - they're aliases) give you the size of the pointer. So regardless of string length, the size is 8. Same with uint64, its always gonna be 8. So it's not quite the same since we want to know the byte length of the data itself. 

And with some benchmarking, the `json.Marshal` seems harmless 👇 

```
BenchmarkSize-12         	1000000000	         0.000454 ns/op	       0 B/op	       0 allocs/op     #2.8MB file
BenchmarkJSON-12         	1000000000	         0.00634 ns/op	       0 B/op	       0 allocs/op     #2.8MB file
BenchmarkSizeSmall-12    	1000000000	         0.000077 ns/op	       0 B/op	       0 allocs/op     #200KB file
BenchmarkJSONSmall-12    	1000000000	         0.000429 ns/op	       0 B/op	       0 allocs/op     #200KB file
```

test/go/row_batch_test.go
floriecai: Removed since I'm arleady testing in the callback.

row_batch.go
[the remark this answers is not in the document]
shuhaowu: Yeah you'd have to traverse into the pointer and things gets ugly quickly (although json.Marshal is technically doing it).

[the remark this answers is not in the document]
tiwilliam: Thinking out loud here, but we could technically with minimal performance hit wrap `mysql.writeExecutePacket` and `len` the return data after filtering for `INSERTS`.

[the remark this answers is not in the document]
tiwilliam: Or no we can't, it doesn't return the data, just an error.

Manan007224:
There are some performance issues with the estimation of the row batch size in bytes noted here - https://github.com/Shopify/ghostferry/issues/240. I have currently added a flag to turn on/off the row batch size estimation which can be passed in the ghostferry config.

batch_writer.go
```diff
@@ -123,7 +125,12 @@ func (w *BatchWriter) WriteRowBatch(batch *RowBatch) error {
 		// Note that the state tracker expects us the track based on the original
 		// database and table names as opposed to the target ones.
 		if w.StateTracker != nil {
-			w.StateTracker.UpdateLastSuccessfulPaginationKey(batch.TableSchema().String(), endPaginationKeypos, uint64(batch.Size()))
+			bytesWrittenForThisBatch := uint64(0)
+			if w.enableRowBatchSize {
+				bytesWrittenForThisBatch = batch.EstimateByteSize()
+			}
+			w.StateTracker.UpdateLastSuccessfulPaginationKey(batch.TableSchema().String(), endPaginationKeypos,
+				RowStats{NumBytes: bytesWrittenForThisBatch, NumRows: uint64(batch.Size())})
```
Manan007224, line 133: This behaviour seems a bit sketchy. If someone has turned off the `EnableRowBatchSize` to off they'll see the `NumBytes` to 0 which would create confusion. I don't think there's much we can do in here other then documenting this behaviour.

shuhaowu (approved):

PR #231, opened 2021-01-13
Assert presence of docker-compose on dev up

paarthmadan:
For internal contributors, let's assert `docker-compose` is present in the `dev up` pipeline. On my machine, `Docker.app` was present but `Docker`, and its related tooling, wasn't yet installed. 

Let's strengthen the `met?` condition by asserting `which docker-compose` returns a non-error status code.

I also made some small additions to the README.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -9,8 +9,11 @@ up:
       version: 1.14.1
   - custom:
       name: Docker for Mac
-      met?: test -e /Applications/Docker.app
-      meet: echo "Docker not found. Install from https://docs.docker.com/docker-for-mac/"; false
+      met?: test -e /Applications/Docker.app && which docker-compose
+      meet: |
+        echo "Docker.app not found." ;
+        echo "Download Docker Desktop from https://docs.docker.com/docker-for-mac/install.\nInstallation includes necessary dependencies (ie. docker-compose)";
+        open "https://docs.docker.com/docker-for-mac/install" ;
   - custom:
       name: Go Dependencies
       met?: go mod download
--- a/README.md
+++ b/README.md
@@ -32,7 +32,13 @@ https://shopify.github.io/ghostferry/master/technicaloverview.html
 Development Setup
 -----------------
 
-Install:
+### Installation
+
+#### For Internal Contributors
+
+`dev up`
+
+#### For External Contributors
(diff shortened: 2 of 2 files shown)


dev.yml
```diff
@@ -9,8 +9,10 @@ up:
       version: 1.14.1
   - custom:
       name: Docker for Mac
-      met?: test -e /Applications/Docker.app
-      meet: echo "Docker not found. Install from https://docs.docker.com/docker-for-mac/"; false
+      met?: test -e /Applications/Docker.app && which docker-compose
+      meet: |
+        echo "Docker not found. Download and Install from https://docs.docker.com/docker-for-mac/install" &&
```
fjordan, line 14: should probably add some instructions about installing docker-compose as well now that we're checking for it? or is it included in docker-for-mac now?

fjordan (approved):
One small suggestion. 

Thanks for the improvements 💯

dev.yml
paarthmadan: Upon a successful installation of Docker, `docker-compose` should be installed as well. 

As per the documentation, 
> The Docker Desktop installation includes Docker Engine, Docker CLI client, **Docker Compose**, Notary, Kubernetes, and Credential Helper.

Really, the docker-compose check is a proxy for ensuring the user has _actually installed_ Docker.

fjordan: it may be confusing if the user has an older version of docker that didn't come with docker-compose. Is there a way for us to check the version and assert it is a version with docker-compose included?

shuhaowu (approved):

dev.yml
paarthmadan: Ah, this is a good point. 

I've confirmed the following:

We've been enforcing the presence of the Docker Desktop app (transitively `docker-compose`), since this commit https://github.com/Shopify/ghostferry/commit/77e217e63c4534a2956ae50aa1b637876870f8fa. 

If someone has the docker binary installed currently, but not the Docker Desktop app (and possibly not docker-compose), our dev up pipeline in its current state, will fail. This change doesn't particularly influence that behaviour.

---

I'll make sure to edit the copy to provide a hint that `docker-compose` is being checked for.

PR #233, opened 2021-01-15
Report progress before cutover begins

### LINKED ISSUE #232: ReportProgress at the beginning of cutover
After row copy is complete and during cutover, Ghostferry may still run some other code. Applications that run custom code at this phase may error and crash the Ghostferry process, which means the progress struct at the end of the copy may never be reported.

It would be nice to report progress immediately when row copy is complete, which should be around the invocation of `f.notifyRowCopyComplete` (this marks the beginning of cutover). This would guarantee that the last progress received reflects the state when all copy is done. We should leave the progress report at the end of the process as well, so the binlog position can be reported as well.

paarthmadan:
Closes #232

#### What was done?

This PR helps ensure progress is reported before beginning cutover.

Ghostferry, as a library, can't enforce or ensure that clients don't terminate/error-out during cutover. Custom code leaves this as a possibility.

Before notifying clients that row copy is complete, we should dispatch a progress callback. 

This conditional change comes before the notification (`notifyRowCopyCompletion`), to avoid any possible race conditions. If the notification of row completion occurs first, the client (ex: `copydb`, `sharding`) could execute, and possibly fail, before the progress callback is emitted.

---

#### Testing

I've add an integration test for this specific use case. Namely, simulate a Ghostferry client that is killed after row copy complete has finished.

Assertions are put in place to now make sure that – even in this case – progress is reported.

I've manually tested it out as well.

### DIFF
--- a/test/integration/callbacks_test.rb
+++ b/test/integration/callbacks_test.rb
@@ -42,6 +42,57 @@ class CallbacksTest < GhostferryTestCase
     assert progress.last["TimeTaken"] > 0
   end
 
+  def test_progress_callback_triggered_at_beginning_of_cutover_phase
+    # It's possible that during the cutover phase, a client of the Ghostferry library terminates.
+    # We report progress before entering this phase.
+    # This tests the case when the process exits before the last progress report.
+    seed_simple_database_with_single_table
+    ghostferry = new_ghostferry(MINIMAL_GHOSTFERRY, config: { verifier_type: "Inline" })
+
+    # Simulate a bad actor: the Ferry wrapper process fails after row copy completion.
+    ghostferry.on_status(Ghostferry::Status::ROW_COPY_COMPLETED) do
+      ghostferry.kill_and_wait_for_exit
+    end
+
+
+    progress = []
+    ghostferry.on_callback("progress") do |progress_data|
+      progress << progress_data
+    end
+
+    ghostferry.run_expecting_failure
+
+    assert progress.length >= 1
+
+    assert_equal "cutover", progress.last["CurrentState"]
+
+    assert_equal 1111, progress.last["Tables"]["gftest.test_table_1"]["LastSuccessfulPaginationKey"]
+    assert_equal 1111, progress.last["Tables"]["gftest.test_table_1"]["TargetPaginationKey"]
+    assert_equal "completed", progress.last["Tables"]["gftest.test_table_1"]["CurrentAction"]
+
+    result = target_db.query("SELECT COUNT(*) AS cnt FROM #{DEFAULT_FULL_TABLE_NAME}")
+    count = result.first["cnt"]
+    assert count > 0, "There should be some rows on the target, not 0."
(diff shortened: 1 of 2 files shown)


test/test_helper.rb
```diff
@@ -57,6 +57,16 @@ def new_ghostferry(filename, config: {})
     g
   end
 
+  def new_ghostferry_with_termination_after_row_copy_completed(filename, config: {})
```
shuhaowu, line 60: Minor nitpick: if there's no other tests that uses this method, might as well inline it so people who read it don't have to chase this function to a separate file?

shuhaowu (approved):
Thanks so much!

test/test_helper.rb
paarthmadan: Addressed in 766ef53: I'll clean up the history before I merge.

shuhaowu (approved):

tiwilliam (approved):

test/integration/callbacks_test.rb
```diff
@@ -41,6 +41,56 @@ def test_progress_callback
...
+    assert_equal "completed", progress.last["Tables"]["gftest.test_table_1"]["CurrentAction"]
+
+    result = target_db.query("SELECT COUNT(*) AS cnt FROM #{DEFAULT_FULL_TABLE_NAME}")
+    count = result.first["cnt"]
+    assert count > 0, "There should be some rows on the target, not 0."
+    assert_equal count, progress.last["Tables"]["gftest.test_table_1"]["RowsWritten"]
+
+    assert_equal 0, progress.last["ActiveDataIterators"]
+
+    # Note: we can make _no_ assertions about the last successful BinLog position and the Final pos, because the last progress callback is never dispatched.
```
fjordan, line 79: i'm a bit confused by this comment. Aren't the changes in this PR dispatching the progress callback before being killed?

fjordan (approved):
Looks great. One small question.

test/integration/callbacks_test.rb
paarthmadan: Ah yes, this is poorly written.

In the comment I'm trying to highlight the following observation:

Because the Ferry is killed before it reaches the end of its `Run`, there will be no progress callback dispatched after the `BinlogStreamer` has been flushed and stopped.

The progress dump that would proceed the end of cutover would have more information as it pertains to progress: for instance the `LastSuccessfulBinlogPos` would be equal to the `FinalBinlogPos`.

This comment states that in this particular case (the case of the Ferry being killed early), we can't assert that those two are equal.

1) Do you think this observation makes sense – my context is still sparse, so its possible I've misinterpreted.
2) I can remove the comment or improve it. It was primarily helpful when I was reasoning through the results – it may not be helpful generally.

fjordan: Yes, what you say above makes sense. The way it's worded was confusing, so I suggest something like:

```
 # Note: we can't know the exact below positions, so we only refute they are empty
```

PR #235, opened 2021-01-18
Added binlog verify store entry count

### LINKED ISSUE #234: InlineVerifier.currentRowCount tracks the number of rows total, not the number of entries in the BinlogVerifyStore due to pk duplication
When the same row is changed twice, it gets added to InlineVerifier's BinlogVerifyStore twice, causing `currentRowCount` to increment twice. This number is emitted as a part of the progress, as a part of the VerifierMessage so we can estimate how big the state dump will grow to. 

However, this is inaccurate because currentRowCount is incremented multiple times if the same row is changed. We need an additional, `currentEntryCount`, which only increments once if the row is added to the BinlogVerifyStore and is removed when that PK is deleted from the store.

shuhaowu:
The currentRowCount emitted before can count the same row multiple times, which does not allow for a great estimate of how big the state dump could be. With this patch, we emit both the row count (which is maybe useful to estimate how long it would take to reverify) and the entries count (which is useful to estimate how big the statedump is) in the VerifierMessage that's dumped to the progress, which will allow us to estimate the size of the state dump.

Fixes #234

### DIFF
--- a/inline_verifier.go
+++ b/inline_verifier.go
@@ -67,6 +67,17 @@ func (s BinlogVerifySerializedStore) RowCount() uint64 {
 	return v
 }
 
+func (s BinlogVerifySerializedStore) EntriesCount() uint64 {
+	var v uint64 = 0
+	for _, dbStore := range s {
+		for _, tableStore := range dbStore {
+			v += uint64(len(tableStore))
+		}
+	}
+
+	return v
+}
+
 func (s BinlogVerifySerializedStore) Copy() BinlogVerifySerializedStore {
 	copyS := make(BinlogVerifySerializedStore)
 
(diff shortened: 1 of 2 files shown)


xliang6 (approved):
The change looks good. Quick question re: number of rows vs. number of entries. Why do we need to keep track of the intermediate rows in the verify store? Don't we only need to verify the last update or do we verify all the updates sequentially?

shuhaowu:
> Why do we need to keep track of the intermediate rows in the verify store? Don't we only need to verify the last update or do we verify all the updates sequentially?

This is explained in a comment in the code: 

```
	// db => table => paginationKey => number of times it changed.
	//
	// We need to store the number of times the row has changed because of the
	// following series of events:
	//
	// 1. A row changed.
	// 2. The row gets verified in a batch.
	// 3. Still while verifying the same batch, the row verified in step 2 gets
	//    changed on the source databse. This should cause the row to be requeued
	//    to into the BinlogVerifyStore.
	// 4. The batch verification finishes and we need to delete the rows in that
	//    batch from the BinlogVerifyStore. But we cannot lose the row verified
	//    and subsequently changed in step 2 and step 3.
	//
	// Storing the number of times the row has changed means that step 4 will
	// simply decrement this number.
	//
	// We may waste some CPU cycles by verify a row multiple times unnecessarily,
	// but at least this approach is correct for now without major rework of the
	// BinlogVerifyStore data structure and any code that manipulates this store.
	store BinlogVerifySerializedStore
```

fjordan (approved):

shuhaowu:
I'm merging this for now, but will leave an issue to revisit these metrics variables.

PR #239, opened 2021-01-19
Use docker-compose v3

kirs:
It's been using docker-compose format v1 which doesn't allow containers to talk to each other by a hostname.

I know it doesn't matter too much for this project, but it's still generally helpful when debugging.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,63 +1,65 @@
-mysql-1:
-  image: percona:5.7
-  command: --server-id=1
-    --log-bin=mysql-bin
-    --max-binlog-size=4096
-    --binlog-format=ROW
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,63 +1,65 @@
...
-    --max-connections=1000
-    --read-only=OFF
-    --binlog-rows-query-log-events=ON
-  environment:
-    MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-  volumes:
-    - /var/lib/mysql
-  ports:
-    - "29291:3306"
+version: "3.9"
```
shuhaowu, line 1: It looks like travis needs a version lower than this.

PR #242, opened 2021-01-23
Refactor control server to use progress and deprecate status

Manan007224:
fixes https://github.com/Shopify/ghostferry/issues/227


progress.go
```diff
@@ -11,27 +13,48 @@ const (
...
 	CustomPayload string
 
-	Tables                  map[string]TableProgress
-	LastSuccessfulBinlogPos mysql.Position
-	BinlogStreamerLag       float64 // seconds
-	Throttled               bool
+	Tables                      map[string]TableProgress
+	LastSuccessfulBinlogPos     mysql.Position
+	TargetBinlogPos             mysql.Position
+	BinlogStreamerLag           time.Duration // seconds
```
Scalvando, line 54: Will this still show seconds over 60s? Doesn't duration format to something like `1m2s`?

ferry.go
```diff
@@ -836,13 +837,21 @@ func (f *Ferry) Progress() *Progress {
 		CurrentState:  f.OverallState.Load().(string),
 		CustomPayload: f.Config.ProgressCallback.Payload,
 		VerifierType:  f.VerifierType,
+		OverallState:  f.OverallState.Load().(string),
 	}
 
+	s.GhostferryVersion = VersionString
```
Scalvando, line 843: Do we use `s` and not `p` since it used to be `Status`?

Scalvando (approved):
Looks like it's just a swap, nice!

progress.go
Manan007224: yeah you're correct. This comment should be removed as the BinlogStreamerLag is not in seconds.

ferry.go
Manan007224: probably.

tiwilliam (approved):
Nice 🧹

progress.go
```diff
@@ -11,27 +13,48 @@ const (
...
+	GhostferryVersion string
+
+	SourceHostPort string
+	TargetHostPort string
+
+	StartTime   time.Time
+	CurrentTime time.Time
+	TimeTaken   time.Duration
+
+	AllTableNames    []string
```
shuhaowu, line 34: Adding these will significantly balloon the progress dump if you have a lot of tables, which may cause issues with downstream consumers. These table names never change during a run so there is no reason for it to be attached to the `Progress` object.

shuhaowu (changes requested):
I see that the PR currently puts everything that's on Status directly on Progress. This is not a good idea as the progress dump will always contain a lot of unchanging attributes.

So instead of keeping all variables on Progress, we should actually have another struct for templating the Ghostferry control server UI that builds on top of Progress but is independent of it.

Scalvando:
> So instead of keeping all variables on Progress, we should actually have another struct for templating the Ghostferry control server UI that builds on top of Progress but is independent of it.

Are you talking about having a struct/section that contains unchanging/shard metadata in addition to the actual progress of the migration?

shuhaowu:
> Are you talking about having a struct/section that contains unchanging/shard metadata in addition to the actual progress of the migration?

So a lot of the variables added here are only used in the Ghostferry web UI rendering and not really relevant to the progress. I'm saying we should have another struct for that, instead of adding these variables into the progress struct.

Manan007224:
> So a lot of the variables added here are only used in the Ghostferry web UI rendering and not really relevant to the progress. I'm saying we should have another struct for that, instead of adding these variables into the progress struct.

Well in that case what was wrong with the `StatusDeprecated`. It basically had all the variables required by control_server and was only used by the control_server entirely. Also `progress` struct is itself fetching values from the Ferry, so I don't see any downside of the `Status` struct fetching values from there.

shuhaowu:
> Well in that case what was wrong with the StatusDeprecated. It basically had all the variables required by control_server and was only used by the control_server entirely. Also progress struct is itself fetching values from the Ferry, so I don't see any downside of the Status struct fetching values from there.

I guess the idea is that the code fetching variables from StatusDeprecated is duplicated with Progress. It also hold its own locks, which could slowdown ghostferry a bit. It would be OK imo to rename StatusDeprecated and then populate its variables from a call to Ferry.Progress. I'm under the impression that that is the simplest way to write this code.

copydb.conf.json
```diff
@@ -0,0 +1,35 @@
+{
```
shuhaowu, line 1: What's this file for?

status_deprecated.go
```diff
@@ -12,15 +11,15 @@ import (
...
+type TableStatus struct {
 	TableName                   string
 	PaginationKeyName           string
 	Status                      string
 	LastSuccessfulPaginationKey uint64
 	TargetPaginationKey         uint64
 }
 
-type StatusDeprecated struct {
+type Status struct {
```
shuhaowu, line 22: Can we call this something other than status, since it is a synonym for Progress? Since this struct is only used for the control server's index variable templating, it should be located in that file, instead of being on its own. The name should be changed accordingly to indicate that.

status_deprecated.go
```diff
@@ -56,143 +55,67 @@ type StatusDeprecated struct {
 	VerificationErr     error
 }
 
-func FetchStatusDeprecated(f *Ferry, v Verifier) *StatusDeprecated {
-	status := &StatusDeprecated{}
+func FetchStatus(f *Ferry, v Verifier) *Status {
```
shuhaowu, line 58: If this is moved to collocate with the ControlServer, you can then attach this method directly on the ControlServer and then remove the need for the arguments.

status_deprecated.go
shuhaowu: Also, can't the Progress be embedded here to make this struct simpler? That would get rid of a bunch of copy of member variables, no?

status_deprecated.go
```diff
@@ -2,206 +2,116 @@ package ghostferry
```
shuhaowu, line 2: This file is still called `status_deprecated.go`?

Manan007224: Moved to the code to fetch status to control_server.go so this file should be removed.

control_server.go
```diff
@@ -1,16 +1,57 @@
...
 	"path/filepath"
+	"sort"
 	"sync"
 	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/sirupsen/logrus"
 )
 
+type TableStatus struct {
```
shuhaowu, line 16:
```suggestion
type ControlServerTableStatus struct {
```

control_server.go
```diff
@@ -136,3 +175,80 @@ func (this *ControlServer) HandleVerify(w http.ResponseWriter, r *http.Request)
...
+	status.BinlogStreamerStopRequested = this.F.BinlogStreamer.stopRequested
+
+	// Getting all table statuses
+	status.TableStatuses = make([]*TableStatus, 0, len(status.Tables))
+	status.CompletedTableCount = 0
+	status.TotalTableCount = len(status.Tables)
+	status.AllTableNames = make([]string, 0, len(status.Tables))
+
+	dbSet := make(map[string]bool)
+	for name, tableProgress := range status.Tables {
```
shuhaowu, line 206: Question: can we use `status.Tables` directly?

control_server.go
```diff
@@ -136,3 +175,80 @@ func (this *ControlServer) HandleVerify(w http.ResponseWriter, r *http.Request)
...
+		status.AllDatabaseNames = append(status.AllDatabaseNames, dbName)
+	}
+
+	sort.Strings(status.AllDatabaseNames)
+	sort.Strings(status.AllTableNames)
+
+	// ETA estimation
+	// We do it here rather than in DataIteratorState to give the lock back
+	// ASAP. It's not supposed to be that accurate anyway.
+	status.ETA = time.Duration(status.ETA)
```
shuhaowu, line 234: cc: @lstp this PR might impact your work as it moves a bunch of code around.

shuhaowu (approved):
Some minor questions about `TableStatus`, otherwise looks good for now. We might need to revisit this when we provide an JSON-api to consume this status, but we will cross that bridge when we get there.

control_server.go
Manan007224: Umm status.Tables contains mapping of table names to `TableProgress` which is sort of similar to `ControlServerTableStatus`. but there are some fields missing there which we need in `ControlServerTableStatus` so I just decided to keep this instead of using `status.Tables` directly.

PR #243, opened 2021-01-26
Add data iteration batch size map

Lincoln23:
After analysis, we found that the ideal batch size varies for each table this PR adds the ability to have `DataIterationBatchSizePerTableOverride ` to be configurable per table.

We can pass in `map[schema_name]: map[table_name]: batchSize` to the config to set the batchSize for each table. If no batchSize is specified, default `200` is used.

### DIFF
--- a/config.go
+++ b/config.go
@@ -443,6 +443,8 @@ type Config struct {
 	//
 	// Optional: defaults to 200
 	DataIterationBatchSize uint64
+	// Optional: Map of schemaName => tableName => uint64 to override default values for certain tables
+	DataIterationBatchSizePerTableOverride map[string]map[string]uint64
 
 	// The maximum number of retries for reads if the reads fail on the source
 	// database.
--- a/cursor.go
+++ b/cursor.go
@@ -39,6 +39,7 @@ type CursorConfig struct {
 	ColumnsToSelect []string
 	BuildSelect     func([]string, *TableSchema, uint64, uint64) (squirrel.SelectBuilder, error)
 	BatchSize       uint64
+	BatchSizeMap    map[string]map[string]uint64
 	ReadRetries     int
 }
 
@@ -58,6 +59,13 @@ func (c *CursorConfig) NewCursorWithoutRowLock(table *TableSchema, startPaginati
 	cursor := c.NewCursor(table, startPaginationKey, maxPaginationKey)
 	cursor.RowLock = false
 	return cursor
+}
+
+func (c CursorConfig) GetBatchSize(schemaName string, tableName string) uint64 {
+	if batchSize, found := c.BatchSizeMap[schemaName][tableName]; found {
+		return batchSize
+	}
+	return c.BatchSize
 }
 
 type Cursor struct {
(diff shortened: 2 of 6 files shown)


config.go
```diff
@@ -442,7 +443,7 @@ type Config struct {
 	// the copy will likely be.
 	//
 	// Optional: defaults to 200
-	DataIterationBatchSize uint64
```
shuhaowu, line 445: We should keep this so we can specify a default value, and the `map` will be an override. Otherwise if you have a lot of tables, you would have to specify this number for every table.

config.go
```diff
@@ -440,9 +454,7 @@ type Config struct {
 	// we select. This means, the larger this number is, the longer we need to
 	// hold this lock. On the flip side, the smaller this number is, the slower
 	// the copy will likely be.
-	//
-	// Optional: defaults to 200
-	DataIterationBatchSize uint64
+	DataIterationBatchSize DataIterationBatchSize
```
tiwilliam, line 457: This is breaking config backward-compatibility, is this intended? We could also keep the default variable as is and have the map as a separate struct if we want to keep compability.

shuhaowu: Agreed. we should not break backward compatibility right now. We can have an override config:

```
DataIterationBatchSize uint64
DataIterationBatchSizePerTableOverride ....
```

Lincoln23: not intended, will use a separate `DataIterationBatchSizePerTableOverride` to not break backwards compatibility

cursor.go
```diff
@@ -39,6 +39,7 @@ type CursorConfig struct {
 	ColumnsToSelect []string
 	BuildSelect     func([]string, *TableSchema, uint64, uint64) (squirrel.SelectBuilder, error)
 	BatchSize       uint64
+	BatchSizeMap    map[string]map[string]uint64
```
tiwilliam, line 42: nit: `BatchSizeOverrides` maybe?

tiwilliam (approved):

cursor.go
tiwilliam: ...and keep it in sync with the config variable name (without the `DataIteration` prefix, like `BatchSize`)

Lincoln23: ok I agree, how about `BatchSizePerTableOverride`

iterative_verifier.go
```diff
@@ -68,22 +68,23 @@ func (r *ReverifyStore) Add(entry ReverifyEntry) {
 	}
 }
 
-func (r *ReverifyStore) FlushAndBatchByTable(batchsize int) []ReverifyBatch {
+func (r *ReverifyStore) FlushAndBatchByTable(c *CursorConfig) []ReverifyBatch {
```
shuhaowu, line 71: Hmm maybe we shouldn't change this one? Since we don't really have a problem with this.

shuhaowu: Oh this is the interactive verifier. We should leave it be as that component is deprecated.

Lincoln23: ok, so for this we won't look in the override map and just use the BatchSize default

floriecai (commented):
Could you add some tests?

test/go/data_iterator_test.go
```diff
@@ -164,6 +166,20 @@ func (this *DataIteratorTestSuite) completedTables() map[string]bool {
...
+		if _, found := this.Ferry.Config.DataIterationBatchSizePerTableOverride[table.Schema]; !found {
+			this.Ferry.Config.DataIterationBatchSizePerTableOverride[table.Schema] = map[string]uint64{}
+		}
+		this.Ferry.Config.DataIterationBatchSizePerTableOverride[table.Schema][table.Name] = uint64(30)
+	}
+
+	for _, table := range this.tables {
+		this.Require().Equal(uint64(30), this.di.CursorConfig.GetBatchSize(table.Schema, table.Name))
+	}
+	this.Require().Equal(uint64(2), this.di.CursorConfig.GetBatchSize("DNE", "DNE"))
```
floriecai, line 180: nit: can you change this to `this.Ferry.Config.DataIterationBatchSize` so we know why it's "2"?

floriecai (approved):

shuhaowu (approved):

config.go
```diff
@@ -443,6 +443,8 @@ type Config struct {
 	//
 	// Optional: defaults to 200
 	DataIterationBatchSize uint64
+	// Optional: Map of schemaName => tableName => uint64 to override default values for certain tables
```
shuhaowu, line 446: We should document that the schemaName here is the "source" schema name.

Lincoln23: ok, added here https://github.com/Shopify/ghostferry/pull/246

PR #244, opened 2021-01-28
Benchmark batch size vs row size

shuhaowu:
Some scripts to benchmark Ghostferry batch size vs row size.

With this here, others can try it out too and possibly modify it for their own uses.


benchmark/studies/batch_size.rb
```diff
@@ -0,0 +1,33 @@
+require "fileutils"
+
+require_relative "../ghostferry_benchmark.rb"
+
+row_sizes = [1, 100, 200, 500]
+batch_sizes = (5000..10000).step(1000).to_a
+
+max_total_bytes = 100_000_000
```
Manan007224, line 8: Is this the max packet size that can be written to mysql ?

Manan007224 (approved):

benchmark/README.md
```diff
@@ -0,0 +1,8 @@
+Ghostferry benchmark setup
```
Manan007224, line 1: Ia this benchmark setup specific to linux environment, and if yes we should mention it in the README

Manan007224:
LGTM. Although I haven't reviewed the python analysis in depth, as am unfamiliar with how these python libraries work.

benchmark/studies/batch_size.rb
shuhaowu: No. It's just a maximum byte size I want to insert to the source so the benchmark doesn't run forever.

PR #245, opened 2021-01-29
Tagged target verifier's binlog streamer with tgt_binlog_streamer

shuhaowu:
So we can differentiate the logs...

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -443,7 +443,9 @@ func (f *Ferry) Initialize() (err error) {
 	f.BinlogStreamer = f.NewBinlogStreamer(f.SourceDB, f.Config.Source)
 
 	if !f.Config.SkipTargetVerification {
-		targetVerifier, err := NewTargetVerifier(f.TargetDB, f.StateTracker, f.NewBinlogStreamer(f.TargetDB, f.Config.Target))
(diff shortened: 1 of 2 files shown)


floriecai (approved):

ferry.go
```diff
@@ -443,7 +443,9 @@ func (f *Ferry) Initialize() (err error) {
 	f.BinlogStreamer = f.NewBinlogStreamer(f.SourceDB, f.Config.Source)
 
 	if !f.Config.SkipTargetVerification {
-		targetVerifier, err := NewTargetVerifier(f.TargetDB, f.StateTracker, f.NewBinlogStreamer(f.TargetDB, f.Config.Target))
+		targetBinlogStreamer := f.NewBinlogStreamer(f.TargetDB, f.Config.Target)
+		targetBinlogStreamer.LogTag = "tgt_binlog_streamer"
```
fjordan, line 447: This works, but I'd rather see an additional tag of `source`/`target` in addition to the `binlog_streamer` tag, unless we only get one tag, then potentially prepending it would make sense.

fjordan (approved):
It works, but a suggested alternative.

ferry.go
shuhaowu: I can tag with a hostname, but it maybe a bit verbose.. Let's go with this for now and change it later if we want to?

PR #247, opened 2021-02-03
Dynamically calculate batch size per table

Lincoln23:
This PR adds the ability to dynamically calculate the batch size using linear interpolation for all tables
Added a new `DataIterationBatchSizePerTableOverride` struct to the config 
For example, we can pass this into the config
```
{
  MinAvgRowLength: 200
  MaxAvgRowLength: 10000
  ControlPoints: {
    200:  5000,
    10000: 200,
    3000:  4000,
    4000:  3500,
    5000:  3000,
  }
}

Optionally, we can also pass in a set batch size for a given table if we don't want to calculate it for that specific table
{
  MinAvgRowLength: 200
  MaxAvgRowLength: 10000
  ControlPoints: {
    200:  5000,
    10000: 200,
    3000:  4000,
    4000:  3500,
    5000:  3000,
  }
  TableOverride : {
    "SchemaA": {
      "TableA": 200
    }
  }
}
```
For a table with `avg_row_length` of 3500, we would use the data points `3000` and `4000` to evaluate the batch size of `3750`

MySQL has an [upper limit](https://github.com/mysql/mysql-server/blob/5.7/sql/sql_prepare.cc#L2061-L2066) for [`num_params`](https://dev.mysql.com/doc/internals/en/com-stmt-prepare-response.html#packet-COM_STMT_PREPARE_OK) of `65536` for prepared statements which can cause problems [when we execute the constructed insert query](https://github.com/Shopify/ghostferry/blob/master/row_batch.go#L46-L61). if the calculated `batch size` > `65535/num_columns` we will use `65535/num_columns`

### DIFF
--- a/config.go
+++ b/config.go
@@ -442,7 +442,7 @@ func (d *DataIterationBatchSizePerTableOverride) UpdateBatchSizes(db *sql.DB, ta
 			// MySQL has an upper limit of 2^16 for num_params for prepared statements. This will cause an error when
 			// we attempt to execute a INSERT statement with > 65535 parameters.
 			// If batchSize > 65535/num_columns, we will use 65535/num_columns
-			batchSize := d.calculateBatchSize(avgRowLength)
+			batchSize := d.CalculateBatchSize(avgRowLength)
 			if batchSize >= maxBatchSize {
 				batchSize = maxBatchSize
 			}
@@ -453,7 +453,7 @@ func (d *DataIterationBatchSizePerTableOverride) UpdateBatchSizes(db *sql.DB, ta
 	return nil
 }
 
-func (d *DataIterationBatchSizePerTableOverride) calculateBatchSize(rowSize int) int {
+func (d *DataIterationBatchSizePerTableOverride) CalculateBatchSize(rowSize int) int {
 	if batchSize, found := d.ControlPoints[rowSize]; found {
 		return int(batchSize)
 	}
--- a/config.go
+++ b/config.go
@@ -376,11 +376,11 @@ func (c *CascadingPaginationColumnConfig) FallbackPaginationColumnName() (string
 }
 
 type DataIterationBatchSizePerTableOverride struct {
-	// Lower limit for rowLength, if a rowLength <= MinAvgRowLength, ControlPoints[MinAvgRowLength] will be used
-	MinAvgRowLength int
-	// Upper limit for rowLength, if a rowLength >= MaxAvgRowLength, ControlPoints[MaxAvgRowLength] will be used
-	MaxAvgRowLength int
-	// Map of rowLength => batchSize used to calculate batchSize for new rowLengths, results stored in TableOverride
+	// Lower limit for rowSize, if a rowSize <= MinRowSize, ControlPoints[MinRowSize] will be used
+	MinRowSize int
+	// Upper limit for rowSize, if a rowSize >= MaxRowSize, ControlPoints[MaxRowSize] will be used
+	MaxRowSize int
+	// Map of rowSize  => batchSize used to calculate batchSize for new rowSizes, results stored in TableOverride
 	ControlPoints map[int]uint64
 	// Map of schemaName(source schema) => tableName => batchSize to override default values for certain tables
 	TableOverride map[string]map[string]uint64
@@ -388,16 +388,15 @@ type DataIterationBatchSizePerTableOverride struct {
 
 func (d *DataIterationBatchSizePerTableOverride) Validate() error {
 	if d != nil {
-		if _, found := d.ControlPoints[d.MinAvgRowLength]; !found {
-			return fmt.Errorf("must provide batch size for MinAvgRowLength")
+		if _, found := d.ControlPoints[d.MinRowSize]; !found {
+			return fmt.Errorf("must provide batch size for MinRowSize")
 		}
-		if _, found := d.ControlPoints[d.MaxAvgRowLength]; !found {
-			return fmt.Errorf("must provide batch size for MaxAvgRowLength")
+		if _, found := d.ControlPoints[d.MaxRowSize]; !found {
+			return fmt.Errorf("must provide batch size for MaxRowSize")
 		}
 		if d.TableOverride == nil {
 			d.TableOverride = map[string]map[string]uint64{}
 		}
-
 	}
 	return nil
 }
@@ -470,7 +469,6 @@ type Config struct {
 	//
 	// Optional: defaults to 200
 	DataIterationBatchSize uint64
-	// Optional: Data Points to dynamically calculate batch size
 
 	// This optional config uses different data points to calculate
 	// batch size per table using linear interpolation
--- a/config.go
+++ b/config.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"strings"
 	"time"
 
 	sql "github.com/Shopify/ghostferry/sqlwrapper"
@@ -20,8 +21,9 @@ const (
 	VerifierTypeInline         = "Inline"
 	VerifierTypeNoVerification = "NoVerification"
 
-	DefaultNet        = "tcp"
-	DefaultMarginalia = "application:ghostferry"
+	DefaultNet          = "tcp"
+	DefaultMarginalia   = "application:ghostferry"
+	MySQLNumParamsLimit = 1<<16 - 1 // see num_params https://dev.mysql.com/doc/internals/en/com-stmt-prepare-response.html
 )
 
 type TLSConfig struct {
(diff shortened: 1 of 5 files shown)


config.go
```diff
@@ -375,6 +375,33 @@ func (c *CascadingPaginationColumnConfig) FallbackPaginationColumnName() (string
...
+}
+
+func (d *DataIterationBatchSizePerTableOverride) Validate() error {
+	if d != nil {
+		if _, found := d.ControlPoints[d.MinAvgRowLength]; !found {
+			return fmt.Errorf("must provide batch size for MinAvgRowLength")
+		}
+		if _, found := d.ControlPoints[d.MaxAvgRowLength]; !found {
+			return fmt.Errorf("must provide batch size for MaxAvgRowLength")
+		}
```
Lincoln23, line 396: Should we set `MinAvgRowLength` and `MaxAvgRowLength` to have defaults values as well?

ferry.go
```diff
@@ -1021,3 +1030,84 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
...
+	if batchSize, found := batchSizePoints.ControlPoints[rowLength]; found {
+		return batchSize
+	}
+	if rowLength >= batchSizePoints.MaxAvgRowLength {
+		return batchSizePoints.ControlPoints[batchSizePoints.MaxAvgRowLength]
+	}
+	if rowLength <= batchSizePoints.MinAvgRowLength {
+		return batchSizePoints.ControlPoints[batchSizePoints.MinAvgRowLength]
+	}
+	if rowLength >= maxRowLength {
```
floriecai, line 1089: need to move this above L1083 because database "max" should override anything we write in the config.

ferry.go
```diff
@@ -36,6 +36,7 @@ const (
 	StateVerifyBeforeCutover = "verify-before-cutover"
 	StateCutover             = "cutover"
 	StateDone                = "done"
+	MySQLNumParamsLimit      = 65535
```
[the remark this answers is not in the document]
Lincoln23: Not able to use math.Pow(2,16) since it is in a `const`, I linked to the doc and explained the usage of that number in `UpdateBatchSizes`

ferry.go
Lincoln23: This is my mistake, we should actually be checking the result of `calculateBatchSize()` against `MaxBatchSize` not using `avgRowLength`

updated to check after we return from `calculateBatchSize()`
```
batchSize := f.calculateBatchSize(avgRowLength)
if batchSize >= maxBatchSize{
  batchSize = maxBatchSize
}
f.Config.DataIterationBatchSizePerTableOverride.TableOverride[schemaName][tableName] = uint64(batchSize)
```

config.go
```diff
@@ -443,8 +469,10 @@ type Config struct {
 	//
 	// Optional: defaults to 200
 	DataIterationBatchSize uint64
-	// Optional: Map of schemaName(source schema) => tableName => uint64 to override default values for certain tables
-	DataIterationBatchSizePerTableOverride map[string]map[string]uint64
```
Lincoln23, line 447: Note: we will need to update the config to use the `DataIterationBatchSizePerTableOverride` struct when we bump the ghostferry version

ferry.go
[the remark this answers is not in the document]
shuhaowu: You can do `1 << 16` iirc.

cursor.go
```diff
@@ -62,8 +62,10 @@ func (c *CursorConfig) NewCursorWithoutRowLock(table *TableSchema, startPaginati
 }
 
 func (c CursorConfig) GetBatchSize(schemaName string, tableName string) uint64 {
-	if batchSize, found := c.BatchSizePerTableOverride[schemaName][tableName]; found {
-		return batchSize
+	if c.BatchSizePerTableOverride != nil{
```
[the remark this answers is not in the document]
shuhaowu: Should setup go fmt in your editor if you haven't done so.

ferry.go
```diff
@@ -1021,3 +1030,88 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
 
 	return nil
 }
+
+func (f *Ferry) UpdateBatchSizes() error {
```
shuhaowu, line 1034: IMO it's a bit weird to have this method, which acts on the Config to be a function attached to the Ferry. I understand that you need access to the source db connection, but maybe it's better/less surprising to have this method be attached to the config (or even just the DataIterationBatchSizePerTableOverride struct) and take a source db connection?

ferry.go
```diff
@@ -1021,3 +1030,88 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
...
+			}
+			f.Config.DataIterationBatchSizePerTableOverride.TableOverride[schemaName][tableName] = uint64(batchSize)
+		}
+
+	}
+	return nil
+}
+
+func (f *Ferry) calculateBatchSize(rowSize int) int {
+	batchSizePoints := f.Config.DataIterationBatchSizePerTableOverride
```
shuhaowu, line 1085: This method really feels like it should be attached to the DataIterationBatchSizePerTableOverride struct as opposed to be on the ferry. In fact, the first line of this function assigns `DataIterationBatchSizePerTableOverride` as a variable and no where else do we need `f` anymore.

config.go
```diff
@@ -443,8 +469,10 @@ type Config struct {
 	//
 	// Optional: defaults to 200
 	DataIterationBatchSize uint64
```
shuhaowu, line 471: This is just a comment, no action is needed for now: I can kinda see that if we moved this variable into the struct with the rest of the DataIterationBatchSize config, we can refactor some of the code a bit better (including things like GetBatchSize from CursorConfig, to the not-yet created DataIterationBatchSizeConfig struct). 

For a later time, perhaps.

shuhaowu (commented):
We should consider moving some methods around, as I find it a bit confusing to have the methods that acts on the configuration be attached to the Ferry directly. What I suggest might not be better tho, as I didn't actually try moving the code, so some judgement is required.

ferry.go
Lincoln23: I agree, it makes more sense to be on the `DataIterationBatchSizePerTableOverride` struct

ferry.go
Lincoln23: Moved into `DataIterationBatchSizePerTableOverride` struct

config.go
```diff
@@ -375,6 +377,115 @@ func (c *CascadingPaginationColumnConfig) FallbackPaginationColumnName() (string
 	return c.FallbackColumn, true
 }
 
+type DataIterationBatchSizePerTableOverride struct {
+	// Lower limit for rowSize, if a rowSize <= MinRowSize, ControlPoints[MinRowSize] will be used
+	MinRowSize int
+	// Upper limit for rowSize, if a rowSize >= MaxRowSize, ControlPoints[MaxRowSize] will be used
+	MaxRowSize int
```
shuhaowu, line 384: I'm fine with having this, but isn't this information kind of redundant? You can figure out a min and max by looking at the keys of the ControlPoints map? I suppose if you do it that way the documentation on ControlPoints will need to be a bit more complex.

config.go
```diff
@@ -375,6 +377,115 @@ func (c *CascadingPaginationColumnConfig) FallbackPaginationColumnName() (string
...
+	return linearInterpolation(
+		rowSize, closestPointLessThanRowSize,
+		int(d.ControlPoints[closestPointLessThanRowSize]),
+		closestPointGreaterThanRowSize,
+		int(d.ControlPoints[closestPointGreaterThanRowSize]),
+	)
+}
+
+func linearInterpolation(x, x0, y0, x1, y1 int) int {
+	return y0*(x1-x)/(x1-x0) + y1*(x-x0)/(x1-x0)
```
shuhaowu, line 486: Question: can we run into trouble with integer division here?

shuhaowu (commented):
A few minor nitpicks and questions. Main question for me is if integer linear interpolation is a problem. I haven't done the math, so I'm asking...

Other than those points this looks good to go.

config.go
```diff
@@ -375,6 +377,115 @@ func (c *CascadingPaginationColumnConfig) FallbackPaginationColumnName() (string
...
+		}
+		query := fmt.Sprintf(
+			`SELECT T1.TABLE_NAME, AVG_ROW_LENGTH, MAX_BATCH_SIZE
+					FROM information_schema.TABLES T1
+						JOIN (SELECT TABLE_NAME, TABLE_SCHEMA, floor(%d / count(*)) MAX_BATCH_SIZE
+						FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = '%s' AND TABLE_NAME in (%s)
+						GROUP BY table_name) T2 USING (TABLE_SCHEMA, TABLE_NAME);`,
+			MySQLNumParamsLimit,
+			schemaName,
+			"'"+strings.Join(tableNames, "', '")+"'",
```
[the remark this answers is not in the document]
Lincoln23: Found a simpler way without string concatenation with 
```
...AND TABLE_NAME in ('%s')...,
 strings.Join(tableNames, `', '`),
```

config.go
Lincoln23: It is a bit redundant but I think this makes it more clear that we must provide a `Min` and `Max` RowSize. It also makes the code a little easier since we don't need to loop through and find the min and max

config.go
Lincoln23: The only case where we can get a division by zero error is when x1 == x0, but that is not possible since we can't have duplicate keys in the control map. Also, we won't return a negative number since `x0` < `x` < `x1`

shuhaowu: I'm more thinking about rounding errors, but i guess it doesn't matter that much.

shuhaowu (approved):

config.go
Lincoln23: yea I don't think it will matter too much. we do 2 division, it will be at most off by 2

PR #250, opened 2021-02-04
Add metric to log binlog position of target verifier

floriecai:


ferry.go
```diff
@@ -831,6 +832,23 @@ func (f *Ferry) SerializeStateToJSON() (string, error) {
 		serializedState.LastKnownTableSchemaCache = nil
 	}
 
+	lastTargetPos := f.StateTracker.lastStoredBinlogPositionForTargetVerifier
+
+	binlogFile := strings.Split(lastTargetPos.Name, ",")[1]
+
+	// ex. binlog.259484 position=105060633
+	pos, err := strconv.ParseFloat(fmt.Sprintf("%s.%s", lastTargetPos.Pos, binlogFile), 64)
```
floriecai, line 840: Ex. `last_filename=binlog.259484 last_position=105060633`. If we do `259484.105060633` this will truncate the decimal place, which is why I reversed the file number and the position.

But now I'm thinking how to decipher the data in DD since we won't have a linearly increasing line anymore so maybe another possibility could be: keep the order <259484.105060633>, but pad it with 000000s. Multiply this value by 1000000 before we convert the value into a `float64` and in data dog we can apply a function like `/ 100000` to get the "true" position.

PR #253, opened 2021-02-09
Enforce foreign key constraints are not present during initialization

### LINKED ISSUE #3: Check and disallow foreign key during the beginning of the run
Foreign key is not supported (same as ghost). We should ensure ghostferry bails on start up when it detects a FK on a table as oppose to crashing in the middle of a run.

paarthmadan:
Closes #3 

Ghostferry doesn't support copying tables with Foreign Key constraints. Currently, we let the Ghostferry run begin to eventually crash midway through the run.

We should enforce this decision by erroring out during initialization, instead.

This PR queries the `information_schemas` to determine if any `FOREIGN KEY` constraints exist on any of the tables that will be copied by the Ghostferry (namely those after filtering / that are applicable).

Clients of Ghostferry may want to skip the enforcement, therefore, we leave it as a configurable option.

One instance, for example, is #166. That PR allows the ability to configure a specific ordering on tables, intended to be used to circumvent FOREIGN KEY issues. In the case when `TablesToBeCreatedFirst ` is non-zero, the `SkipForeignKeyConstraintsCheck` is turned on.

#### Testing

To test this PR, one can do the following:

1. Setup ghostferry-copydb and the necessary MySQL instances (and their permissions) as [outlined in the tutorial.](https://shopify.github.io/ghostferry/master/tutorialcopydb.html)

2. Create a database and table on the source with a FOREIGN KEY.

```sql
DROP DATABASE IF EXISTS testfk;
CREATE DATABASE testfk;
CREATE TABLE IF NOT EXISTS testfk.t1 (
  id INTEGER AUTO_INCREMENT PRIMARY KEY,
  d1 VARCHAR(16)
);

CREATE TABLE IF NOT EXISTS testfk.t2 (
  id INTEGER AUTO_INCREMENT PRIMARY KEY,
  d2 VARCHAR(16),
  t1_id INTEGER,
  FOREIGN KEY (t1_id)
    REFERENCES t1(id)
);
```

3. Seed the database.
```sh
# seed.sh
if [ "$#" -ne 1 ]; then
  echo "Usage: seed.sh <n>"
  exit
fi

N=$1

for i in $(seq 1 $N); do
  echo "INSERT INTO testfk.t1 (id, d1) VALUES (${i}, '$(cat /dev/urandom | tr -cd 'a-z0-9' | head -c 16)');"
done

for i in $(seq 1 $N); do
  echo "INSERT INTO testfk.t2 (id, d2, t1_id) VALUES (${i}, '$(cat /dev/urandom | tr -cd 'a-z0-9' | head -c 16)', $(jot -r 1 1 $N));"
done

```

4. Run the ghostferry-copydb and observe the error printed during initialization.

```
DEBU[0000] loading tables from database                  database=testfk tag=table_schema_cache
DEBU[0000] fetching table schema                         database=testfk table=t1 tag=table_schema_cache
DEBU[0000] fetching table schema                         database=testfk table=t2 tag=table_schema_cache
DEBU[0000] caching table schema                          database=testfk table=t1 tag=table_schema_cache
DEBU[0000] caching table schema                          database=testfk table=t2 tag=table_schema_cache
INFO[0000] table schemas cached                          tables="[testfk.t1 testfk.t2]" tag=table_schema_cache
ERRO[0000] source contains foreign key constraints       error="source contains at least 1 foreign key constraint" tag=ferry
error: failed to initialize ferry: source contains at least 1 foreign key constraint
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -700,6 +700,18 @@ type Config struct {
 	//
 	// Required: defaults to false
 	SkipTargetVerification bool
+	//
+	// During initialization, Ghostferry will raise an error if any
+	// foreign key constraints are detected in the source database.
+	//
+	// This check can be bypassed by setting this value to true.
+	//
+	// WARNING:
+	// Using Ghostferry with foreign keys is highly discouraged and
+	// disabling this check makes no guarantees of the success of the run.
+	//
+	// Required: defaults to false
+	SkipForeignKeyConstraintsCheck bool
 }
 
 func (c *Config) ValidateConfig() error {
--- a/copydb/config.go
+++ b/copydb/config.go
@@ -50,6 +50,8 @@ type Config struct {
 	// If a table is to be created on start and appears in this list, it is
 	// created before any other table, and is created in the order listed here.
 	// All tables not specified in this list are created in arbitrary order.
+	//
+	// Note: if used to overcome foreign key restrictions, ensure SkipForeignKeyConstraintsCheck is true.
 	TablesToBeCreatedFirst []string
 
 	// If you're running Ghostferry from a read only replica, turn this option
--- a/ferry.go
+++ b/ferry.go
@@ -443,6 +443,15 @@ func (f *Ferry) Initialize() (err error) {
 		f.Tables = f.StateToResumeFrom.LastKnownTableSchemaCache
 	}
 
+	if !f.Config.SkipForeignKeyConstraintsCheck {
+		err = f.checkSourceForeignKeyConstraints()
+
+		if err != nil {
+			f.logger.WithError(err).Error("foreign key constraints detected on the source database. Enable SkipForeignKeyConstraintsCheck to ignore this check and allow foreign keys")
+			return err
+		}
+	}
+
 	if f.Config.DataIterationBatchSizePerTableOverride != nil {
 		err = f.Config.DataIterationBatchSizePerTableOverride.UpdateBatchSizes(f.SourceDB, f.Tables)
 		if err != nil {
@@ -1033,6 +1042,42 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
 		}
 		if strings.ToUpper(value) != "ON" {
 			return fmt.Errorf("binlog_rows_query_log_events must be ON, not %s", value)
+		}
+	}
+
+	return nil
+}
+
+func (f *Ferry) checkSourceForeignKeyConstraints() error {
+	for _, table := range f.Tables {
+		query := fmt.Sprintf(`
+		SELECT COUNT(*), constraint_name
+		FROM information_schema.table_constraints
+		WHERE constraint_type = 'FOREIGN KEY'
+		AND table_schema='%s'
+		AND table_name='%s'
+		GROUP BY constraint_name
+		`,
+			table.Schema,
+			table.Name,
+		)
+
+		rows, err := f.SourceDB.Query(query)
+		defer rows.Close()
+
+		var count int
+		var name string
+
+		for rows.Next() {
+			err = rows.Scan(&count, &name)
+
+			if err != nil {
+				return err
+			}
+
+			if count > 0 {
+				return fmt.Errorf("found at least 1 foreign key constraint on source DB. table: %s.%s, constraint: %s", table.Schema, table.Name, name)
+			}
 		}
 	}
 
--- a/test/go/ferry_test.go
+++ b/test/go/ferry_test.go
@@ -18,11 +18,6 @@ func (t *FerryTestSuite) SetupTest() {
 	t.GhostferryUnitTestSuite.SetupTest()
 }
 
-func (t *FerryTestSuite) TearDownTest() {
-	_, err := t.Ferry.TargetDB.Exec("SET GLOBAL read_only = OFF")
-	t.Require().Nil(err)
-}
-
 func (t *FerryTestSuite) TestReadOnlyDatabaseFailsInitialization() {
 	_, err := t.Ferry.TargetDB.Exec("SET GLOBAL read_only = ON")
 	t.Require().Nil(err)
--- a/test/go/ferry_test.go
+++ b/test/go/ferry_test.go
@@ -39,6 +39,33 @@ func (t *FerryTestSuite) TestReadOnlyDatabaseFailsInitialization() {
 	t.Require().Nil(err)
 }
 
+func (t *FerryTestSuite) TestSourceDatabaseWithForeignKeyConstraintFailsInitialization() {
+	createTableWithFkConstraint := `
+		CREATE TABLE gftest.test_fk (
+			id bigint(20) NOT NULL AUTO_INCREMENT,
+			tt1_id BIGINT(20),
+			PRIMARY KEY(ID),
+			FOREIGN KEY(tt1_id) REFERENCES test_table_1(id)
+		)
+	`
+
+	testhelpers.SeedInitialData(t.Ferry.SourceDB, "gftest", "test_table_1", 0)
+	t.Ferry.SourceDB.Exec(createTableWithFkConstraint)
+
+	ferry := testhelpers.NewTestFerry().Ferry
+	err := ferry.Initialize()
+
+	t.Require().Equal("found at least 1 foreign key constraint on source DB. table: gftest.test_fk, constraint: test_fk_ibfk_1", err.Error())
+
+	// Initialize a ferry with SkipForeignKeyConstraintsCheck, assert no error
+	ferry = testhelpers.NewTestFerry().Ferry
+	ferry.Config.SkipForeignKeyConstraintsCheck = true
+
+	err = ferry.Initialize()
+
+	t.Require().Nil(err)
+}
+
 func TestFerryTestSuite(t *testing.T) {
 	suite.Run(t, &FerryTestSuite{GhostferryUnitTestSuite: &testhelpers.GhostferryUnitTestSuite{}})
 }


test/go/ferry_test.go
```diff
@@ -37,6 +32,31 @@ func (t *FerryTestSuite) TestReadOnlyDatabaseFailsInitialization() {
 	ferry = testhelpers.NewTestFerry().Ferry
 	err = ferry.Initialize()
 	t.Require().Nil(err)
+
+	_, err = t.Ferry.TargetDB.Exec("SET GLOBAL read_only = OFF")
+	t.Require().Nil(err)
+}
+
+func (t *FerryTestSuite) TestSourceDatabaseWithForeignKeyConstraintFailsInitialization() {
```
shuhaowu, line 40: Can you add a test for `SkipForeignKeyConstraintsCheck`

test/go/ferry_test.go
```diff
@@ -37,6 +32,31 @@ func (t *FerryTestSuite) TestReadOnlyDatabaseFailsInitialization() {
 	ferry = testhelpers.NewTestFerry().Ferry
 	err = ferry.Initialize()
 	t.Require().Nil(err)
+
+	_, err = t.Ferry.TargetDB.Exec("SET GLOBAL read_only = OFF")
+	t.Require().Nil(err)
```
shuhaowu, line 37: Why is this added?

copydb/config.go
```diff
@@ -107,6 +107,11 @@ func (c *Config) InitializeAndValidateConfig() error {
 		}
 	}
 
+	// Don't enforce that there are no foreign keys present if TablesToBeCreatedFirst is specified.
+	if len(c.TablesToBeCreatedFirst) > 0 {
+		c.SkipForeignKeyConstraintsCheck = true
+	}
```
shuhaowu, line 113: Not sure if we should have this implicit behavior.

test/go/ferry_test.go
paarthmadan: TL;DR I've removed the teardown, and inlined that code into the specific test that required it.

These lines were initially part of the test suite teardown. Previously, the test suite contained a single test, so it didn't matter whether this code lived in the teardown, or in the test itself.

Now, I've added a second test which fails because of the teardown. 

Ghostferry requires that the target DB is not a replica (read only OFF). The original test set the target DB to read only, performed assertions, and then turned read only off.

Without this change, the first test may run, followed by the newly-added second test. The teardown is run at the end of the suite, so the read only flag is still on when the second test runs, and thus fails.

We should make sure we set the target DB back to its correct state, before moving on to the next test.

copydb/config.go
paarthmadan: I agree, this conditional was definitely a code smell. 

Right now, `SkipForeignKeyConstraintsCheck` lives on the ferry, while `TablesToBeCreatedFirst` lives at the `copydb` level of abstraction. There's an inherent design boundary here.

It appears there are two options given this setup:

1. Add the implicit behaviour at the `copydb` level to set `SkipForeignKeyConstraintsCheck` (current)
2. Let the end user of the `copydb` manually configure `SkipForeignKeyConstraintsCheck` when they use `TablesToBeCreatedFirst`. Perhaps I can add a comment in the description of `TablesToBeCreatedFirst`.

There's downsides to both. In one case, there's implicit behaviour. In the other, we create a sense of coupling between two unrelated config options.

I'll go with option 2, because there _may_ be a use case of `TablesToBeCreatedFirst`, that doesn't necessarily relate to foreign key constraints.

Thoughts?

paarthmadan: 4434cc8

test/go/ferry_test.go
paarthmadan: 👍 Added a set of assertions in the same test.

copydb/config.go
shuhaowu: I would go with option 2.

shuhaowu (approved):

config.go
```diff
@@ -700,6 +700,17 @@ type Config struct {
...
+	// The Ferry will enforce that the source DB has no Foreign Key constraints.
+	// It will error out before running.
+	//
+	// It can be overriden to skip this enforcement.
+	//
+	// This allows for custom implementations to circumvent
+	// issues that come with FK constraints.
+	//
+	// Required: defaults to false
+	SkipForeignKeyConstraintsCheck bool
```
fjordan, line 713:
```suggestion
	// SkipForeignKeyConstraintsCheck is used to enforce that the 
	// source database has no Foreign Key constraints at runtime.
	// 
	// Setting the value to true will allow foreign keys on the source 
	// database during the run.
	//
	// Required: defaults to false
	SkipForeignKeyConstraintsCheck bool
```

config.go
```diff
@@ -700,6 +700,17 @@ type Config struct {
 	//
 	// Required: defaults to false
 	SkipTargetVerification bool
+	//
+	// The Ferry will enforce that the source DB has no Foreign Key constraints.
+	// It will error out before running.
+	//
+	// It can be overriden to skip this enforcement.
+	//
+	// This allows for custom implementations to circumvent
```
fjordan, line 709: what do you mean by custom implementations?

ferry.go
```diff
@@ -443,6 +443,15 @@ func (f *Ferry) Initialize() (err error) {
 		f.Tables = f.StateToResumeFrom.LastKnownTableSchemaCache
 	}
 
+	if !f.Config.SkipForeignKeyConstraintsCheck {
+		err = f.checkSourceForeignKeyConstraints()
+
+		if err != nil {
+			f.logger.WithError(err).Error("source contains foreign key constraints")
```
fjordan, line 450: nit: 
```suggestion
			f.logger.WithError(err).Error("foreign key constraints detected on the source database. Enable SkipForeignKeyConstraintsCheck to ignore this check and allow foreign keys")
```

ferry.go
```diff
@@ -1038,3 +1047,36 @@ func (f *Ferry) checkConnectionForBinlogFormat(db *sql.DB) error {
...
+
+		rows.Next()
+		err = rows.Scan(&count)
+
+		if err != nil {
+			return err
+		}
+
+		if count > 0 {
+			return fmt.Errorf("source contains at least 1 foreign key constraint")
```
fjordan, line 1077: I think it would be helpful to add the table and the constraint in the error message. Is that doable?

test/go/ferry_test.go
```diff
@@ -18,11 +18,6 @@ func (t *FerryTestSuite) SetupTest() {
 	t.GhostferryUnitTestSuite.SetupTest()
 }
 
-func (t *FerryTestSuite) TearDownTest() {
```
fjordan, line 21: Why was this removed?

fjordan (commented):
Overall looks good, just a few questions and suggested improvements.

copydb/config.go
```diff
@@ -107,6 +107,11 @@ func (c *Config) InitializeAndValidateConfig() error {
 		}
 	}
 
+	// Don't enforce that there are no foreign keys present if TablesToBeCreatedFirst is specified.
+	if len(c.TablesToBeCreatedFirst) > 0 {
+		c.SkipForeignKeyConstraintsCheck = true
```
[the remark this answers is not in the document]
paarthmadan: This was part of an older revision. It's changed now, but for context, I wanted to make sure that the copydb usage of `TablesToBeCreatedFirst` isn't hindered by the default value for the `SkipForeignKeyConstraintsCheck`. 

If someone wanted to specify the ordering of tables to circumvent foreign key restrictions, they'd set tables, but they'd _also_ need to skip the check. I added that behaviour implicitly. I scrapped this behaviour when revisiting the decision in the first review. Now, the user configures both options independently.

After all, there may bye a case when a user specifies and order, and still wants foreign key constraints to be checked.

config.go
paarthmadan: I like these changes, though there is one point I'd like to emphasize. I don't want to give the impression that by enabling this check, the ferry won't error out during runtime. 

I feel as though the statement: "Setting the value to true will allow foreign keys on the source database during the run", may suggest that Ghostferry _can_ handle foreign key constraints, which is false.

test/go/ferry_test.go
paarthmadan: My mistake, I resolved a previous clarification pre-emptively. See [here](https://github.com/Shopify/ghostferry/pull/253#discussion_r580348436) for more context.

TL;DR This teardown was specific to the _single_ test that lived in this suite, so it went uncaught. The teardown doesn't apply to all tests, so I removed it.

config.go
paarthmadan: I realize now, the wording is quite ambiguous. I meant that a ferry could employ its own table ordering to "work around" foreign key constraints. I'll make some edits along with your suggestions below 👍

ferry.go
paarthmadan: Addressed in fef68e5

config.go
paarthmadan: I think I found a good description that makes the usage clearer:

Addressed in 8fe9a75

config.go
```diff
@@ -700,6 +700,18 @@ type Config struct {
...
+	// Ghostferry does not support using foreign key constraints.
+	// During initialization, Ghostferry will raise an error
+	// if any foreign key constraints are detected in the source DB.
+	//
+	// This check can be disabled.
+	//
+	// Note: Disabling the check does not guarantee the success of the run.
+	//
+	//
+	// Required: defaults to false
```
fjordan, line 713: some small nits:

```suggestion
	// During initialization, Ghostferry will raise an error if any
	// foreign key constraints are detected in the source database.
	//
	// This check can be bypassed by setting this value to true.
	//
	// WARNING: 
	// Using Ghostferry with foreign keys is highly discouraged and
	// disabling this check makes no guarantees of the success of the run.
	//
	// Required: defaults to false
```

fjordan (approved):
Nice work. 

Another small suggestion on the wording of the comments above the setting.

PR #254, opened 2021-02-09
Update docs with necessary replication permissions

paarthmadan:
The [copydb tutorial](https://shopify.github.io/ghostferry/master/tutorialcopydb.html) lists an example configuration file. 

The example configuration file uses the `ghostferry` user, which is the user we're instructed to create permissions for. 

The actual example config listed in the source at `examples/copydb/conf.json` uses the root user. None of those grants are necessary then. 

If you use the config listed in the tutorial as is, and the permissions listed, errors are thrown.

We should:
1. Correct the list of necessary permissions to be granted
2. Update the config file in the examples to align with the docs


fjordan (approved):
do we still need `SUPER` privileges as well? And also on the target?

paarthmadan:
> do we still need `SUPER` privileges as well? 

It my local tests, I don't require `SUPER` privileges for the source or the target.

> And also on the target?

The change in this PR ensures we list out `REPLICATION SLAVE, REPLICATION CLIENT` as some of the necessary privileges on the target, as our docs didn't say this prior.

shuhaowu:
Wait. We need SUPER if source is a read only replica, because we use FOR UPDATE.

shuhaowu:
So we don't need those permissions if we set SkipTargetVerification: true in the example config...

In fact we might want to default it to that value for copydb, because copydb tends to run without target verification....

paarthmadan:
@shuhaowu My mistake, I should've been clearer. For the `copydb` tutorial, the minimum permissions required are those mentioned. The source in the tutorial is not setup as a read-only replica, thus we don't need `SUPER`. Do you think this is okay?

I'll update the PR so that the copydb tutorial skips target verification (and thus does not require privileges for the target)

shuhaowu:
> The source in the tutorial is not setup as a read-only replica, thus we don't need SUPER. Do you think this is okay?
> I'll update the PR so that the copydb tutorial skips target verification (and thus does not require privileges for the target)

Both sounds good to me.

tiwilliam:
These configs work fine with the committed docker compose + seed commands in docs, is this still the case after switching to the ghostferry user? If not I think we should create the user and grants on `dev up`.

PR #255, opened 2021-02-10
allow checksumtable verification during cutover only when overallstat…

Manan007224:
Currently we allow running the verificaiton via webui if and only if ghostferry is in `wait-for-cutover` or `done` phase. We should not allow `ChecksumTable` verification if we're in the `wait-for-cutover` because `wait-for-cutover`  phase means that binlog-streamer is still running and neither source or target DB are read-only. Given that writes might be happening to source and target `ChecksumTable` verification will definetly fail. 

This PR makes the `ChecksumTable` verification available only if ghostferry is in `done` phase.

### DIFF
--- a/control_server.go
+++ b/control_server.go
@@ -241,8 +241,14 @@ func (this *ControlServer) fetchStatus() *ControlServerStatus {
 		status.VerificationStarted = result.IsStarted()
 		status.VerificationDone = result.IsDone()
 
-		// We can only run the verifier if we're not copying and not verifying
-		status.VerifierAvailable = status.OverallState != StateStarting && status.OverallState != StateCopying && (!status.VerificationStarted || status.VerificationDone)
+		// If the VerifierType is CheckSumTable we can only run the verifier when the we're done.
+		// For other verifiers we can run verification if we're not copying or verifying.
+		if this.F.Config.VerifierType == VerifierTypeChecksumTable {
+			status.VerifierAvailable = (status.OverallState == StateDone)
+		} else {
+			status.VerifierAvailable = status.OverallState != StateStarting && status.OverallState != StateCopying && (!status.VerificationStarted || status.VerificationDone)
+		}
+
 		status.VerificationResult = result.VerificationResult
 		status.VerificationErr = err
 	} else {
--- a/docs/source/tutorialcopydb.rst
+++ b/docs/source/tutorialcopydb.rst
@@ -234,11 +234,12 @@ to:
   allow external scripts (configured via the json configuration) to be
   automatically executed with the push of this button so you can perform
   operations you need to perform during cutover.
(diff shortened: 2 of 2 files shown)


Manan007224:
> This is not correct. One common mode of operations is to turn the source database to be read-only during "wait-for-cutover". This would allow you to run the checksum table verifier and get a "correct" result.

Although the source database would be read-only during "wait-for-cutover" this shouldn't mean that the target database would no longer have any writes. The reason is :- 
- The binlog-writer stops once binlog-streamer has stopped running - https://github.com/Shopify/ghostferry/blob/master/ferry.go#L734-L742. The binlog-streamer only stops streaming once copydb instructs to stop it - https://github.com/Shopify/ghostferry/blob/master/copydb/copydb.go#L106. 

- The point at which copydb instructs binlog-streamer to stop, ghostferry is already in `cutover` phase and not `wait-cutover` - https://github.com/Shopify/ghostferry/blob/master/ferry.go#L734-L742

To conclude the above points we can be for sure that binlog-streamer and binlog-writer has stopped only in `done` phase. Since binlog-writer and binlog-streamer can still be running during `wait-for-cutover` phase we can't gurantee that there are not writes happening to the target database and hence ChecksumTable verifier can fail.

shuhaowu:
You're right in the absolute/mathematical sense: we cannot be sure that the binlog streamer and writer is done until the `done` phase. After all, the OS scheduler could forever not schedule those goroutines and enter some sort of [stuttering scenario](https://lamport.azurewebsites.net/tla/stuttering.html?back-link=advanced.html) such that the writes are forever buffered within Ghostferry. This, however, doesn't happen in practice. Generally, the binlog streamer and writer typically are done only a few seconds/minutes after you set the source db to be read-only. This is easy to verify by comparing the binlog streamer lag reported by Ghostferry with amount of time that passed since you made the DB read-only. After this verification, you can run the verifier at this stage, which should give you a "correct" result. We should assume that the users of Ghostferry understand how Ghostferry work and this "advanced" usage of Ghostferry should not be artificially restricted simply due to a theoretical race condition that is unlikely to be observed in practise.

Additionally, there's an advantage to running this way: when Ghostferry is in `wait-for-cutover`, binlog streaming is technically not terminated (as you mentioned), even though in practise there should be no event. This allows you to "resume" streaming the binlog should you decide to abort the cutover, without having to go through interrupt and resume.

Now, with all that said, we can question if we named the states correctly, or if we should introduce extra states within Ghostferry to make all of this more clear. Right now, it's common for copydb, that cutover happens during the `wait-for-cutover` stage (which includes things like setting the source to read only, flipping the application to the new database, etc). Clicking "Allow automatic cutover" in the UI is simply a synonym to quit Ghostferry. This causes some confusion, even for me, and I'm interested in a way to refactor that part of the code.

Issue #251, opened 2021-02-08
ControlServer ETA is negative after "Done"


shuhaowu:
It seems like the ETA in the ControlServer UI continues to calculate after the Done state, which results in a negative ETA value. This should be fixed.

![image](https://user-images.githubusercontent.com/338100/107240265-bb96e880-69f7-11eb-9446-2a9078e6d65f.png)

lstp:
@shuhaowu I wasn't able to replicate this locally. Do you mind adding steps to replicate the issue. I see ETA as 0 when I run a test locally. 


![Screen Shot 2021-02-09 at 1 23 55 PM](https://user-images.githubusercontent.com/4635189/107419288-9c768480-6ae5-11eb-8d54-9f2bfc4f4b5e.png)

tiwilliam:
This could be due to `targetPaginationKeys` being empty or almost empty if resuming a run. That would cause `totalPaginationKeysToCopy` to be smaller than `completedPaginationKeys` and generate a negative number.

lstp:
resolved in https://github.com/Shopify/ghostferry/pull/256

PR #256, opened 2021-02-11
Prevent ETA from becoming a negative value

lstp:
This PR addresses this issue: https://github.com/Shopify/ghostferry/issues/251#issuecomment-776562382

I wasn't able to reproduce the issue myself but this should help in case  we somehow get there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Ghostferry is a library that enables you to selectively copy data from one mysql
 
 It is inspired by Github's [gh-ost](https://github.com/github/gh-ost),
 although instead of copying data from and to the same database, Ghostferry
-copies data from one database to another and have the ability to only
+copies data from one database to another and has the ability to only
 partially copy data.
 
 There is an example application called ghostferry-copydb included (under the
--- a/ferry.go
+++ b/ferry.go
@@ -917,8 +917,12 @@ func (f *Ferry) Progress() *Progress {
 	for _, completedPaginationKey := range serializedState.LastSuccessfulPaginationKeys {
 		completedPaginationKeys += completedPaginationKey
 	}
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -54,6 +54,8 @@ Testing
 #### Run example copydb usage
 
 - `make copydb && ghostferry-copydb -verbose examples/copydb/conf.json`
+- In your browser you can visit http://localhost:8000/ to access the UI
+- Click on `Allow Automatic Cutover` button
```
shuhaowu, line 58: Isn't this in the tutorial already?

ferry.go
```diff
@@ -917,8 +917,9 @@ func (f *Ferry) Progress() *Progress {
 	for _, completedPaginationKey := range serializedState.LastSuccessfulPaginationKeys {
 		completedPaginationKeys += completedPaginationKey
 	}
-
-	s.ETA = (time.Duration(math.Ceil(float64(totalPaginationKeysToCopy-completedPaginationKeys)/estimatedPaginationKeysPerSecond)) * time.Second).Seconds()
+	var remainingPaginationKeys float64 = 0
+	remainingPaginationKeys = math.Max(0, float64(totalPaginationKeysToCopy-completedPaginationKeys))
```
shuhaowu, line 921: Can you make a comment inline on why this is possible? iirc it's possible because we can overshoot with the batching, if rows are inserted between when ghostferry is started and when all the copy is completed.

shuhaowu (approved):
A minor comment in the code is requested, otherwise looks good.

README.md
lstp: It definitely is in the tutorial. I thought it would be helpful to mention here at the time. I can remove it.

tiwilliam (approved):

PR #258, opened 2021-02-18
Fix BinlogStreamer for TargetVerifier by correctly handling rewrites

fjordan:
## Overview

Fixes an issue in the `TargetVerifier` if there happened to be any configured schema or table rewrites.

This patch correctly handles any configured rewrites by reversing the configurations to lookup DML events using the Source `TableSchemaCache` from DML events streamed on the Target. 

Note that If there are duplicate values in the original rewrites, an error will be raised and Target Verification will not be able to be used as Ghostferry will not be able to correctly map the DML back to the Source.

### Details

Previous to this fix,  if running Ghostferry with schema or table rewrites, it would fail to inspect binlog events on the Target, and, as a result, fail to update its position in the `StateTracker`.

As you can see in the snippet below, the `BinlogStreamer` in the `TargetVerifier` would silently return from any events that did not exist in the `TableSchema`, which is correct, because Ghostferry should only ever care about the schema and tables for which it is configured: 

https://github.com/Shopify/ghostferry/blob/203cd1a16ef38300638770b1c9fcc7fa6e230ab0/binlog_streamer.go#L353-L356

However, because the `TargetVerifier` may be running against a Target with configured rewrites (differently named databases or tables), these lookups must either:
1.  Be mapped back to the original `TableSchema` of the Source or 
2. Be mapped to another `TableSchema` with the configured rewrites applied. 

I originally attempted the second approach, but abandoned the idea after failing to track down some supposed side effects that were occurring on the `f.Tables` (the Source `TableSchemaCache`) within an allotted amount of time. Eventually, I was able to get a working solution to the first approach, which is this PR. 

### Other considerations

We should consider rethinking the `TableSchema` and `TableSchemaCache` components and how they are used within the Ferry. The abstractions are difficult to work with and made this patch a bit more difficult than it should have been. 

Performing the lookups and applying the rewrites to the DML events in all the various places (`BatchWriter`, `InlineVerifier`, `BinlogWriter`, and now `BinlogStreamer`) is a bit of a duplicated effort. It may make more sense to apply the rewrites to the applicable `TableSchema`s and `TableSchemaCache`s once and then hand them off to their respective components. 

cc @Shopify/db-engineering

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -29,6 +29,11 @@ type BinlogStreamer struct {
 	binlogSyncer   *replication.BinlogSyncer
 	binlogStreamer *replication.BinlogStreamer
 
+	// These rewrite structures are used specifically for the Target
+	// Verifier as it needs to map events streamed from the Target back
+	// to the TableSchemaCache of the Source
+	//
+	// See https://github.com/Shopify/ghostferry/pull/258 for details
 	DatabaseRewrites map[string]string
 	TableRewrites    map[string]string
 
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -29,6 +29,9 @@ type BinlogStreamer struct {
 	binlogSyncer   *replication.BinlogSyncer
 	binlogStreamer *replication.BinlogStreamer
 
+	DatabaseRewrites map[string]string
+	TableRewrites    map[string]string
+
 	lastStreamedBinlogPosition  mysql.Position
 	lastResumableBinlogPosition mysql.Position
 	stopAtBinlogPosition        mysql.Position
@@ -350,12 +353,22 @@ func (s *BinlogStreamer) handleRowsEvent(ev *replication.BinlogEvent, query []by
 		Pos:  ev.Header.LogPos,
 	}
 
-	table := s.TableSchema.Get(string(rowsEvent.Table.Schema), string(rowsEvent.Table.Table))
-	if table == nil {
+	db := string(rowsEvent.Table.Schema)
+	if rewrittenDBName, exists := s.DatabaseRewrites[db]; exists {
+		db = rewrittenDBName
+	}
+
+	table := string(rowsEvent.Table.Table)
+	if rewrittenTableName, exists := s.TableRewrites[table]; exists {
+		table = rewrittenTableName
+	}
+
+	tableFromSchemaCache := s.TableSchema.Get(db, table)
+	if tableFromSchemaCache == nil {
 		return nil
 	}
 
-	dmlEvs, err := NewBinlogDMLEvents(table, ev, pos, s.lastResumableBinlogPosition, query)
+	dmlEvs, err := NewBinlogDMLEvents(tableFromSchemaCache, ev, pos, s.lastResumableBinlogPosition, query)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


tiwilliam (approved):
Nice work!

binlog_streamer.go
```diff
@@ -29,6 +29,9 @@ type BinlogStreamer struct {
 	binlogSyncer   *replication.BinlogSyncer
 	binlogStreamer *replication.BinlogStreamer
 
+	DatabaseRewrites map[string]string
+	TableRewrites    map[string]string
```
shuhaowu, line 33: Maybe worth adding a comment linking to this PR as the description is very good. 

Also maybe confusing because this rewrite map **intended to be used** for cases going from target -> source, not source -> target in every other place where we do this. That said, hypothetically we _could_ rewrite the source binlog streamer so this is _theoretically_ generic, but not sure if we should/would.

ferry.go
```diff
@@ -452,11 +473,14 @@ func (f *Ferry) Initialize() (err error) {
 
 	// The iterative verifier needs the binlog streamer so this has to be first.
 	// Eventually this can be moved below the verifier initialization.
-	f.BinlogStreamer = f.NewBinlogStreamer(f.SourceDB, f.Config.Source)
+	f.BinlogStreamer = f.NewSourceBinlogStreamer()
 
 	if !f.Config.SkipTargetVerification {
-		targetBinlogStreamer := f.NewBinlogStreamer(f.TargetDB, f.Config.Target)
-		targetBinlogStreamer.LogTag = "tgt_binlog_streamer"
+		targetBinlogStreamer, err := f.NewTargetBinlogStreamer()
```
shuhaowu, line 479: This is fine, but it was a bit surprising at first to see the the `SourceBinlogStreamer` doesn't return an error while this one does. Makes sense, but it did cause me to re-read that code again.

test/go/table_schema_cache_test.go
```diff
@@ -428,6 +427,23 @@ func (this *TableSchemaCacheTestSuite) TestGetTableListWithPriorityIgnoreUnknown
 	this.Require().Equal(creationOrder[0], "schema.table2")
 }
 
+func (this *TableSchemaCacheTestSuite) TestTargetVerifierErrorsOnDuplicateRewriteValue() {
```
shuhaowu, line 430: nit: 
```suggestion
func (this *TableSchemaCacheTestSuite) TestTargetToSourceRewritesErrorsOnDuplicateRewriteValue() {
```

shuhaowu (approved):
Thanks for the fix!

binlog_streamer.go
fjordan: > Also maybe confusing because this rewrite map intended to be used for cases going from target -> source, not source -> target in every other place where we do this. That said, hypothetically we could rewrite the source binlog streamer so this is theoretically generic, but not sure if we should/would.

Yes, agreed this is potentially confusing. I think reworking this is something to be considered as part of a larger rewrite as mentioned in the PR description.

ferry.go
fjordan: Right. Would you prefer to see a common interface (ie: both return errors) even if the `NewSourceBinlogStreamer` will always return a `nil` error for now?

shuhaowu: I think what you have here is fine so we can leave as is. I just tend to write a comment when I spend more than 5 minutes investigating a single line of code, which maybe a bad rule to follow as in this case it didn't result in anything.

shuhaowu (approved):

PR #261, opened 2021-02-22
Add sanity check to InlineVerifier and TargetVerifier

Lincoln23:
#263

For InlineVerifier: 
 check if `rowsChecked > 0` when we run `VerifyDuringCutover`.
 if no rows were checked, return `VerificationResult.DataCorrect = False`

For TargetVerifier:
 check after we run `StopTargetVerifier` to see if we verified any events
 if no events were checked, we log and return an error, copydb and sharding will then panic

### DIFF
--- a/inline_verifier.go
+++ b/inline_verifier.go
@@ -248,6 +248,7 @@ type InlineVerifier struct {
 	reverifyStore              *BinlogVerifyStore
 	verifyDuringCutoverStarted AtomicBoolean
 
+	rowsChecked     int
 	sourceStmtCache *StmtCache
 	targetStmtCache *StmtCache
 	logger          *logrus.Entry
@@ -435,6 +436,13 @@ func (v *InlineVerifier) VerifyDuringCutover() (VerificationResult, error) {
 		return VerificationResult{}, err
 	}
 
+	if v.rowsChecked == 0 {
+		return VerificationResult{
+			DataCorrect: false,
+			Message:     "cutover verification failed, no rows were compared",
+		}, nil
+	}
+
 	if !mismatchFound {
 		return VerificationResult{
 			DataCorrect: true,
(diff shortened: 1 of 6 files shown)


test/integration/inline_verifier_test.rb
```diff
@@ -68,7 +68,7 @@ def test_different_compressed_data_is_detected_inline_with_batch_writer
 
     assert verification_ran
     assert_equal ["#{DEFAULT_DB}.#{DEFAULT_TABLE}"], incorrect_tables
-    assert_equal "cutover verification failed for: gftest.test_table_1 [paginationKeys: 1 ] ", ghostferry.error_lines.last["msg"]
+    assert_equal "cutover verification failed for: gftest.test_table_1 [paginationKeys: 1 ] ", ghostferry.error_lines.first["msg"]
```
Lincoln23, line 71: These tests will log `target verifier did not check any events` since there are no events on the target and is the last item in the `error_lines` array. I added a separate test to check target verifier events below so I will ignore that error here

Same changes for tests below

inline_verifier.go
```diff
@@ -435,6 +436,13 @@ func (v *InlineVerifier) VerifyDuringCutover() (VerificationResult, error) {
 		return VerificationResult{}, err
 	}
 
+	if v.rowsChecked == 0 {
+		return VerificationResult{
+			DataCorrect: false,
+			Message:     "cutover verification failed, no rows were compared",
+		}, nil
+	}
```
Lincoln23, line 444: I wasn't sure how to test this in the ruby integration tests since in integrationferry [here](https://github.com/Shopify/ghostferry/blob/master/test/lib/go/integrationferry.go#L164) we only pass the incorrect tables. Any ideas?

PR #264, opened 2021-03-09
JSON Enpoints

Scalvando:
Adding in a way to communicate with the API via JSON.

This is a rough draft of what the API will look like and how it will behave.

### DIFF
--- a/control_server.go
+++ b/control_server.go
@@ -1,6 +1,7 @@
 package ghostferry
 
 import (
+	"encoding/json"
 	"fmt"
 	"html/template"
 	"net/http"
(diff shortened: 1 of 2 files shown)


control_server.go
```diff
@@ -180,18 +185,22 @@ func (this *ControlServer) HandleUnpause(w http.ResponseWriter, r *http.Request)
...
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
 
-	if vars["type"] == "automatic" {
+	if cutover == "automatic" {
 		this.F.AutomaticCutover = true
-	} else if vars["type"] == "manual" {
+	} else if cutover == "manual" {
```
shuhaowu, line 196: lol at the `"manual"`.. I definitely wasn't careful when I wrote this code originally with regards to how to name different things. We should definitely refactor the cutover process to make things more clear (cc @floriecai )

control_server.go
```diff
@@ -210,7 +219,13 @@ func (this *ControlServer) HandleVerify(w http.ResponseWriter, r *http.Request)
...
 	}
 
-	http.Redirect(w, r, "/", http.StatusSeeOther)
+	returnSuccess(w, r)
+}
+
+func (this *ControlServer) HandleProgress(w http.ResponseWriter, r *http.Request) {
+	status := this.fetchStatus()
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(status)
```
shuhaowu, line 228: Is there no error for this?

control_server.go
```diff
@@ -86,6 +87,10 @@ func (this *ControlServer) Initialize() (err error) {
 
 	this.router = mux.NewRouter()
 	this.router.HandleFunc("/", this.HandleIndex).Methods("GET")
+	this.router.HandleFunc("/api/progress", this.HandleProgress).Headers("Content-Type", "application/json").Methods("GET")
+	this.router.HandleFunc("/api/cutover", this.HandleCutover).Headers("Content-Type", "application/json").Methods("POST")
+	this.router.HandleFunc("/api/verify", this.HandleVerify).Headers("Content-Type", "application/json").Methods("POST")
+	this.router.HandleFunc("/api/custom_script", this.HandleScript).Headers("Content-Type", "application/json").Methods("POST")
```
shuhaowu, line 93: I see the previous routes are `/api/actions/cutover`, worth it to unify the routes?

shuhaowu (approved):
Generally looks good, some minor nitpicks

control_server.go
Scalvando: I think so

control_server.go
Scalvando: It can return an error, at this point it will have to be logging I think

floriecai (approved):

control_server.go
```diff
@@ -210,7 +216,16 @@ func (this *ControlServer) HandleVerify(w http.ResponseWriter, r *http.Request)
...
 	}
 
-	http.Redirect(w, r, "/", http.StatusSeeOther)
+	returnSuccess(w, r)
+}
+
+func (this *ControlServer) HandleStatus(w http.ResponseWriter, r *http.Request) {
+	status := this.fetchStatus()
+	w.Header().Set("Content-Type", "application/json")
+	err := json.NewEncoder(w).Encode(status.Progress)
```
Lincoln23, line 225: We want to encode the whole `status` and not only `status.Progress` right?

Scalvando: Yup, I was messing around with things and forgot to revert this change

Lincoln23 (approved):

PR #268, opened 2021-03-26
sharding: made metrics buffer bigger

shuhaowu:
I'm seeing metrics sink full warnings in production spamming our logs. I tested making the buffer bigger and it appears to have helped...

That said, I'm not sure why making metrics buffer bigger actually solves the problem, because there's should be a "conservation of mass" type dynamics: the number of metrics going into the channel should equal the number coming out for the channel to not be filled. I guess buffers for audio workstations also have similar behaviours, so maybe I should just read up more about how buffers work.

### DIFF
--- a/sharding/metrics.go
+++ b/sharding/metrics.go
@@ -22,7 +22,7 @@ func InitializeMetrics(prefix string, config *Config) error {
 		return err
 	}
 
-	metricsChan := make(chan interface{}, 1024)
+	metricsChan := make(chan interface{}, 4096)  // TODO: make this queue size configurable
 	SetGlobalMetrics(prefix, metricsChan)
 
 	metrics.DefaultTags = []ghostferry.MetricTag{


milanatshopify (approved):
So, fewer times, but larger payload each time?  Are we worried about the payload size?

Does this mean that there is more chance of us not getting logs when something bad happens, because they're more likely to be stuck in the buffer?

shuhaowu:
This is just the datadog/statsd metrics that we get. When this channel gets full, a warning is emitted by Ghostferry. If it gets full fairly often, the ghostferry logs in stderr gets spammed. I'm not sure what you mean by payload?

I'm not even sure the metrics are being used anywhere, but i didn't want to break it for now if I can resolve the log spam issue. Also statsd metrics are best-effort anyway, as my understanding is that it is transmitted via UDP.

Scalvando (approved):

PR #270, opened 2021-03-29
add errorcallback to copydb

Manan007224:
Adding a field in `ErrorCallback` in copydb config, so that the errors can be posted to a callback server endpoint.

### DIFF
--- a/config.go
+++ b/config.go
@@ -722,6 +722,8 @@ type Config struct {
 	// in the Progress. This behaviour is perfectly okay and doesn't mean there are no rows being written
 	// to the target DB.
 	EnableRowBatchSize bool
+
+	ErrorCallback HTTPCallback
 }
 
 func (c *Config) ValidateConfig() error {
--- a/config.go
+++ b/config.go
@@ -605,6 +605,10 @@ type Config struct {
 	StateCallback        HTTPCallback
 	StateReportFrequency int
 
+	// Report error via an HTTP callback. The Payload field will contain the ErrorType,
+	// ErrorMessage and the StateDump.
+	ErrorCallback HTTPCallback
+
 	// The state to resume from as dumped by the PanicErrorHandler.
 	// If this is null, a new Ghostferry run will be started. Otherwise, the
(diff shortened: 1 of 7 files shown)


floriecai:
Can we put the `ErrorCallback` in the generic `config.go` rather than the sharding/copydb specific ones? Then we can also delete the duplicate code in `copydb`.

config.go
```diff
@@ -722,6 +722,8 @@ type Config struct {
 	// in the Progress. This behaviour is perfectly okay and doesn't mean there are no rows being written
 	// to the target DB.
 	EnableRowBatchSize bool
+
+	ErrorCallback HTTPCallback
```
shuhaowu, line 726: Can you put this near the other callbacks and document it?

copydb/copydb.go
```diff
@@ -21,6 +21,12 @@ func NewFerry(config *Config) *CopydbFerry {
 		Config: config.Config,
 	}
 
+	ferry.ErrorHandler = &ghostferry.PanicErrorHandler{
+		Ferry:             ferry,
+		ErrorCallback:     config.ErrorCallback,
+		DumpStateToStdout: false,
+	}
```
shuhaowu, line 28: if the `ErrorCallback` config is in the ghostferry/ferry.go file, and the only place where we use it in the copydb package?  This doesn't seem right. If an error callback is provided to the base ghostferry library, it should just use it.

floriecai (approved):

sharding/sharding.go
```diff
@@ -62,12 +62,6 @@ func NewFerry(config *Config) (*ShardingFerry, error) {
 
 	logger := logrus.WithField("tag", "sharding")
 
-	ferry.ErrorHandler = &ghostferry.PanicErrorHandler{
-		Ferry:             ferry,
-		ErrorCallback:     config.ErrorCallback,
-		DumpStateToStdout: false,
```
shuhaowu, line 68: Removing this means that now sharding will dump state to stdout, as that's a default value inside Ghostferry. We should either preserve the previous behaviour, or make this configurable.

PR #272, opened 2021-04-06
Enable Cutover callback in config for copydb

floriecai:
Refactoring the code so the shared config between copydb and sharding allow `CutoverLock` and `CutoverUnlock` callbacks.

### DIFF
--- a/config.go
+++ b/config.go
@@ -618,9 +618,22 @@ type Config struct {
 	StateCallback        HTTPCallback
 	StateReportFrequency int
 
+
 	// Report error via an HTTP callback. The Payload field will contain the ErrorType,
 	// ErrorMessage and the StateDump.
 	ErrorCallback HTTPCallback
+
+	// Report when ghostferry is entering cutover
+	CutoverLock   			HTTPCallback
+
+	// Report when ghostferry is finished cutover
+	CutoverUnlock 			HTTPCallback
+
+	// These two values configure the amount of times Ferry should attempt to
+	// retry acquiring the cutover lock, and for how long the Ferry should wait
+	// before attempting another lock acquisition
+	MaxCutoverRetries       int
+	CutoverRetryWaitSeconds int
 
 	// The state to resume from as dumped by the PanicErrorHandler.
 	// If this is null, a new Ghostferry run will be started. Otherwise, the
@@ -798,6 +811,14 @@ func (c *Config) ValidateConfig() error {
 
 	if c.WebBasedir == "" {
 		c.WebBasedir = "."
+	}
+
+	if c.MaxCutoverRetries == 0 {
+		c.MaxCutoverRetries = 1
+	}
+
+	if c.CutoverRetryWaitSeconds == 0 {
+		c.CutoverRetryWaitSeconds = 1
 	}
 
 	return nil
--- a/config.go
+++ b/config.go
@@ -629,9 +629,11 @@ type Config struct {
 	// Report when ghostferry is finished cutover
 	CutoverUnlock 			HTTPCallback
 
-	// These two values configure the amount of times Ferry should attempt to
+	// If the callback returns a non OK status, these two values configure the number of times Ferry should attempt to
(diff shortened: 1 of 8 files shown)


config.go
```diff
@@ -618,10 +618,23 @@ type Config struct {
...
 
+	// Report when ghostferry is entering cutover
+	CutoverLock   			HTTPCallback
+
+	// Report when ghostferry is finished cutover
+	CutoverUnlock 			HTTPCallback
+
+	// These two values configure the amount of times Ferry should attempt to
+	// retry acquiring the cutover lock, and for how long the Ferry should wait
+	// before attempting another lock acquisition
```
shuhaowu, line 634: Probably should add:
MaxCutoverRetries Default = 1
CutoverRetryWaitSeconds Second Default ...

config.go
```diff
@@ -618,10 +618,23 @@ type Config struct {
...
 	ErrorCallback HTTPCallback
 
+	// Report when ghostferry is entering cutover
+	CutoverLock   			HTTPCallback
+
+	// Report when ghostferry is finished cutover
+	CutoverUnlock 			HTTPCallback
+
+	// These two values configure the amount of times Ferry should attempt to
+	// retry acquiring the cutover lock, and for how long the Ferry should wait
```
shuhaowu, line 633: Probably should add before:

If the server for CutoverLock returns a non OK status, then these two values ... <carry on>

copydb/cmd/main.go
```diff
@@ -121,7 +120,7 @@ func main() {
 		return
 	}
 
-	if resumeState == nil {
+	if config.StateToResumeFrom == nil {
```
floriecai, line 123: `resumeState` was previously only set if we passed in a path for the statedump file. We want to be able to pass in the state dump directly in the config through the field `StateToResumeFrom` as well.

copydb/copydb.go
```diff
@@ -100,6 +100,7 @@ func (this *CopydbFerry) Run() {
 	// binlog streamer catching up.
 	this.Ferry.WaitUntilBinlogStreamerCatchesUp()
 
+	cutoverStart := this.Ferry.StartCutover()
```
shuhaowu, line 103: Can you comment on what this function is supposed to do? Like:


> Optionally (configurable) post to an HTTP endpoint telling that service Ghostferry is ready for cutover. The external service can then perform steps needed immediately prior to cutover. For example, an external service can set database to be readonly when the this endpoint is invoked.

copydb/copydb.go
```diff
@@ -112,6 +113,7 @@ func (this *CopydbFerry) Run() {
 
 	this.Ferry.StopTargetVerifier()
 
+	this.Ferry.EndCutover(cutoverStart)
```
shuhaowu, line 116: Similar, I would add a comment. This flow is already a bit confusing, adding more without a comment will likely add to confusion.

shuhaowu (approved):
Minor comment changes requested, other than that this looks good.

Not relevant for now, but for the future: we want to consider rewording some of the steps in cutover (and thus the function and variable names), as the list is long and confusing. For example, if `AutomaticCutover` is false, Ghostferry does not notify  via callback that copy is complete and cutover can occur anytime. The current `cutover lock` callback occurs immediately prior to flushing the binlog streamer and cutting over, which may be too late for certain applications.

copydb/cmd/main.go
```diff
@@ -99,6 +78,26 @@ func main() {
...
+		}
+
+		resumeState := &ghostferry.SerializableState{}
+		parser := json.NewDecoder(f)
+		err = parser.Decode(&resumeState)
+		if err != nil {
+			errorAndExit(fmt.Sprintf("failed to parse state file: %v", err))
+		}
+
+		config.StateToResumeFrom = resumeState
```
shuhaowu, line 98: This will overwrite if the state to resume is specified twice. I suppose that's ok?

floriecai: yeah, state in filepath always wins.

PR #274, opened 2021-04-09
Binlog writer lag and target binlog streamer lag 

shuhaowu:
The binlog writer and binlog streamer lag can diverge. This allows us to track them individually in the Progress struct. 

Furthermore, the target binlog streamer lag is a thing when the TargetVerifier is active, this allows us to track that lag when it is in use.

Also ran go fmt in a few places. We should probably setup CI to catch format issues.

### DIFF
--- a/binlog_writer.go
+++ b/binlog_writer.go
@@ -2,6 +2,7 @@ package ghostferry
 
 import (
 	"fmt"
+	"time"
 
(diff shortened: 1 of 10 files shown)


xliang6:
This looks good overall. 👍

shuhaowu:
I tested this and it works. would appreciate a review

xliang6 (approved):

PR #277, opened 2021-04-28
Move CI to Github actions (tests only)

shuhaowu:
Haven't figured out how to build the documentation as well as building and publishing the packages yet, but this is a first step.

One note is that on Github actions, I found that I'm getting a lot of `mysql2::Error: Incorrect arguments to mysqld_stmt_execute`. This error appears to happen if I have a `nil` value for the PK column in the seed_random_data function's prepared statement. Not sure why this fails on Github actions but works locally, I suspect it may be related to one of the following problems, but I can't confirm because I'm not sure how to debug in Github actions yet:

- go-sql-driver/mysql#209
- launchbadge/sqlx#125
- https://stackoverflow.com/questions/16726752/sending-null-in-a-prepared-statement-for-mysql-doesnt-seem-to-work?rq=1

Packaging and documentation building will be done in a follow up PR.


tiwilliam (approved):
Nice work!

PR #283, opened 2021-05-12
report error to the callback if ferry initialization fails in copydb

Manan007224:
During the initial setup for the `ferry` in copydb we were not reporting errors to error callback. Before the ferry runs we are doing a bunch of initialization work which is prone to fail but we were simply exiting from copydb and not reporting these errors to error callback-server. This PR fixes that.

### DIFF
--- a/copydb/cmd/main.go
+++ b/copydb/cmd/main.go
@@ -112,6 +112,7 @@ func main() {
 
 	err = ferry.Start()
 	if err != nil {
+		ferry.Ferry.ErrorHandler.ReportError("ferry.start", err)
 		errorAndExit(fmt.Sprintf("failed to start ferry: %v", err))
 	}
 
@@ -123,6 +124,7 @@ func main() {
 	if config.StateToResumeFrom == nil {
 		err = ferry.CreateDatabasesAndTables()
 		if err != nil {
+			ferry.Ferry.ErrorHandler.ReportError("ferry.createDatabasesAndTables", err)
 			errorAndExit(fmt.Sprintf("failed to create databases and tables: %v", err))
 		}
 	}
--- a/copydb/cmd/main.go
+++ b/copydb/cmd/main.go
@@ -106,6 +106,8 @@ func main() {
 
 	err = ferry.Initialize()
 	if err != nil {
+		// This is not a good idea to reach deep within ferry from copydb. The entire ErrorHandler needs
+		// refactoring which is defined in this issue - https://github.com/Shopify/ghostferry/issues/284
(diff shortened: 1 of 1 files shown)


floriecai (approved):

copydb/cmd/main.go
```diff
@@ -106,11 +106,13 @@ func main() {
 
 	err = ferry.Initialize()
 	if err != nil {
+		ferry.Ferry.ErrorHandler.ReportError("ferry.initialize", err)
 		errorAndExit(fmt.Sprintf("failed to initialize ferry: %v", err))
 	}
 
 	err = ferry.Start()
 	if err != nil {
+		ferry.Ferry.ErrorHandler.ReportError("ferry.start", err)
```
Manan007224, line 115: It felt awkard to name this error `ferry.initialize` as well. Lemme know what you folks think.

Manan007224:
The tests are failiing due to legit errors, but can't seem to figure how the code changes in this PR are affecting the failed tests.

copydb/cmd/main.go
```diff
@@ -106,11 +106,13 @@ func main() {
 
 	err = ferry.Initialize()
 	if err != nil {
+		ferry.Ferry.ErrorHandler.ReportError("ferry.initialize", err)
```
floriecai, line 109: `ferry.ErrorHandler` gets initialized part way through in `ferry.Initialize` but there are several errors that could happen before it is initialized (ex.[ can't connect to source](https://github.com/Shopify/ghostferry/blob/b7ed73a3bcbeed243a92ed3b961d960e92afa309/ferry.go#L336-L340)). https://github.com/Shopify/ghostferry/blob/b7ed73a3bcbeed243a92ed3b961d960e92afa309/ferry.go#L425-L431 


We should initialize the error handler before to avoid this nil pointer exception.

shuhaowu (approved):
I'm tentatively approving this to unblock. That said, I'm not a fan of this code at all. Reaching deep into `ghostferry.Ferry` from the outside is probably not a good idea. That said, it's not this PR's fault that the `ErrorHandler` is not well designed.

We should maybe log and issue about refactoring the ErrorHandler, and leave a todo inline with the code here to that issue...

copydb/cmd/main.go
Manan007224: shuhao made a fix for this - https://github.com/Shopify/ghostferry/pull/276. I think the ghostferry version you have is an old one, pulling the latest master you will the changes.

floriecai: 🤦 I approved that PR too.

floriecai (approved):

PR #286, opened 2021-06-01
add AllowSuperUserOnReadOnly flag

Manan007224:
Add `AllowSuperUserOnReadOnly` so that ghostferry user can write to the target DB even if it is read_only.

### DIFF
--- a/config.go
+++ b/config.go
@@ -753,6 +753,17 @@ type Config struct {
 	// in the Progress. This behaviour is perfectly okay and doesn't mean there are no rows being written
 	// to the target DB.
 	EnableRowBatchSize bool
+
+	// If the target DB is set to read_only ghostferry will throw an error during the initialization step.
+	// AllowSuperUserOnReadOnly flag allows to run ghostferry even if the target DB is read_only. This is helpful in
+	// scenarios where target DB needs to be restricted from writes made by any other user then the ghostferry user.
+	//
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -753,6 +753,17 @@ type Config struct {
 	// in the Progress. This behaviour is perfectly okay and doesn't mean there are no rows being written
 	// to the target DB.
 	EnableRowBatchSize bool
+
+	// Currently if the target DB is set to read_only ghostferry will throw an error during the initialization step.
```
shuhaowu, line 757:
```suggestion
	// If the target DB is set to read_only ghostferry will throw an error during the initialization step.
```

ferry.go
```diff
@@ -379,13 +379,15 @@ func (f *Ferry) Initialize() (err error) {
...
 	}
 
-	isReplica, err := CheckDbIsAReplica(f.TargetDB)
-	if err != nil {
-		f.logger.WithError(err).Error("cannot check if target db is writable")
-		return err
-	}
-	if isReplica {
-		return fmt.Errorf("@@read_only must be OFF on target db")
+	if !(f.Config.AllowSuperUserOnReadOnly) {
```
shuhaowu, line 382: You don't need to `()`? Can you run go fmt?

shuhaowu (approved):
Looks good other than some cosmetic changes.

Issue #285, opened 2021-05-20
Data corruption with composite PK and non-unique ID column with ghostferry-sharding only


shuhaowu:
This is a very specific corner case, but the corruption is caught by the inlineverifier without actually corrupting data for real. I'm not sure why it is the case for now, but I can reliably reproduce it.

To reproduce it yourself, clone this repo, checkout the branch `sharding-data-corruption-with-pagination-key-not-ordered`, then:

```
$ cd examples/sharding
$ ./bugreport.sh
```

This will run you through an interactive session on bug, which should give you approximately the following output:

```
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
ghostferry_mysql-1_1 is up-to-date
ghostferry_mysql-3_1 is up-to-date
ghostferry_mysql-2_1 is up-to-date
Here's a composite PK table
*************************** 1. row ***************************
       Table: t
Create Table: CREATE TABLE `t` (
  `tenant_id` bigint(20) NOT NULL,
  `col1` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `d` datetime(6) NOT NULL,
  PRIMARY KEY (`tenant_id`,`col1`),
  KEY `id` (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
Press enter to continue...

It has two rows in it. We can SELECT * FROM abc.t:
+-----------+------+----+----------------------------+
| tenant_id | col1 | id | d                          |
+-----------+------+----+----------------------------+
|         1 | a    |  2 | 2021-05-20 21:25:53.000000 |
|         1 | z    |  1 | 2021-05-20 21:25:53.000000 |
+-----------+------+----+----------------------------+
Notice how the id column are in reverse order in the above case, likely because the order is in PK order, which is composite on (tenant_id, col1).

ghostferry-sharding built with ghostferry 1.1.0+20210520205254+43a8894
will move tenant tenant_id=1
INFO[0000] hello world from 1.1.0+20210520205254+43a8894  tag=ferry
[...]

PR #287, opened 2021-06-07
Corruption when using FallbackColumn in sharding

tiwilliam:
Resolving #285.

After optimizing `ShardedCopyFilter.BuildSelect` [a while back](https://github.com/Shopify/ghostferry/commit/f548623f1384b0ff08faef73306914cba2561b38) we switched to using [a sub-select](https://github.com/Shopify/ghostferry/blob/9394cbf6114555ed635ed35f9dc0874d40773fd8/sharding/filter.go#L114-L117).

This new select works fine in most cases, as long as you don't try to paginate using something other than the primary key. When trying to paginate using a non-primary key it will return columns in an order which is different from the table schema. Ghostferry have before this bug was detected assumed that data is read and written in the table structure column order. But since `USING(FallbackColumn)` will put `FallbackColumn` first in the result set, it will confuse Ghostferry and lose track of which value needs to go where since the batch writer only have access to the table structure column order.

This PR resolves this by passing the column order in the `RowBatch` struct. Making it possible for the batch writer to generate a correct INSERT query. This allow us to do changes to our queries fetching data without having to be too careful with in which order columns are returned.


.github/workflows/tests.yml
```diff
@@ -7,6 +7,24 @@ on:
   pull_request:
 
 jobs:
+  gh-285:
```
shuhaowu, line 10: thisisfine.jpg 😄

ferry.go
```diff
@@ -38,6 +38,13 @@ const (
 	StateDone                = "done"
 )
 
+func quoteFields(fields []string) (out []string) {
```
shuhaowu, line 41: Minor nit: might want to have this in `utils.go` (along with `quoteField`) as I can see here.

cursor.go
```diff
@@ -261,13 +261,27 @@ func (c *Cursor) Fetch(db SqlPreparer) (batch *RowBatch, paginationKeypos uint64
...
 
+func filterByTableColumns(columnsToFilter []string, tableColumns []schema.TableColumn) (out []string) {
+	exists := make(map[string]struct{}, len(tableColumns))
+	for _, c := range tableColumns {
+		exists[c.Name] = struct{}{}
+	}
+	for _, c := range columnsToFilter {
+		if _, ok := exists[c]; ok {
+			out = append(out, c)
+		}
```
shuhaowu, line 280: Why do we need this filter? Also this seems like the columns coming from the query could be a subset of tables.Columns? Shouldn't it match exactly otherwise we should raise an error?

cursor.go
```diff
@@ -261,13 +261,27 @@ func (c *Cursor) Fetch(db SqlPreparer) (batch *RowBatch, paginationKeypos uint64
 		values:             batchData,
 		paginationKeyIndex: paginationKeyIndex,
 		table:              c.Table,
+		columns:            filterByTableColumns(columns, c.Table.Columns),
```
shuhaowu, line 264: Since this function is not used anywhere else, can we just have the logic inline here (to avoid having to jump around functions and remap the variable names) and also leave a comment above the block of code linking to either this PR or the original reporting issue? It's a bit of non-obvious code that will be hard to understand for people just reading it.

shuhaowu (commented):
A first pass for now. I need to do something else for a few and will come back to this to finish the rest of the review

cursor.go
tiwilliam: It's simply to filter out the MD5 column if fingerprinting is on.

cursor.go
tiwilliam: It's kind of meaty to have in-line and the method interface helps to explain it a bit. It also doesn't have much to do with the bug itself, this is only for filtering out the virtual MD5 column we create for fingerprinting. But I can expand on that in a comment.

row_batch.go
```diff
@@ -10,13 +10,15 @@ type RowBatch struct {
...
 	fingerprints       map[uint64][]byte
+	columns            []string
 }
 
 func NewRowBatch(table *TableSchema, values []RowData, paginationKeyIndex int) *RowBatch {
 	return &RowBatch{
 		values:             values,
 		paginationKeyIndex: paginationKeyIndex,
 		table:              table,
+		columns:            ConvertTableColumnsToStrings(table.Columns),
```
tiwilliam, line 21: Only used in tests.

shuhaowu: Maybe we should move this function into the test only (as I can see, only in `row_batch_test.go`), to not cause additional confusion? Certainly I'm a bit confused when I looked at this code (and maybe you too given that you left a comment here? 😛 )

shuhaowu: Alternatively, we can keep this method, and use this method to sanity check that `columns` and `table.Columns` match up?

cursor.go
```diff
@@ -261,6 +261,7 @@ func (c *Cursor) Fetch(db SqlPreparer) (batch *RowBatch, paginationKeypos uint64
 		values:             batchData,
 		paginationKeyIndex: paginationKeyIndex,
 		table:              c.Table,
+		columns:            columns,
```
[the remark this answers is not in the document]
tiwilliam: I was thinking a bit about it and thought it was not worth the extra compute time since this is a critical flow for performance. Inline verifier would catch if these are different.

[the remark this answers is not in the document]
shuhaowu: I guess I feel pretty paranoid, but you're right the InlineVerifier should catch this so this is probably fine.. 

A note about performance: the RowBatch inserts should be done with hundreds to thousands of rows at the same time. My last measurement of this show that bulk of the time is spent waiting on MySQL as opposed to be within Ghostferry (something like 90%+ time is spent in MySQL). I would imagine checking the columns once per batch is going to be negligible in terms of compute cost.

shuhaowu (approved):
Thanks for the fix!

PR #288, opened 2021-06-15
[Experimental] Migrating data with foreign key constraints

Manan007224:


docs/source/copydbforeignkey.rst
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+- Even though foreign key constraint checks are disabled on target db, table and db creation must happen in a specific order (eg parent should be created
+  before child table). This creation order can be specified by passing ``TablesToBeCreatedFirst`` in the config, or else the table creation order will be 
+  figured out by copydb. 
+
+Limitations
+-------------
+
+- While migrating tables with foreign key constraints the source db should be read_only as there are some fundamental issues when migrating tables with foreign key constraints at the same time when writes are occurring to the source database. This issue descibes briefly why the source database should be read_only during the migration - https://github.com/Shopify/katesql-migration-backend/issues/194.
```
shuhaowu, line 35: You're linking to an internal issue here.

Manan007224: My bad I was tend to forget this is an open source repo. Made a similar issue in ghostferry - https://github.com/Shopify/ghostferry/issues/289 and linked that one.

table_schema_cache.go
```diff
@@ -342,6 +342,75 @@ func (c TableSchemaCache) GetTableListWithPriority(priorityList []string) (prior
...
+		referencedTables, err := getForeignKeyTablesOfTable(db, t[0], t[1])
+		if err != nil {
+			logrus.WithField("table", table).Error("cannot fetch foreign keys")
+		}
+		tableMap[table] = referencedTables
+	}
+
+	for table := range c {
+		if _, found := visitedTables[table]; !found {
+			visitedTables, tableOrder = getTableCreationOrderUtil(visitedTables, tableOrder, tableMap, table)
```
shuhaowu, line 361: Can you document how this algorithm is supposed to work (in either comments attached to a function here, or via a separate document)? It looks feasible to trace through myself, but it's better for a non-trivial algorithm to have its intent documented, so reviewers can focus on if the implementation is right or wrong, not reverse engineering what you're trying to do.

PR #297, opened 2021-07-08
Added binlog events from MariaDB that can be skipped

### LINKED ISSUE #296: MariaDB binlog events differ from MySQL
Running the binlog_streamer from MariaDB seems to be a little bit different.

an error seems to occur:

```
PANI[0001] tried to advance to a zero log position: mysql-bin.047756 0 string  tag=source_binlog_streamer
```

I tracked it down to the following events:
*replication.TableMapEvent, *replication.GenericEvent, *replication.QueryEvent

Which is documented to happen when a replica is unable to understand MariaDB's protocol:
"Dummy event replacing event type X that slave cannot handle."

To me it looks like it comes from the following events:
ANNOTATE_ROWS_EVENT (https://mariadb.com/kb/en/annotate_rows_event/) 
As this is what I sniffed from the traffic:
```
# Dummy event replacing event type 160 that slave cannot handle
```

and the plain text "BEGIN" which both gets shoved in the replication.QueryEvent

Further more the also state that the TABLE_MAP_EVENT is used for row-based binlog:
https://mariadb.com/kb/en/table_map_event/

Currently debugging for more data, will come with a PR that works with my current MariaDB: 10.3.28-MariaDB-1:10.3.28+maria~focal-log

HemeraOne:
resolves #296

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -188,6 +188,7 @@ func (s *BinlogStreamer) Run() {
 
 		isEventPositionResumable := false
 		isEventPositionValid := true
+		skipEvent := false
 
 		switch e := ev.Event.(type) {
 		case *replication.RotateEvent:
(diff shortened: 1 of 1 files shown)


shuhaowu:
> Hope somebody can review this and perhaps say if there's any flaws

I'm not sure if I have enough context to review this code either. Skipping events like this may break interrupt and resume. Without a proper understanding (and good documentation) on why the ev.Pos is 0 and how MariaDB is different from MySQL, I don't really know how to review and merge this safely.

HemeraOne:
Ok after looking again through the code I come to the conclusion I went in the complete wrong direction.
For all events there's a binlog position except for 1 specific:

Each binlog file holds a new special Binlog Event, type 0xa4 (164), called START_ENCRIPTION event:

however documentation also states:
```
it’s written to disk but never sent to connected slave servers.
```

I retrieved it as following:

```
		case *replication.GenericEvent:
			fmt.Println("logPosc:", ev.Header.LogPos)
			fmt.Println("raw type:", ev.RawData[4])
```

logPosf: &{[153 64 231 96 164 60 0 0 0 40 0 0 0 0 0 0 0 128 0 1 1 0 0 0 131 225 116 62 161 1 56 114 104 248 22 224 93 202 121 243] 0xc00009a078 0xc000094090}
raw type: 164

In this case we should be able to ignore binlog event with type 0xa4 (164)

HemeraOne:
Created a bug report for MariaDB at https://jira.mariadb.org/browse/MDEV-26118
Updated the code to only specifically filter this binlog event

HemeraOne:
They have reviewed the ticket and updated the documentation at https://mariadb.com/kb/en/start_encryption_event/
This means we can safely skip this event

PR #300, opened 2021-07-12
299 mediumint unsigned fixed

### LINKED ISSUE #299: unsigned mediumint value through binlog streamer wrongfully parsed
```
table:
CREATE TABLE `customer` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `geo_city` mediumint(8) unsigned DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=7 DEFAULT CHARSET=utf8mb4

queries:
insert into customer values (null, 1);
update customer set geo_city = 8592090 where id = 1; //crashes

error:
ERRO[0009] fatal error detected                          errfrom=binlog_writer error="exec query at pos (mysql-bin.049833, 1049512401) -> (mysql-bin.049833, 1049512401) (163 bytes): Error 1264: Out of range value for column 'geo_city' at row 1" tag=error_handler

ghostferry query:
/*application:ghostferry*/ BEGIN;\n/*application:ghostferry*/ UPDATE `martijntest`.`customer` SET `id`=6,`geo_city`=4286782170 WHERE `id`=6 AND `geo_city`=1;\nCOMMIT
```

so somehow the unsigned value which lies outside of the signed value gets wrongfully converted: 8592090 -> 4286782170 

value parsed from binlog: -8185126

should be converted to int24 which seems to be not a thing in golang

HemeraOne:
fixes #299

### DIFF
--- a/uint24.go
+++ b/uint24.go
@@ -1,0 +1,27 @@
+// with the courtesy of this github file: https://github.com/Cryptkeeper/go-fseq/blob/v0.2.6/pkg/uint24/uint24.go
+// used to convert a int32 to uint24 (mediumint)
+package ghostferry
+
+const MaxUint24 = 1<<24 - 1
+
+func NewUint24(val int32) *Uint24 {
+	var u = new(Uint24)
+	u.Set(val)
+	return u
+}
+
+type Uint24 [3]uint8
+
+func (u *Uint24) Set(val int32) {
+	// panic since this is closer to how a compiler/runtime would treat an overflow compared to err returns
+	if val > MaxUint24 {
+		panic("cannot set Uint24 larger than uint24.MaxUint24")
+	}
+	(*u)[0] = uint8(val & 0xFF)
+	(*u)[1] = uint8((val >> 8) & 0xFF)
+	(*u)[2] = uint8((val >> 16) & 0xFF)
+}
+
+func (u Uint24) Uint32() uint32 {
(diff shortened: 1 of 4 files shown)


shuhaowu:
Can you create a PR with only the fix for mediumint?

HemeraOne:
Done!

shuhaowu:
Thanks for the PR. Is it possible for you to add a test here? https://github.com/Shopify/ghostferry/blob/master/test/integration/types_test.rb

HemeraOne:
I can't get the ruby test thing to work on my machine, so I'm afraid I can't

uint24.go
```diff
@@ -0,0 +1,27 @@
+// with the courtesy of this github file: https://github.com/Cryptkeeper/go-fseq/blob/v0.2.6/pkg/uint24/uint24.go
+// used to convert a int32 to uint24 (mediumint)
+package ghostferry
+
+const MaxUint24 = 1<<24 - 1
+
+func NewUint24(val int32) *Uint24 {
```
shuhaowu, line 7: Is this needed? Can we not just use https://github.com/go-mysql-org/go-mysql/blob/master/mysql/parse_binary.go#L22-L31

HemeraOne: It would take a int32 to byte conversion, which in itself would require some code which you might not want to put in that location as well

PR #301, opened 2021-07-14
299 mediumint unsigned fixed

HemeraOne:
I closed it previously as I encountered issues, however I had a git issue on the machine I was using it on and had resolved the merge conflict in a faulty way. I have done a few separate runs now and it works fine

### DIFF
--- a/dml_events.go
+++ b/dml_events.go
@@ -306,8 +306,10 @@ func NewBinlogDMLEvents(table *TableSchema, ev *replication.BinlogEvent, pos, re
 					row[i] = uint(v)
 				}
 			} else if col.IsUnsigned && col.Type == schema.TYPE_MEDIUM_INT {
-				val := row[i].(int32)
-				row[i] = NewUint24(val).Uint32()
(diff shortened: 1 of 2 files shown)


Manan007224:
@HemeraOne I see in your previous PR you had encountered certain issues regarding running the ruby integration tests locally - https://github.com/Shopify/ghostferry/pull/300#issuecomment-878377949. Looking at the logs it seems that there is an issue in the instally mysql2 gem in your environment. 

If you're using macos, try to uninstall the gem and then again install it using extra flags. 
`sudo gem install mysql2 -v '0.5.3' -- --with-ldflags=-L/usr/local/opt/openssl/lib --with-cppflags=-I/usr/local/opt/openssl/include`

If you're on a linux based ssytem try to search if someone has already encountered this issue before, as this is a know problem. 

Also make sure you're system uses the same ruby version as to what ghostferry integration tests are configured to use which is `2.7.3`.

Lemme know if this problem still persists, then I can write an integration tests and commit to this branch.

PR #303, opened 2021-08-17
Added TotalRows and TotalBytes to the `/status` api for Ghostferry

EtienneBerubeShopify:
Added TotalRows and TotalBytes to the `/status` api for Ghostferry for better time estimations when logging with Splunk.

This included getting the data from the `information_schema` and adding the data into the `StateTracker` so that it could be serialized later to be sent via HTTP.

DEBATE (DEPRECATED):
~~Should it fetch the data even when running from an interrupted state (reading from a serialized state) or should it only query the data when coming from a "clean" run?~~

Notes: 
The Go compiler was complaining about a few things and would not allow me to run any commands before it was happy so I had to fix a few warnings. Here are the files with these "fixes":
- config.go
- iterative_verifier.go

### DIFF
--- a/ferry.go
+++ b/ferry.go
@@ -964,7 +964,11 @@ func (f *Ferry) Progress() *Progress {
 	}
 
 	tables := f.Tables.AsSlice()
-	totalRowsPerTable, totalBytesPerTable := f.GetTotalRowsAndBytesMap()
+	totalRowsPerTable, totalBytesPerTable, err := f.GetTotalRowsAndBytesMap()
+
+	if err != nil {
+		f.logger.Error(err)
+	}
 
 	for _, table := range tables {
 		var currentAction string
@@ -1163,7 +1167,7 @@ func (f *Ferry) checkSourceForeignKeyConstraints() error {
 	return nil
 }
 
-func (f *Ferry) GetTotalRowsAndBytesMap() (totalRowsPerTable map[string]uint64, totalBytesPerTable map[string]uint64) {
+func (f *Ferry) GetTotalRowsAndBytesMap() (totalRowsPerTable map[string]uint64, totalBytesPerTable map[string]uint64, err error) {
 	totalRowsPerTable = make(map[string]uint64)
 	totalBytesPerTable = make(map[string]uint64)
 
@@ -1180,7 +1184,7 @@ func (f *Ferry) GetTotalRowsAndBytesMap() (totalRowsPerTable map[string]uint64, 
 
 		rows, err := f.SourceDB.Query(query)
 		if err != nil {
-			return
+			return totalRowsPerTable, totalBytesPerTable, err
 		}
 		defer rows.Close()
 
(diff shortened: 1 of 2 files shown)


shuhaowu:
> This included getting the data from the information_schema and adding the data into the StateTracker so that it could be serialized later to be sent via HTTP.

StateTracker is used only for critical state that is needed to reconstruct Ghostferry after it is interrupted and resumed. Things that can be discovered should not be included in that struct. See the `Progress` struct.

config.go
```diff
@@ -116,7 +116,7 @@ func (c *DatabaseConfig) MySQLConfig() (*mysql.Config, error) {
 			return nil, fmt.Errorf("failed to build TLS config: %v", err)
 		}
 
-		cfgName := fmt.Sprintf("%s@%s:%s", c.User, c.Host, c.Port)
+		cfgName := fmt.Sprintf("%s@%s:%d", c.User, c.Host, c.Port)
```
Manan007224, line 119: is this change made purposefully ?

ferry.go
```diff
@@ -1158,3 +1161,39 @@ func (f *Ferry) checkSourceForeignKeyConstraints() error {
...
+			table.Schema,
+			table.Name,
+		)
+
+		rows, err := f.SourceDB.Query(query)
+		defer rows.Close()
+
+		var totalRows int
+		var totalBytes int
+
```
Manan007224, line 1185: When there are no rows in the table mysql will return values for `totalRows` and `totalBytes` as `NULL` , and this will create panic in the main process if not handled correctly. 

As a result `totalRows` and `totalBytes` should be `sql.NullInt64`. 

`sql.NullInt64` is a specia type provided by `database/sql` package we use in ghostferry to inteact with mysql. For more information regarding this data type and how NULL values are handled from database rows in go  read this article - https://medium.com/aubergine-solutions/how-i-handled-null-possible-values-from-database-rows-in-golang-521fb0ee267

iterative_verifier.go
```diff
@@ -337,7 +337,7 @@ func (v *IterativeVerifier) reverifyUntilStoreIsSmallEnough(maxIterations int) e
 		before := v.reverifyStore.RowCount
 		start := time.Now()
 
-		_, err := v.verifyStore("reverification_before_cutover", []MetricTag{{"iteration", string(iteration)}})
+		_, err := v.verifyStore("reverification_before_cutover", []MetricTag{{"iteration", fmt.Sprint(iteration)}})
```
Manan007224, line 340: Also is this change made purposefully ?

config.go
EtienneBerubeShopify: Yes, the compiler was complaining and wouldnt compile until this was fixed

iterative_verifier.go
EtienneBerubeShopify: Yes, the compiler was complaining and wouldnt compile until this was fixed

string(int) creates a rune with the ascii / UTF-8 int code

ferry.go
```diff
@@ -1158,3 +1162,49 @@ func (f *Ferry) checkSourceForeignKeyConstraints() error {
...
+		if err != nil {
+			return
+		}
+		defer rows.Close()
+
+		var totalRows sqlorig.NullInt64
+		var totalBytes sqlorig.NullInt64
+
+		for rows.Next() {
+			if err = rows.Scan(&totalRows, &totalBytes); err != nil {
```
Manan007224, line 1191: Is there a reason why you're not returning the error from here and making the stat values as 0 ?

Manan007224: Errors happened while scanning would have nothing to do with issues in connection. Although I get your reasoning on why you're not returning from the error, but just for the sake of consistency with codebase we should return the errors.

EtienneBerubeShopify: Update: Paired with Manan

Manan007224 (approved):

ferry.go
```diff
@@ -963,6 +964,11 @@ func (f *Ferry) Progress() *Progress {
 	}
 
 	tables := f.Tables.AsSlice()
+	totalRowsPerTable, totalBytesPerTable, err := f.GetTotalRowsAndBytesMap()
+
+	if err != nil {
+		f.logger.Error(err)
```
shuhaowu, line 970: If this error occurs, won't line 996 below experience a nil pointer error?

PR #310, opened 2021-09-24
Switch to official Datadog client

coding-chimp:
Switches to the official Datadog client instead of the deprecated one we were using before.

Ghostferry is sometimes spamming our logs because the metrics buffer is full. I think this happens because ghostferry is generating those metrics faster than datadog can consume them. We already tried increasing the buffer in #268 but that didn't really help too much. My hope is that the official client is more efficient, which might mitigate the issue somewhat.


tiwilliam (approved):

shuhaowu (approved):
This is fine, but the dd metric is not _that_ useful, as we can only see aggregate metrics if multiple instances of Ghostferry is running.

Issue #265, opened 2021-03-10
Dynamic reconfiguration


shuhaowu:
We should use the control server to allow certain variables to be changed on the fly:

- Log verbosity
- Batch size
- ~DataIteration concurrency~

There may be other variables too.

shuhaowu:
Can this be closed?

yevsafronov:
yes

PR #316, opened 2021-10-19
Dynamic reconfiguration for DataIterationBatchSize

yevsafronov:
Re: https://github.com/Shopify/ghostferry/issues/265

As part of the above issue I was looking into a way to implement dynamic configuration via the control server.
This PR adds a new endpoint to the `ControlServer` which allows dynamic reconfiguration, currently it only supports updating the `DataIterationBatchSize` value.

Another important change is the move of  `ControlServer` into `Ferry` as an optional component, disabled by default and enabled for copydb. This will allow to use the dynamic reconfiguration and other `ControlServer` endpoints outside of copydb.

### DIFF
--- a/config.go
+++ b/config.go
@@ -284,7 +284,7 @@ func (c *IterativeVerifierConfig) Validate() error {
 
 type ControlServerConfig struct {
 	// enable/disable http control server
-	EnableControlServer bool
+	Enabled bool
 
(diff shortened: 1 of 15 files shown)


shuhaowu (approved):
Looks good. So test failures but not sure if they're real so I restarted it.

PR #319, opened 2021-11-16
added a GET /api/config endpoint to expose updatable config values

yevsafronov:
This PR adds a GET `/api/config` endpot to expose updatable config, this could be used to prepopule UI input prior to update or to validate that values were updated

### DIFF
--- a/control_server.go
+++ b/control_server.go
@@ -105,7 +105,8 @@ func (this *ControlServer) Initialize() (err error) {
 	this.router.HandleFunc("/api/stop", this.HandleStop).Methods("POST")
 	this.router.HandleFunc("/api/verify", this.HandleVerify).Methods("POST")
 	this.router.HandleFunc("/api/script", this.HandleScript).Methods("POST")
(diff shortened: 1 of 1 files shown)


control_server.go
```diff
@@ -368,7 +369,15 @@ func (this *ControlServer) HandleScript(w http.ResponseWriter, r *http.Request)
 	returnSuccess(w, r)
 }
 
-func (this *ControlServer) HandleConfig(w http.ResponseWriter, r *http.Request) {
+func (this *ControlServer) HandleConfigGet(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	err := json.NewEncoder(w).Encode(this.F.UpdatableConfig)
```
shuhaowu, line 374: Just the updatable ones? Or should we return all of it?

yevsafronov: just the updatable, this endpoint is added to improve the usability of the `POST /api/config`

Issue #318, opened 2021-11-03
Ghostferry should abort if it sees a DDL command that can compromise data integrity


shuhaowu:
For the time being, Ghostferry should abort on DDL commands that it sees in the binlog if it cannot handle it safely. For the time being, this includes all schema changes. Other commands need also to be identified on a case by case basis. We should use this issue to discuss which ones we should forbid.

PR #320, opened 2021-11-25
Abort on DDL in binlog stream

insom:
related to https://github.com/Shopify/ghostferry/issues/318

Co-authored-by: Pawan Dubey <pawan.dubey@shopify.com>


binlog_streamer.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/go-mysql-org/go-mysql/replication"
 	"github.com/sirupsen/logrus"
+	"vitess.io/vitess/go/vt/sqlparser"
```
shuhaowu, line 15: Presumably this is where the +143k lines of code comes from..? Generally speaking every time we do a vendor upgrade, I read through all the changes (and even review some of the code) of the dependencies I'm upgrading (only on the most important ones, like the go-mysql library). Will adding vitess as a dependency make it a lot harder as it introduces a lot of noise?

Are there any other "easy" options to do this parsing?

sharding/test/trivial_integration_test.go
```diff
@@ -142,3 +143,38 @@ func TestErrorsIfShardingKeyChanged(t *testing.T) {
...
+
+func (this *InsertDDLWriter) Stop() {}
+
+func (this *InsertDDLWriter) Wait() {}
+
+func (this *InsertDDLWriter) SetDB(db *sql.DB) {
+	this.db = db
+}
+
+func TestErrorsIfDDLDetected(t *testing.T) {
```
shuhaowu, line 163: The Golang integration testing environment is theoretically "deprecated" and we're supposed to use the ruby one, even tho that one is also imperfect. Is there a reason to have the test here as opposed to be in `test/integration`? Does this need to be a sharding-specific test?

binlog_streamer.go
insom: It is no doubt a huge dependency, so I expected this comment. There are some libraries which have extracted the parser (only) out of Vitess, but each of those seems to have stopped development (or alternatively: they are simply complete).

For example https://github.com/blastrain/vitess-sqlparser exposes nearly the same API and is a much smaller dependency.

sharding/test/trivial_integration_test.go
insom: I didn't realise the Golang env is deprecated, I can rework this to be a Ruby test. I actually pinged you for review by mistake as this is still a draft, but this is good feedback, thanks 🙏

shuhaowu:
I'm okay with this dependency for now. I would make the DDL event a callback instead of erroring right here. As you mentioned, we can pass the event based on the filters passed to Ghostferry as well, but we can do that in a follow up PR if you would like.

binlog_streamer.go
```diff
@@ -230,6 +231,22 @@ func (s *BinlogStreamer) Run() {
 			// the full query that was executed on the master (with annotations)
 			// that is otherwise not possible to reconstruct
 			query = ev.Event.(*replication.RowsQueryEvent).Query
+		case *replication.QueryEvent:
+			query = ev.Event.(*replication.QueryEvent).Query
```
shuhaowu, line 235: This is a switch case structure with a type, so I think you can just do `e.Query` here?

binlog_streamer.go
```diff
@@ -230,6 +231,22 @@ func (s *BinlogStreamer) Run() {
...
+			if string(query) == "BEGIN" {
```
shuhaowu, line 236: Can you leave a comment on why this is ignored here? It's not super obvious to me.

binlog_streamer.go
```diff
@@ -230,6 +231,22 @@ func (s *BinlogStreamer) Run() {
...
+				break
+			}
+			statement, err := sqlparser.Parse(string(query))
```
shuhaowu, line 239: Nitpick: I see `string(query)` twice here. Can we modify the above line to simply go: `query := string(e.Query)`?

test/integration/trivial_test.rb
```diff
@@ -23,6 +23,40 @@ def test_copy_data_with_writes_to_source
...
+  end
+
+  def test_copy_data_with_alter_fails_part_way_through
+    seed_simple_database_with_single_table
+
+    datawriter = AlterTableDataWriter.new
+    ghostferry = new_ghostferry(MINIMAL_GHOSTFERRY)
+
+    ghostferry.on_status(GhostferryHelper::Ghostferry::Status::BINLOG_STREAMING_STARTED) do
+      datawriter.start
```
shuhaowu, line 48: The `datawriter` concept is mostly for when you want a background thread hitting MySQL in the test.

In this case, I think you can just go directly to:

```suggestion
      source_db.query(....)
      source_db.query(....)
```

See types_test for more examples of this. This should save you from using another class.

test/integration/trivial_test.rb
```diff
@@ -23,6 +23,40 @@ def test_copy_data_with_writes_to_source
...
+    end
+
+    ghostferry.run_expecting_failure
+
+    source, target = source_and_target_table_metrics
+    source_count = source[DEFAULT_FULL_TABLE_NAME][:row_count]
+    target_count = target[DEFAULT_FULL_TABLE_NAME][:row_count]
+    
+    refute_equal(source_count, target_count, "target should have fewer rows than source")
```
shuhaowu, line 57: `ghostferry.stderr` should also give you the log messages, if you want to match the log message here. I just realized that nothing actually uses it so maybe what you have is fine 🤷

shuhaowu (commented):
Look pretty good. Some minor feedback on the tests and the code. We can merge after they're addressed. if you want, in a follow up, we can turn this into callbacks.

PR #322, opened 2021-12-07
Add handlers for query events

shivnagarajan:
@pawandubey 

Co-authored-by: Pawan Dubey <pawan.dubey@shopify.com>

PR #323, opened 2021-12-08
Detect ddl add handlers

shivnagarajan:
This is a slightly alternate version of 

https://github.com/Shopify/ghostferry/pull/322

@pawandubey


binlog_streamer.go
```diff
@@ -49,6 +55,7 @@ type BinlogStreamer struct {
 
 	logger         *logrus.Entry
 	eventListeners []func([]DMLEvent) error
+	eventHandlers  map[string]func(*replication.BinlogEvent, []byte, *BinlogEventState) ([]byte, error)
```
pawandubey, line 58: Wonder if we can create an enum of "supported" events that you could create handlers for, and then use that enum as the map key. Would be slightly more robust than string keys at the cost of some extra code to maintain.

binlog_streamer.go
```diff
@@ -175,125 +266,47 @@ func (s *BinlogStreamer) Run() {
...
-			}
-		case *replication.RowsEvent:
-			err = s.handleRowsEvent(ev, query)
+		eventTypeString := ev.Header.EventType.String()
+		if handler, ok := s.eventHandlers[eventTypeString]; ok {
+			query, err = handler(ev, query, &es)
 			if err != nil {
-				s.logger.WithError(err).Error("failed to handle rows event")
+				s.logger.WithError(err).Error("failed to handle event")
 				s.ErrorHandler.Fatal("binlog_streamer", err)
```
pawandubey, line 289: We will treat all errors as fatal? I'm fine with that.

shivnagarajan:
This changes the code to provide the ability to add a `binlogEvent` handler for individual events. The `golang` test has been updated to test adding such a handler for a `query event` and validates that it actually works.

Since the code no longer needs to parse the binlog events explicitly, we did not need to pull in the full parser for the ddl events that we had earlier.

binlog_streamer.go
```diff
@@ -174,109 +266,47 @@ func (s *BinlogStreamer) Run() {
...
-			// interruption to EITHER the start (if using GTIDs) or the end of the
-			// last transaction
-			isEventPositionResumable = true
-
-			// Here we also reset the query event as we are either at the beginning
-			// or the end of the current/next transaction. As such, the query will be
-			// reset following the next RowsQueryEvent before the corresponding RowsEvent(s)
-			query = nil
+		} else {
+			query, err = s.defaultEventHandler(ev, query, &es)
```
shuhaowu, line 292: So the `defaultEventHandler` fires events to the `eventListeners`, but if one were to override the `eventHandler`, `eventListeners` will no longer be called? This also means that the default event listener that handles the row event (by sending it to the BinlogWriter) will no longer be called. You would have to manually reregister `handleRowEvent` as a eventHandler? This feels very confusing and somewhat error prone? Are there any cases where we don't want to row events to be handled as well?

sharding/test/trivial_integration_test.go
```diff
@@ -142,3 +155,39 @@ func TestErrorsIfShardingKeyChanged(t *testing.T) {
...
+
+func (this *InsertDDLWriter) Stop() {}
+
+func (this *InsertDDLWriter) Wait() {}
+
+func (this *InsertDDLWriter) SetDB(db *sql.DB) {
+	this.db = db
+}
+
+func TestErrorsIfDDLDetected(t *testing.T) {
```
shuhaowu, line 175: I thought we had a ruby integration test for this? Do we have to use the Golang version instead?

binlog_streamer.go
shivnagarajan: I've cleaned this up a bit, and we have some bounded validations around the type of events you can attach a handler to.

pawandubey (approved):

binlog_streamer.go
```diff
@@ -133,6 +142,90 @@ func (s *BinlogStreamer) ConnectBinlogStreamerToMysqlFrom(startFromBinlogPositio
 	return s.lastStreamedBinlogPosition, err
 }
 
+func (s *BinlogStreamer) defaultEventHandler(ev *replication.BinlogEvent, query []byte, es *BinlogEventState) ([]byte, error) {
```
shuhaowu, line 145: I didn't go through the function of this body very carefully, but I assume it is copied more or less verbatim from the original code (with some minor adjustments to the variable names).

shuhaowu: One thing that is a bit surprising (maybe not?) is that the `es` parameter is being modified, and the "other loop" in the binlog streamer code is relying on that being modified to work (the `isEventResumableVariable`). Thus, to understand how the binlog streamer works works, you have to go into this function and trace the evolution of the parameter `es`, which to me feels a bit more convoluted than before. THAT SAID, in the absence of a bigger refactor, I think this is fine, but may be worth pointing out somewhere in a comment.

Or maybe I'm writing too much C++ these days and the lack of the `const` qualifier is getting to me 😅. Let me know either way.

shuhaowu (approved):
I requested a few minor comments, as I feel the complexity of the code in the BinlogStreamer increasing a bit, but that's likely OK for now.

PR #326, opened 2022-01-07
fix gotestsum arch checks

shivnagarajan:
This makes the download for gotestsum work on the various `arch` cases that we require (arm64 and x86_64)

Co-authored-by: Aaron Brady <aaron.brady@shopify.com>


pawandubey (approved):
Aah, thank you. Much better than what I was trying to do 🙃

PR #327, opened 2022-01-07
Rename lag to lag_value in tests

pawandubey:
`lag` is a keyword in MySQL 8, so using it as a column name creates syntax problems.

E.g. 

This is OK in MySQL 5.7:
```sql
CREATE TABLE foo (lag FLOAT NOT NULL);
```

But in MySQL 8.0+, it spits out this error:
```
Schema Error: Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'lag FLOAT NOT NULL)' at line 1
```

⏯️ [Try it on DB Fiddle](https://www.db-fiddle.com/f/mgiTY55MPjynhbu5aqRKez/1)

----

Ref. [New keywords in MySQL 8](https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-new-8-0-L)

### DIFF
--- a/test/go/lag_throttler_test.go
+++ b/test/go/lag_throttler_test.go
@@ -38,12 +38,12 @@ func setupLagTable(db *sql.DB, ctx context.Context) {
 	_, err = db.Exec("CREATE DATABASE meta")
 	testhelpers.PanicIfError(err)
 
-	_, err = db.Exec("CREATE TABLE meta.lag_table (lag FLOAT NOT NULL, server_id int unsigned NOT NULL PRIMARY KEY)")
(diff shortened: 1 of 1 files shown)


insom:
We also just did a version of this :nomouth: and had to change it in a few more places:

https://github.com/Shopify/ghostferry/commit/ae055aec20d023d371fef9c726222fd52dd4936a#diff-72669bd10f87a5422ac232bc8c315440c9ba0d58c996fdab16efb4628aef05a3R32

Happy for this to go in as a separate PR though as we hone in on MySQL 8.0 compatibility.

pawandubey:
Aah, I think your commit seems more comprehensive. It would make sense to keep all the compatibility changes together. Will close this 🙏🏼

shuhaowu:
You can also enclose it with the back ticks `.

PR #328, opened 2022-01-10
add bits for running arm64 mysql8

shivnagarajan:
Co-authored-by: Aaron Brady <aaron.brady@shopify.com>


docker-compose.yml
```diff
@@ -1,8 +1,7 @@
 version: "3"
 services:
   mysql-1:
-    image: percona:5.7
-    platform: linux/x86_64
+    image: docker.io/mysql/mysql-server:8.0
```
shuhaowu, line 4: Do we not want to test on 5.7 locally anymore? It seems valuable to have that option..?

test/go/race_conditions_integration_test.go
```diff
@@ -25,6 +26,7 @@ func annotate(queries []string, marginalia string) []string {
 }
 
 func TestSelectUpdateBinlogCopy(t *testing.T) {
+	t.Skip("Skipping for MySQL8")
```
shuhaowu, line 29: This is a pretty important test as it covers one of the Ghostferry race conditions that it is designed to handle. Although I think this is already replicated in the Ruby test suite?

testhelpers/helper_methods.go
```diff
@@ -30,7 +31,9 @@ func ProcessListContainsQueries(db *sql.DB, queries []string) bool {
 	}
 
 	for rows.Next() {
-		data, err := ghostferry.ScanGenericRow(rows, 10)
+		//columns, _ := rows.Columns()
+		//data, err := ghostferry.ScanGenericRow(rows, len(columns))
+		data, err := ghostferry.ScanGenericRow(rows, 8)
```
shuhaowu, line 36: I assume this is WIP?

vendor/github.com/go-mysql-org/go-mysql/schema/schema.go
```diff
@@ -354,6 +354,8 @@ func (ta *Table) fetchIndexesViaSqlDB(conn *sql.DB) error {
 			&unused,
 			&unused,
 			&unused,
+			&unused,
+			&unused,
```
shuhaowu, line 358: What happened here? Did we update go.mod?

shuhaowu (commented):
I assume this is a pretty WIP PR? Should we label it as a Draft?

vendor/github.com/go-mysql-org/go-mysql/schema/schema.go
shivnagarajan: This is something that needed an upstream fix https://github.com/go-mysql-org/go-mysql/pull/527. Since that has merged, I'll pull in those changes instead.

.github/workflows/start-mysql.sh
```diff
@@ -18,7 +18,7 @@ wait_for_mysql() {
   port=$1
   echo "Waiting for MySQL at port $port..."
   attempts=0
-  while ! nc -w 1 localhost $port | grep -q "mysql"; do
+  while nc -vvw 1 localhost $port | grep -q "succeeded"; do
```
shivnagarajan, line 21: mysql 8 has a differently formatted response (at least when using the `mysql/mysql-server` version

.github/workflows/start-mysql.sh
```diff
@@ -29,7 +29,41 @@ wait_for_mysql() {
   echo "MySQL at port $port has started!"
 }
 
+wait_for_version () {
```
shivnagarajan, line 32: We need to check for two different states

1. can we connect and retrieve the correct version
2. can we validate that the health check succeeds.

pawandubey (approved):
Tested locally and looks good to me. Have one small suggestion.

(also perhaps there's scope for squashing some of the commits together?)

test/go/lag_throttler_test.go
```diff
@@ -141,7 +142,8 @@ func TestThrottlerRunErrors(t *testing.T) {
 
 	err = throttler.Run(ctx)
 	assert.NotNil(t, err)
-	assert.Equal(t, "Error 1146: Table 'meta.lag_table' doesn't exist", err.Error())
+	validError := (err.Error() == "Error 1146: Table 'meta.lag_table' doesn't exist") || (err.Error() == "Error 1049: Unknown database 'meta'")
```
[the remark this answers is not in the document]
shivnagarajan: Yeah, on MySQL 8 the error was unknown database, as opposed to table does not exist. On 5.7, the  error was table does not exist.

dev.yml
```diff
@@ -20,11 +23,11 @@ up:
...
+      met?: docker-compose -f docker-compose_8.0.yml up -d mysql-1 mysql-2
       meet: echo 'mysql failed to start'; false
-      down: docker-compose stop mysql-1 mysql-2
+      down: docker-compose -f docker-compose_8.0.yml stop mysql-1 mysql-2
 
 commands:
   test:
     desc: Run the test suite.
-    run: make test
+    run: MYSQL_VERSION="8.0" make test
```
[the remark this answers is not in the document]
shivnagarajan: I'll fix this.

docker-compose.yml
shivnagarajan: We are testing 5.7 in the github workflows as a secondary CI pipeline.

.github/workflows/start-mysql.sh
shivnagarajan: This has been cleaned up now.

vendor/github.com/go-mysql-org/go-mysql/schema/schema.go
shivnagarajan: This re-done and the updated version from https://github.com/go-mysql-org/go-mysql/pull/658 was pulled in instead.

test/go/race_conditions_integration_test.go
shivnagarajan: We've fixed this test, so no need to skip it any more.

insom (approved):
I did one more 🎩 of this on an M1 and it looks good 👍 -- I think the two remaining things for review are: this adds a second CI test are we okay with that? (I am, although it increases the change of hitting a flaky test) and skipping the MB4->MB3 test for MySQL 8.0 as it's no longer valid to silently convert data that way in MySQL 8 (it was always a bad idea)

test/go/lag_throttler_test.go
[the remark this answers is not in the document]
shivnagarajan: I can split this out for the two different cases we care about (MYSQL8 and MYSQL 5.7)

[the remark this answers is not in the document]
shuhaowu: That would be a bit more clear, yea. Alternatively you can just inline here do the following if it saves some time and characters:

```
if 57 {
...
} else if 80 {
}
```

dev.yml
```diff
@@ -20,9 +26,9 @@ up:
       meet: echo 'go mod failed to download dependencies'; false
   - custom:
       name: MySQL
-      met?: docker-compose up -d mysql-1 mysql-2
+      met?: docker-compose -f docker-compose_8.0.yml up -d mysql-1 mysql-2
       meet: echo 'mysql failed to start'; false
-      down: docker-compose stop mysql-1 mysql-2
+      down: docker-compose -f docker-compose_8.0.yml stop mysql-1 mysql-2
```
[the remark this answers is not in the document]
shuhaowu: I suppose for the OS X development, 8.0 is probably fine given that we don't have a 5.7 ARM64 image. For Linux development, typing `docker-compose up -d` will still get you a 5.7 environment.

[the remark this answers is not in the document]
shivnagarajan: yup.. that's the idea and the `mysql8` env var will not be set. This is how the various github workflows run for 5.7 and 8 also

dml_events.go
```diff
@@ -423,6 +423,9 @@ func appendEscapedValue(buffer []byte, value interface{}, column schema.TableCol
 
 	switch v := value.(type) {
 	case string:
+		if column.Type == schema.TYPE_JSON {
+			return appendEscapedBuffer(buffer, []byte(v), true)
+		}
```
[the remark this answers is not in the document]
shuhaowu: Sounds good! Also it seems like that link jumped me to some random line of code that doesn't seem related (`func (_ *testDecodeSuite) TestTableMapHelperMaps(c *C) {`). I suspect Github's linking feature is currently broken? 

In any case, let me know when you add the comment and I can give a final review on this as everything else looks good.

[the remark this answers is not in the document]
insom: I was wrong! NULL types are handled above by `isNilValue` so there's no situation where we get a `[]byte` back from go-mysql and it's a JSON column. I've unconditionally changed that code to pass `false` for `isJSON` and added a comment.

shuhaowu (approved):
Thanks @insom for the deep dives and concise explanation of all the intricacies present here, and thanks everyone who worked on this PR to make everything work for MySQL 8 and arm64 ❤️. 

Sorry that the review process took this long, but this all looks good to me.

PR #336, opened 2022-11-03
Log checksum in case of inline verification failure

shuhaowu:
This logs the checksum of the failed rows, so we can better debug what happened.

### DIFF
--- a/inline_verifier.go
+++ b/inline_verifier.go
@@ -232,6 +232,12 @@ func (s *BinlogVerifyStore) Serialize() BinlogVerifySerializedStore {
 	return s.store.Copy()
 }
 
+type InlineVerifierMismatches struct {
+	Pk             uint64
+	SourceChecksum string
+	TargetChecksum string
+}
+
 type InlineVerifier struct {
 	SourceDB                   *sql.DB
 	TargetDB                   *sql.DB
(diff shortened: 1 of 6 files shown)


xliang6 (approved):

inline_verifier.go
```diff
@@ -613,17 +631,21 @@ func (v *InlineVerifier) compareDecompressedData(source, target map[uint64]map[s
...
-func (v *InlineVerifier) compareHashesAndData(sourceHashes, targetHashes map[uint64][]byte, sourceData, targetData map[uint64]map[string][]byte) []uint64 {
+func (v *InlineVerifier) compareHashesAndData(sourceHashes, targetHashes map[uint64][]byte, sourceData, targetData map[uint64]map[string][]byte) []InlineVerifierMismatches {
 	mismatches := v.compareHashes(sourceHashes, targetHashes)
 	compressedMismatch := v.compareDecompressedData(sourceData, targetData)
 	for paginationKey, _ := range compressedMismatch {
-		mismatches[paginationKey] = struct{}{}
+		mismatches[paginationKey] = InlineVerifierMismatches{
+			Pk:             paginationKey,
+			SourceChecksum: "compressed-data-mismatch", // TODO: compute the hash of the compressed data and put it here
+			TargetChecksum: "compressed-data-mismatch",
```
saunderst, line 641: Just wondering why ship this without the computed compressed data hashes.  Does it just seem much more likely to hit `compareHashes` mismatches?

saunderst (approved):
LGTM but this is my first time looking at ghostferry.

inline_verifier.go
shuhaowu: It's slightly complex to calculate this for now. Where we are observing this is also not associated with this code path. So for now we'll just ignore it.

PR #346, opened 2023-10-18
Update go to v1.21

driv3r:
- update github actions
- tidy go mod and vendor


.github/workflows/release.yml
```diff
@@ -9,12 +9,12 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - name: Setup Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v4
         with:
-          go-version: 1.16
+          go-version: '1.21.3'
```
Sinclert, line 17: This question applies to all the GHA workflows:

Do we need to be that explicit about the version? I would say stating `1.21` would be enough, regardless of the _patch_ version. WDYT?

milanatshopify (approved):

PR #368, opened 2024-11-19
Patch go-mysql to keep trailing zero in floats in json

mtaner:
Patched json binary decoder from go-mysql library to add a custom marshal json method for the floats to keep the trailing 0. 

This patch only works with mysql version 8, as mysql version 5.7 gets rid of the trailing 0 on `cast...as json`
Screenshot 2024-11-21 at 12 36 53. 

This means that the tests added for checking that the trailing 0 is not dropped will fail for the 5.7 ruby tests so I have updated the test to ignore trailing 0 data for mysql version 5.7

Next steps are the open a PR with go-mysql highlighting that this will only work with mysql version 8 onwards (and add settings option similar to `useDecimal` to `BinlogSyncerConfig` and see what they think.

### DIFF
--- a/vendor/github.com/go-mysql-org/go-mysql/replication/json_binary.go
+++ b/vendor/github.com/go-mysql-org/go-mysql/replication/json_binary.go
@@ -49,12 +49,9 @@ type FloatWithTrailingZero float64
 
 func (f FloatWithTrailingZero) MarshalJSON() ([]byte, error) {
 	if float64(f) == float64(int(f)) {
-		fmt.Println("MarshalJSON", float64(f), int(f))
-		fmt.Println("number", strconv.FormatFloat(float64(f), 'f', 1, 32))
 		return []byte(strconv.FormatFloat(float64(f), 'f', 1, 32)), nil
 	}
 
-	fmt.Println("number2", strconv.FormatFloat(float64(f), 'f', -1, 32))
 	return []byte(strconv.FormatFloat(float64(f), 'f', -1, 32)), nil
 }
 


driv3r (commented):
lets add a test case with something after decimal, just for completes, i.e. `52.19`

vendor/github.com/go-mysql-org/go-mysql/replication/json_binary.go
```diff
@@ -44,6 +45,19 @@ const (
 	jsonbValueEntrySizeLarge = 1 + jsonbLargeOffsetSize
 )
 
+type FloatWithTrailingZero float64
+
+func (f FloatWithTrailingZero) MarshalJSON() ([]byte, error) {
+	if float64(f) == float64(int(f)) {
+		fmt.Println("MarshalJSON", float64(f), int(f))
+		fmt.Println("number", strconv.FormatFloat(float64(f), 'f', 1, 32))
```
driv3r, line 53: I guess we should remove debugging `Println`s before merge :P

driv3r (approved):

test/integration/types_test.rb
```diff
@@ -103,28 +106,33 @@ def test_json_data_insert
     # with a JSON column is broken on 5.7.
     # See: https://bugs.mysql.com/bug.php?id=87847
     res = target_db.query("SELECT COUNT(*) AS cnt FROM #{DEFAULT_FULL_TABLE_NAME}")
-    assert_equal 16, res.first["cnt"]
+    assert_equal ENV["MYSQL_VERSION"] == "8.0" ? 22 : 18, res.first["cnt"] # 22 rows with trailing zero data, 18 without trailing zero
```
yongli-shopify, line 109: Is there any other better way? checking and counting manually sounds not like a sustainable/maintainable solution here

ilikeorangutans (approved):

test/integration/types_test.rb
mtaner: https://github.com/Shopify/ghostferry/pull/368/commits/98d4b922d81770a521f84816898fa91e7fb5d101 - changes here I think its better?

PR #375, opened 2024-12-06
Reorganise Ruby tests

grodowski:
Reorganise to run minitests with `rake` and enable a few neat things:
- run a single test file, thanks to the absence of `main.rb`
- `ruby-lsp` goodness in Code: run/debug single test and test explorer

Screenshot 2024-12-06 at 11 24 40

FWIW I wanted to do the same for golang, but the current Go extension for Code requires `1.19+`, so a `go` update may be due in a separate PR. Any objections to do that?


driv3r:
carve out logrus before updating golang please

.gitignore
```diff
@@ -17,7 +17,6 @@ bin/
 .vagrant/
 build/
 .bundle/
-.vscode/
```
driv3r, line 20: why do we remove this from git ignore?

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+    "version": "0.2.0",
+    "configurations": [
+      {
+        "name": "Minitest - current file",
+        "type": "ruby_lsp",
+        "request": "launch",
+        "program": "rake test TEST=${file}"
+      }
+    ]
+  }
```
driv3r, line 14: please lets avoid any unnecessary IDE related staff out

driv3r (changes requested):

.vscode/launch.json
grodowski: Good point, I added these only before finding out that adding `rake` might make sense, and now `ruby-lsp` runs tests just fine. Removing!

test/integration/trivial_test.rb
```diff
@@ -1,6 +1,6 @@
 require "test_helper"
 
-class TrivialIntegrationTests < GhostferryTestCase
+class TrivialIntegrationTest < GhostferryTestCase
```
grodowski, line 3: TIL: `ruby-lsp` doesn't recognise test classes with plural names

driv3r (approved):

PR #380, opened 2025-02-12
Manage index hinting for sharding data iterator query

mtaner:
Recently we have discovered that `USE INDEX` can slow the data iterator batching queries down as MySQL can decide to do a table scan instead, where as not using it allows the optimizer to select the correct index & avoid table scan. `FORCE INDEX` prevents MySQL from doing a table scan also but it removes the flexibility of selecting another index for the sharding key that might be more suitable. Another theory is that query optimizer is better in MySQL 8 and therefore we might not need to use "USE INDEX" to ensure MySQL is using the correct index with this version (index hinting was originally introduced for optimizer selecting the wrong index with 5.7 for some tables.)

This is why I am introducing a configuration to manage the index hinting for the sharded data iterator. 

```
IndexHint // "use" by default. Options "none", "use", "force"

IndexHintingPerTable // has higher specificity then higher level IndexHint if specified

example:
IndexHintingPerTable: { 
   "blog": { // schema name
      "users": { // table name
	 "IndexHint": "force" // has higher level specificity then higher level IndexHint
	 "IndexName": "ix_users_some_id" // defaults to empty string, will override the index chosen by Ghostferry
	},
      "accounts": {
          "IndexName": "ix_accounts_some_id" // indexName will be ignored if it doesn't exist on the table
          // will inherit the higher level IndexHint as not specified
     }
 }
 ```

### DIFF
--- a/sharding/config.go
+++ b/sharding/config.go
@@ -2,6 +2,7 @@ package sharding
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/Shopify/ghostferry"
 )
@@ -29,6 +30,9 @@ type Config struct {
 	PrimaryKeyTables []string
 
 	Throttle *ghostferry.LagThrottlerConfig
+
+	// ShardedCopyFilterConfig is used to configure the sharded copy filter query.
+	ShardedCopyFilterConfig *ShardedCopyFilterConfig
 }
 
 func (c *Config) ValidateConfig() error {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: sharding/config.go [hunk 1]
```diff
@@ -44,3 +47,44 @@ func (c *Config) ValidateConfig() error {
...
+	// IndexHintingPerTable: {
+	//   "blog": {
+	//     "users": {
+	//       "DisableIndexHint": false,
+	//       "UseForceIndex": true,
+	//       "IndexName": "ix_users_some_id"
+	//     }
+	//   }
+	// }
+	IndexHintingPerTable map[string]map[string]IndexConfigPerTable
```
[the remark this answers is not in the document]
mtaner: Had an offline discussion with @coding-chimp and decided to go with `IndexHint` to combine both options: "none", "use", "force". This way the lower & higher level management of the settings becomes easier too.

sharding/filter.go
```diff
@@ -117,8 +138,14 @@ func (f *ShardedCopyFilter) BuildSelect(columns []string, table *ghostferry.Tabl
...
-	if indexName := f.shardingKeyIndexName(table); indexName != "" {
+func (f *ShardedCopyFilter) shardingKeyIndexHint(table *ghostferry.TableSchema, indexHint string) string {
+	indexName := f.shardingKeyIndexName(table)
+
+	if indexName != "" {
+		if indexHint == "force" {
+			return "FORCE INDEX (`" + indexName + "`)"
+		}
+
 		return "USE INDEX (`" + indexName + "`)"
```
mtaner, line 149: by this point, the validation should have caught "invalid" IndexHint options but for some reason it doesn't we always default to the existing behaviour with "USE INDEX"

sharding/test/copy_filter_test.go
```diff
@@ -116,6 +131,111 @@ func (t *CopyFilterTestSuite) TestFallsBackToIgnoredPrimaryIndex() {
...
+	selectBuilder, err := t.filter.BuildSelect([]string{"*"}, t.normalTable, t.paginationKeyCursor, 1024)
+	t.Require().Nil(err)
+
+	sql, args, err := selectBuilder.ToSql()
+	t.Require().Nil(err)
+	t.Require().Equal("SELECT * FROM `shard_1`.`normaltable` JOIN (SELECT `id` FROM `shard_1`.`normaltable` FORCE INDEX (`good_sharding_index`) WHERE `tenant_id` = ? AND `id` > ? ORDER BY `id` LIMIT 1024) AS `batch` USING(`id`)", sql)
+	t.Require().Equal([]interface{}{t.shardingValue, t.paginationKeyCursor}, args)
+}
+
+func (t *CopyFilterTestSuite) TestHigherSpecifityOfIndexHintingPerTable() {
```
mtaner, line 156: will fix the typo here before merging: `Specificity`

sharding/config.go
```diff
@@ -45,33 +46,35 @@ func (c *Config) ValidateConfig() error {
...
 
 type ShardedCopyFilterConfig struct {
 	// Ghostferry requires an index to be present on sharding key to improve the performance of the data iterator's query.
 
-	// DisableIndexHint disables the named index hint on the sharding key. Defaults to false.
-	DisableIndexHint bool
-
-	// `USE INDEX` is used by default as part of the index hint if DisableIndexHint is false.
-	// UseForceIndex replaces `USE INDEX` with `FORCE INDEX` on the query. Defaults to false.
+	// `USE` is used by default as part of the index hint i.e. `USE INDEX`.
```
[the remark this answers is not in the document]
mtaner: Yeah but I will also transform it to lower case just incase

coding-chimp (approved):

sharding/filter.go
```diff
@@ -61,8 +69,23 @@ func (f *ShardedCopyFilter) BuildSelect(columns []string, table *ghostferry.Tabl
...
+		selectPaginationKeys := "SELECT " + quotedPaginationKey + " FROM " + quotedTable
+
+		if indexHint != "none" {
+			selectPaginationKeys += " " + f.shardingKeyIndexHint(table, indexHint)
+		}
+
+		selectPaginationKeys += " WHERE " + quotedShardingKey + " = ? AND " + quotedPaginationKey + " > ?" +
 			" ORDER BY " + quotedPaginationKey + " LIMIT " + strconv.Itoa(int(batchSize))
 
 		return sq.Select(columns...).
```
ilikeorangutans, line 91: Happily surprised to see Squirrel here.

sharding/filter.go
mtaner: already doing here: https://github.com/Shopify/ghostferry/pull/380/files#diff-98e32b41746dfde3eae5ea7e3ececaa8db1cfcbcc4889e9d4634ebb0c7aa2782R100

PR #386, opened 2025-04-03
[Do not merge] Replace cursor RowLock with FOR SHARE NOWAIT 

grodowski:
⚠️


testhelpers/data_writer.go
```diff
@@ -175,6 +176,8 @@ func (this *MixedActionDataWriter) WriteData(i int) {
 		if err != nil {
 			panic(err)
 		}
+
+		time.Sleep(10 * time.Millisecond)
```
grodowski, line 180: Added this to make the `DataWriter` more realistic, some integration tests ([example](https://github.com/Shopify/ghostferry/blob/e8ffb0f6e792b9674070b0eaddb1bc086dc004f2/test/go/iterative_verifier_integration_test.go#L56-L62)) start it with `NumberOfWriters: 4` and that's too much for `NOWAIT` to work without sometimes taking way too long to obtain a lock.

I'm wondering if this is a signal to consider just a `FOR SHARE` lock 🤔

config.go
```diff
@@ -856,7 +857,7 @@ func (c *Config) ValidateConfig() error {
 	}
 
 	if c.DBReadRetries == 0 {
-		c.DBReadRetries = 5
+		c.DBReadRetries = 60
```
grodowski, line 860: Implies `60 * 1s` retry schedule. I thought it's a reasonable starting point, similarly to how `gh-ost` does it ([60](https://github.com/github/gh-ost/blob/801ebabe343621fc06c92667fdb404f24bc605c0/go/base/context.go#L288) and [1s](https://github.com/github/gh-ost/blob/801ebabe343621fc06c92667fdb404f24bc605c0/go/logic/migrator.go#L140))

testhelpers/data_writer.go
coding-chimp: `FOR SHARE` is equivalent to what we're doing right now, so it's not a step down, right?

I mean, we haven't seen problems with ghostferry holding on to locks so far, AFAIK, so it wouldn't be the worst thing to keep doing it. However, if the goal is the minimize production impact than not holding on to locks would be better.

grodowski: Yep, I agree that `FOR SHARE` is a safer option, but also don't have new ideas on how to de-risk `FOR SHARE NOWAIT` further. I think it's safe enough in a way that if the cursor fails to obtain a lock (which I doubt, the logic is equal to `gh-ost`'s iterator now, which isn't even configurable), we can guarantee that any data will remain unaffected and ghostferry may just fail to run. 

Do you have any objections against shipping with `NOWAIT` (or how we could test it further while shipping `FOR SHARE` first)?

coding-chimp: I think it's fine to go with `NOWAIT`. 👍

PR #430, opened 2026-04-14
Get rid of vendored packages

driv3r:
this should streamline updates and reduce the noise in PRs


austenLacy (commented):
do we not need to update anything in go.mod?

driv3r:
@austenLacy no, it gets picked up by `vendor/modules.txt` only. There was some cache issue which caused builds in ruby tests to fail, but this is fixed now.

austenLacy (approved):

PR #431, opened 2026-04-15
fix: initialise BinlogWriter channel before Run() to prevent nil-channel hang

driv3r:
`Ferry.Run()` launches `BinlogWriter.Run()` and `BinlogStreamer.Run()` in concurrent goroutines (ferry.go:733-745). Previously, the event buffer channel was created inside `BinlogWriter.Run()`, so if the `BinlogStreamer` received a row event and called `BufferBinlogEvents()` before the writer goroutine was scheduled, the send would block forever on a `nil` channel, hanging the test (observed as T`estUpdateBinlogSelectCopy` timing out after 10 minutes with goroutine state "chan send (nil chan)").

Introduce `BinlogWriter.Initialize()` — mirroring the existing pattern on BatchWriter — that allocates the channel and logger eagerly. Call it from `Ferry.NewBinlogWriter()` so the writer is ready for buffering before any goroutine starts. Add a nil-channel guard in `Run()` as a defensive fallback for callers that construct `BinlogWriter` directly.

Add `TestBinlogWriterBufferBinlogEventsBeforeRun` to give deterministic, sub-second coverage of the race without relying on the slow integration test timeout.

### DIFF
--- a/binlog_writer.go
+++ b/binlog_writer.go
@@ -37,8 +37,7 @@ func (b *BinlogWriter) Initialize() {
 
 func (b *BinlogWriter) Run() {
 	if b.binlogEventBuffer == nil {
-		// Defensive fallback: caller forgot Initialize(), behave as before.
-		b.Initialize()
+		panic("Initialize() has not been called prior to Run()")
 	}
 
 	batch := make([]DMLEvent, 0, b.BatchSize)
--- a/binlog_writer.go
+++ b/binlog_writer.go
@@ -25,9 +25,21 @@ type BinlogWriter struct {
 	logger                 Logger
 }
 
-func (b *BinlogWriter) Run() {
+// Initialize allocates the event buffer channel and logger so that
+// BufferBinlogEvents can be called safely before Run() starts in its own
+// goroutine.  Ferry.NewBinlogWriter calls this automatically; external callers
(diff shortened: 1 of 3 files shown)


binlog_writer.go
```diff
@@ -25,9 +25,21 @@ type BinlogWriter struct {
...
+// registering BufferBinlogEvents as an event listener.
+func (b *BinlogWriter) Initialize() {
 	b.logger = LogWithField("tag", "binlog_writer")
 	b.binlogEventBuffer = make(chan DMLEvent, b.BatchSize)
+}
+
+func (b *BinlogWriter) Run() {
+	if b.binlogEventBuffer == nil {
+		// Defensive fallback: caller forgot Initialize(), behave as before.
+		b.Initialize()
```
austenLacy: do you think this is better than a hard failure? seems like it leaves opens the chance of the race-y behavior.

austenLacy (approved):

binlog_writer.go
driv3r: you are correct, I would probably change this to panic instead in case someone generates their own commands from the components, wdyt?

austenLacy: `panic` seems fine. Not sure if we rely on that elsewhere in ghostferry, but I think in the run entrypoint it makes sense.

PR #432, opened 2026-04-15
fix: relax exact-batch-count assertions in interrupt/resume tests

driv3r:
TERM is delivered asynchronously via Go's signal channel. The signal handler calls ErrorHandler.Fatal which panics, but there is a race window between send_signal("TERM") returning in the Ruby callback and the signal actually interrupting the data iterator goroutine. A second 200-row batch can complete before the panic propagates, causing the target row count to be 400 instead of the asserted 200.

The real correctness invariant is already checked by the subsequent assertions: last_successful_id from the target matches the pagination key in the dumped state. The exact count check adds no additional safety and only makes the tests fragile. Relax both the integer-keyed and UUID-keyed variants to assert_operator :>= 200.


austenLacy (approved):
Maybe good to add to the comment why the subsequent ID assertions make a >200 count OK.

PR #433, opened 2026-04-15
Update GitHub actions steps

driv3r:


.github/workflows/release.yml
```diff
@@ -14,18 +14,18 @@ jobs:
...
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@v6.0.2
 
       - name: Setup Go
-        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        uses: actions/setup-go@v6.4.0
         with:
-          go-version: 1.24.1
+          go-version: "1.26.2"
```
mtaner, line 22: just confirming that this doesn't make a difference with quotes or not?

mtaner (approved):
👍  have one unblocking comment

.github/workflows/release.yml
driv3r: it's safer with quotes, I don't trust yaml xD

PR #437, opened 2026-04-16
Feature: Handle read only columns

### LINKED ISSUE #338: Trouble with virtual generated columns
👋  We have a virtual generated column, but it seems like ghostferry is trying to insert data into the virtual generated column when it should not, resulting in an error like...
```
during prepare query near paginationKey <redacted>: Error 3105: The value specified for generated column '<redacted>' in table '<redacted>' is not allowed.
```
Seems like we should just not insert any data into virtual generated columns

driv3r:
ref: #400 & @proton-lisandro-pin 

The back and forth is taking a bit of time, lets speed this up, I've cherry-picked your commits so the attribution is there, CLA was signed on original PR, if you could sign the one here as well it would be 👍 

## Handle MySQL Generated Columns (STORED and VIRTUAL) in Data Replication

This PR adds support for MySQL generated columns (both `VIRTUAL` and `STORED`) to Ghostferry, enabling proper handling of computed columns during selective data replication.

### Problem Statement

MySQL 8.0.23 introduced significant changes to how generated columns are handled in binary log ROW events:
- **Virtual columns** are completely omitted from binlog events (not stored on disk)
- **Stored columns** are included in binlog events (computed once and persisted)

Without special handling, Ghostferry would fail or produce incorrect results when replicating tables with generated columns, as it would attempt to insert values into columns that cannot be modified or would have incorrect column positions.

### Solution

This PR implements a comprehensive solution with four key components:

1. **Row Expansion** - Detects when MySQL omits virtual columns from binlog events and re-inserts nil sentinels to maintain consistent full-schema column indexing throughout the pipeline.

2. **Insert Value Filtering** - Filters out generated column values before constructing `INSERT` statements, allowing only modifiable columns to be inserted while using proper column metadata for value escaping.

3. **Unsigned Integer Normalization** - Fixed the order of operations: row expansion happens before unsigned integer normalization, ensuring consistent full-schema column indexing throughout.

4. **Verification with Generated Columns** - Includes all columns (including generated) in fingerprint queries to detect divergence when computed values differ between source and target databases.

#### Core Implementation
- **dml_events.go**: Modified `INSERT` event handling to filter out generated columns and improved binlog event processing for MySQL 8.0.23+ compatibility
- **table_schema_cache.go**: Added column classification and filtering utilities:
  - `IsColumnGenerated()` - Identifies virtual and stored columns
  - `NonGeneratedColumnNames()` - Returns only insertable columns
  - `FilterGeneratedColumnsOnRowData()` - Removes generated column values from rows
- **row_batch.go**: Updated row batch handling for filtered column data
- **iterative_verifier.go**: Updated verification logic to handle generated columns

#### Test Coverage
- Added unit tests for generated column handling with edge cases:
  - Virtual columns before unsigned integer columns
  - Generated columns before JSON columns (ensures JSON casting is preserved)
  - Mixed `VIRTUAL` and `STORED` columns
- Added integration tests confirming:
  - Verification detects divergence in computed generated column values
  - Stored and virtual generated columns are handled correctly
  - Interrupt/resume scenarios work with generated columns

### Edge Cases Handled
- ✅ Virtual columns before unsigned integer columns
- ✅ Generated columns before JSON columns (JSON casting preserved)
- ✅ Mixed `VIRTUAL` and `STORED` columns in same table
- ✅ MySQL version differences (pre-8.0.23 vs 8.0.23+)
- ✅ Interrupt/resume with generated columns
- ✅ Verification with computed column divergence detection

### Testing
- **5 new commits** with comprehensive testing
- Tests for critical edge cases involving generated columns and other column types
- Integration tests verifying both inline and checkpoint verification modes
- Fixed race condition in interrupt/resume tests

### Related Issue
Closes #338

### DIFF
--- a/dml_events.go
+++ b/dml_events.go
@@ -42,14 +42,14 @@ type RowData []interface{}
 // https://github.com/Shopify/ghostferry/issues/165.
 //
 // In summary:
-// - This code receives values from both go-sql-driver/mysql and
-//   go-mysql-org/go-mysql.
-// - go-sql-driver/mysql gives us int64 for signed integer, and uint64 in a byte
-//   slice for unsigned integer.
-// - go-mysql-org/go-mysql gives us int64 for signed integer, and uint64 for
-//   unsigned integer.
-// - We currently make this function deal with both cases. In the future we can
-//   investigate alternative solutions.
+//   - This code receives values from both go-sql-driver/mysql and
+//     go-mysql-org/go-mysql.
+//   - go-sql-driver/mysql gives us int64 for signed integer, and uint64 in a byte
+//     slice for unsigned integer.
+//   - go-mysql-org/go-mysql gives us int64 for signed integer, and uint64 for
+//     unsigned integer.
+//   - We currently make this function deal with both cases. In the future we can
+//     investigate alternative solutions.
 func (r RowData) GetUint64(colIdx int) (uint64, error) {
 	u64, ok := Uint64Value(r[colIdx])
 	if ok {
@@ -174,8 +174,8 @@ func (e *BinlogInsertEvent) AsSQLString(schemaName, tableName string) (string, e
 
 	query := "INSERT IGNORE INTO " +
 		QuotedTableNameFromString(schemaName, tableName) +
-		" (" + strings.Join(quotedColumnNames(e.table), ",") + ")" +
-		" VALUES (" + buildStringListForValues(e.table.Columns, e.newValues) + ")"
+		" (" + strings.Join(quotedColumnNamesForInsert(e.table), ",") + ")" +
+		" VALUES (" + buildStringListForInsertValues(e.table, e.newValues) + ")"
 
 	return query, nil
 }
@@ -323,10 +323,14 @@ func NewBinlogDMLEvents(table *TableSchema, ev *replication.BinlogEvent, pos, re
 	}
 }
 
-func quotedColumnNames(table *TableSchema) []string {
-	cols := make([]string, len(table.Columns))
-	for i, column := range table.Columns {
-		cols[i] = QuoteField(column.Name)
+func quotedColumnNamesForInsert(table *TableSchema) []string {
+	cols := []string{}
+
+	for _, c := range table.Columns {
+		if c.IsVirtual {
+			continue
+		}
+		cols = append(cols, QuoteField(c.Name))
 	}
 
 	return cols
@@ -347,15 +351,18 @@ func verifyValuesHasTheSameLengthAsColumns(table *TableSchema, values ...RowData
 	return nil
 }
 
-func buildStringListForValues(columns []schema.TableColumn, values []interface{}) string {
+func buildStringListForInsertValues(table *TableSchema, values []interface{}) string {
 	var buffer []byte
 
 	for i, value := range values {
-		if i > 0 {
-			buffer = append(buffer, ',')
+		if table.Columns[i].IsVirtual {
+			continue
 		}
 
-		buffer = appendEscapedValue(buffer, value, columns[i])
+		if len(buffer) != 0 {
+			buffer = append(buffer, ',')
+		}
+		buffer = appendEscapedValue(buffer, value, table.Columns[i])
 	}
 
 	return string(buffer)
@@ -504,10 +511,10 @@ func Int64Value(value interface{}) (int64, bool) {
 //
 // This is specifically mentioned in the the below link:
 //
-//    When BINARY values are stored, they are right-padded with the pad value
-//    to the specified length. The pad value is 0x00 (the zero byte). Values
-//    are right-padded with 0x00 for inserts, and no trailing bytes are removed
-//    for retrievals.
+//	When BINARY values are stored, they are right-padded with the pad value
+//	to the specified length. The pad value is 0x00 (the zero byte). Values
+//	are right-padded with 0x00 for inserts, and no trailing bytes are removed
+//	for retrievals.
 //
 // ref: https://dev.mysql.com/doc/refman/5.7/en/binary-varbinary.html
 func appendEscapedString(buffer []byte, value string, rightPadToLengthWithZeroBytes int) []byte {
--- a/dml_events.go
+++ b/dml_events.go
@@ -282,72 +282,39 @@ func (e *BinlogDeleteEvent) PaginationKey() (string, error) {
 func NewBinlogDMLEvents(table *TableSchema, ev *replication.BinlogEvent, pos, resumablePos mysql.Position, query []byte) ([]DMLEvent, error) {
 	rowsEvent := ev.Event.(*replication.RowsEvent)
 
-	// Count how many columns MySQL actually emits in binlog row images.
-	// In MySQL 8.0.23+, VIRTUAL generated columns are not stored anywhere and
-	// are therefore absent from binlog ROW events.  STORED generated columns
-	// and all real columns are always present.
-	fullCount := len(table.Columns)
-	nonVirtualCount := 0
-	for _, col := range table.Columns {
-		if !col.IsVirtual {
-			nonVirtualCount++
-		}
-	}
-
 	for i, rawRow := range rowsEvent.Rows {
-		switch len(rawRow) {
-		case fullCount:
-			// All columns present (older MySQL or tables without virtual cols).
-			// Nothing to expand.
-
-		case nonVirtualCount:
-			// Virtual generated columns absent from the binlog row image
-			// (MySQL 8.0.23+ behaviour).  Re-insert nil sentinels at each
-			// virtual column's schema position so that all downstream code
-			// can use unmodified full-schema indexes throughout.
-			expanded := make([]interface{}, fullCount)
-			rowIdx := 0
-			for colIdx, col := range table.Columns {
-				if col.IsVirtual {
-					expanded[colIdx] = nil
-				} else {
-					expanded[colIdx] = rawRow[rowIdx]
-					rowIdx++
-				}
-			}
-			rowsEvent.Rows[i] = expanded
-
-		default:
+		if len(rawRow) != len(table.Columns) {
 			return nil, fmt.Errorf(
-				"table %s.%s has %d columns (%d non-virtual) but event has %d columns instead",
+				"table %s.%s has %d columns but event has %d columns instead",
 				table.Schema,
 				table.Name,
-				fullCount,
-				nonVirtualCount,
+				len(table.Columns),
 				len(rawRow),
 			)
 		}
 
-		// Normalize signed-to-unsigned integer values in place on the
-		// (possibly expanded) raw row using full-schema column indexes.
-		// Virtual column sentinels (nil) will not match any integer type.
-		row := rowsEvent.Rows[i]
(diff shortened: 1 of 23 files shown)


driv3r:
Hey @plisandro ! There was a bunch more things to tackle here, everything should be in place now, feel free to review and test, as well as sign the CLA mentioned in the checks for the contributions 👍

proton-lisandro-pin:
> Hey @plisandro ! There was a bunch more things to tackle here, everything should be in place now, feel free to review and test, as well as sign the CLA mentioned in the checks for the contributions 👍

This is much appreciated, thank you! Been testing it today and your PR seems to work well. CLA is now signed as well :smile:

driv3r:
Hey @plisandro I'm not super familiar with the CLA stuff, but the error says:
> - @plisandro: [Sign the CLA](https://cla.shopify.com) and comment "I have signed the CLA!" to re-run the checks and have your PR reviewed. 

leave a comment and lets see, also I think you may have committed under your original account, so you might need to leave/sign under it as well 🤔

plisandro:
I have signed the CLA!

plisandro:
> leave a comment and lets see, also I think you may have committed under your original account, so you might need to leave/sign under it as well 🤔

Done, and apologies for the confusion - i'm in the process of merging accounts now, and this was actually the last contribution left with the old one :facepalm:

driv3r:
@plisandro no problem!

table_schema_cache.go
```diff
@@ -45,6 +45,89 @@ type TableSchema struct {
...
+}
+
+// IsColumnIndexGenerated evaluates whether a TableSchema column is generated, by index.
+func (t *TableSchema) IsColumnIndexGenerated(idx int) bool {
+	return IsColumnGenerated(&t.Columns[idx])
+}
+
+// Evaluates whether a TableSchema column is generated, by name.
+func (t *TableSchema) IsColumnNameGenerated(name string) bool {
+	for _, col := range t.Columns {
```
milanatshopify, line 60: Are we worried about iterating through columns every time? I know there can't be that many, so it's probably not an issue, but we do it in a few places in this file.

plisandro: I tested this locally, and has no discernible runtime impact when compared to a (previous) version which filters columns beforehand.

As noted in https://github.com/Shopify/ghostferry/pull/400#issuecomment-4156633154 , the problem with doing any pre-processing on columns is that they won't later align with DML and Binlog updates, so the logic to handle these becomes a mess.

milanatshopify: Sure, but is it worth it. See comment on row_batch.go

row_batch.go
```diff
@@ -64,23 +74,42 @@ func (e *RowBatch) AsSQLQuery(schemaName, tableName string) (string, []interface
...
 func (e *RowBatch) flattenRowData() []interface{} {
-	rowSize := len(e.values[0])
-	flattened := make([]interface{}, rowSize*len(e.values))
-
-	for rowIdx, row := range e.values {
-		for colIdx, col := range row {
-			flattened[rowIdx*rowSize+colIdx] = col
+	flattened := make([]interface{}, 0, len(e.values)*len(e.columns))
+
+	for _, row := range e.values {
```
milanatshopify, line 117: This is now going through all rows, all columns, then all columns again. #rows * #cols^2. Or am I reading it wrong?

milanatshopify: Asking because this will add up when you have 86 columns in a large table, and we're doing string comparisons. no?

table_schema_cache.go
plisandro: Note that "a mess" includes being a tad slower :smile: To process row updates with generated columns  missing on table definitions means that you effectively need to keep track of those columns, and their indeces, to filter them out _on every update_ - which also involves re-aligning row data.

Filtering generated columns when SQL is constructed is actually simpler/cheaper. Either way, i tested this in a setup with tables with 50+ columns and couldn't measure any significant difference.

row_batch.go
plisandro: The original change followed `TableSchema` as it exists today, where ignored and compressed columns are indexed by name.

Didn't seem to have any relevant performance hit on my test setup, but if this is a consideration, we could use a hashmap instead. Note it'd have to be lazily-initialized, as `TableSchema` is initialized ad-hoc all over ghostferry.

test/integration/iterative_verifier_test.rb
```diff
@@ -23,6 +23,55 @@ def test_iterative_verifier_succeeds_in_normal_run
...
+    datawriter = new_source_datawriter
+    ghostferry = new_ghostferry(MINIMAL_GHOSTFERRY, config: { verifier_type: "Iterative" })
+
+    start_datawriter_with_ghostferry(datawriter, ghostferry)
+    stop_datawriter_during_cutover(datawriter, ghostferry)
+
+    verification_ran = false
+    ghostferry.on_status(Ghostferry::Status::VERIFIED) do |*incorrect_tables|
+      verification_ran = true
+      assert_equal ["gftest.test_table_1"], incorrect_tables
```
grodowski, line 45: Checked whether we can assert the failures here more precisely, but seems like we'd need to add additional output to the verifier. Leaving as-is for now. 👀

row_batch.go
grodowski: [bbd5c31](https://github.com/Shopify/ghostferry/pull/437/commits/bbd5c31a09982f2392b6eb4fce94565987e1e448) adds a new field to cache non-generated columns within a `RowBatch`, which saves a bunch of string comparisons and doesn't escape the struct implementation.

grodowski (approved):
Ran manual tests against our internal tooling with both virtual and stored columns and found no issues on the copy path. We have to trust the integration/unit tests for streaming though, as I wasn't able to exercise that path reliably with test shards are set up on a single db (and binlog). 

Added a few more integration tests in [2717bfd](https://github.com/Shopify/ghostferry/pull/437/commits/2717bfd0a734c594e070617d1da516560026629d) to cover for this.

grodowski:
Closes #338

mtaner (approved):

dml_events.go
```diff
@@ -347,53 +355,71 @@ func verifyValuesHasTheSameLengthAsColumns(table *TableSchema, values ...RowData
...
 	return string(buffer)
 }
 
-func buildStringMapForWhere(columns []schema.TableColumn, values []interface{}) string {
+func buildStringMapForWhere(table *TableSchema, values []interface{}) string {
 	var buffer []byte
 
 	for i, value := range values {
-		if i > 0 {
+		if table.IsColumnIndexGenerated(i) {
```
milanatshopify, line 388: We don't put generated columns in the WHERE...

table_schema_cache.go
```diff
@@ -282,6 +386,13 @@ func (t *TableSchema) paginationKeyColumn(cascadingPaginationColumnConfig *Casca
 	}
 
 	if paginationKeyColumn != nil {
+		// VIRTUAL generated columns are not stored on disk and cannot be used for
+		// data iteration.  STORED generated columns are physically stored and are
+		// safe to use.
+		if paginationKeyColumn.IsVirtual {
```
milanatshopify, line 392: We allow STORED generated columns...

milanatshopify (commented):
Looks like we allow STORED generated columns as pagination key, which is good - this will come in handy. But we don't use it in the WHERE clause for binlog updates.

We should still find the right records, but it's going to be slow, potentially, since we're not using the pagination key in the WHERE for the binlog updates - is that ok?

There is also an edge case where it will just fail, I'll attach the agent created test that fails.

```
func (this *TableSchemaCacheTestSuite) TestStoredPaginationKeyPreservedInBinlogUpdateWhereClause() {
	table := "stored_pk_with_text_body"

	// BINARY(32) holds the raw SHA-256 bytes — a common content-addressed-
	// storage pattern.  BINARY is required (rather than CHAR(64) hex) because
	// the pagination-key validator demands a numeric type or a binary-safe
	// string type; BINARY/VARBINARY satisfy that without needing an explicit
	// _bin collation.
	query := fmt.Sprintf(
		"CREATE TABLE %s.%s (doc TEXT NOT NULL, doc_hash BINARY(32) AS (UNHEX(SHA2(doc, 256))) STORED, PRIMARY KEY (doc_hash))",
		testhelpers.TestSchemaName, table,
	)
	_, err := this.Ferry.SourceDB.Exec(query)
	this.Require().Nil(err)

	cascadingPaginationColumnConfig := &ghostferry.CascadingPaginationColumnConfig{
		PerTable: map[string]map[string]string{
			testhelpers.TestSchemaName: {table: "doc_hash"},
		},
	}

	tables, err := ghostferry.LoadTables(this.Ferry.SourceDB, this.tableFilter, nil, nil, nil, cascadingPaginationColumnConfig)
	this.Require().Nil(err)

	tableSchema := tables[fmt.Sprintf("%s.%s", testhelpers.TestSchemaName, table)]
	this.Require().NotNil(tableSchema)
	this.Require().Equal("doc_hash", tableSchema.GetPaginationColumn().Name)

	tableMapEvent := &replication.TableMapEvent{
		Schema: []byte(testhelpers.TestSchemaName),
		Table:  []byte(table),
	}
	eventBase := ghostferry.NewDMLEventBase(tableSchema, mysql.Position{}, mysql.Position{}, nil, time.Unix(1618318965, 0))

	// Raw 32-byte SHA-256 digests of 'hello' and 'world'.  Real binlogs ship
	// the STORED column value in the row image, so we mirror that here.
	hashOld := []byte{
		0x2c, 0xf2, 0x4d, 0xba, 0x5f, 0xb0, 0xa3, 0x0e,
		0x26, 0xe8, 0x3b, 0x2a, 0xc5, 0xb9, 0xe2, 0x9e,
		0x1b, 0x16, 0x1e, 0x5c, 0x1f, 0xa7, 0x42, 0x5e,
		0x73, 0x04, 0x33, 0x62, 0x93, 0x8b, 0x98, 0x24,
	}
	hashNew := []byte{
		0x48, 0x6e, 0xa4, 0x62, 0x24, 0xd1, 0xbb, 0x4f,
		0xb6, 0x80, 0xf3, 0x4f, 0x7c, 0x9a, 0xd9, 0x6a,
		0x8f, 0x24, 0xec, 0x88, 0xbe, 0x73, 0xea, 0x8e,
		0x5a, 0x6c, 0x65, 0x26, 0x0e, 0x9c, 0xb8, 0xa7,
	}

	rowsEvent := &replication.RowsEvent{
		Table: tableMapEvent,
		Rows: [][]interface{}{
			{[]byte("hello"), hashOld},
			{[]byte("world"), hashNew},
		},
	}

	dmlEvents, err := ghostferry.NewBinlogUpdateEvents(eventBase, rowsEvent)
	this.Require().Nil(err)
	this.Require().Equal(1, len(dmlEvents))

	q, err := dmlEvents[0].AsSQLString(testhelpers.TestSchemaName, table)
	this.Require().Nil(err)

	this.Require().Contains(
		q, "`doc_hash`=",
		"STORED pagination key must be preserved in the WHERE clause of binlog UPDATE replay; "+
			"without it Ghostferry falls back to scanning by TEXT body, which forces full-table scans "+
			"and widens the iterative verifier's lock scope. Generated SQL was: "+q,
	)
}

```

PR #445, opened 2026-07-21
GTID 1/5: Introduce BinlogCoordinate abstraction (file/position backed)

driv3r:
## Stacked PR 1 of 5 — GTID support

This is the first PR in a stack that adds opt-in GTID binlog-coordinate support to Ghostferry, keeping file/position as the default and fully backward compatible.

**Stack**
1. **This PR** — BinlogCoordinate abstraction (file/position backed) → base `main`
2. GTID coordinate mode (representation + config + reads) → base this branch
3. GTID binlog streaming start/stop → base no. 2)
4. Persist GTID coordinates in state + enable resume → base no. 3)
5. GTID-aware replica catchup → base  no. 4)

### What this PR does
Introduces a representation-agnostic `BinlogCoordinate` type as the seam for a future GTID mode, without changing current behavior.

- `BinlogCoordinate` wraps `mysql.Position` today (type `file_position`), with `gtid` reserved. Provides `IsZero`/`Compare`/`String` and self-describing JSON that also decodes bare `{Name,Pos}` payloads for backward compatibility.
- `StateTracker`/`SerializableState` gain coordinate accessors that delegate to the existing file/position storage; legacy `mysql.Position` API and serialized fields are unchanged.
- `DMLEvent` gains `BinlogCoordinate()`/`ResumableBinlogCoordinate()`; `BinlogStreamer` gains `GetLastStreamedBinlogCoordinate()`.
- Unit tests for coordinate basics, comparison, and JSON round-trip (including legacy decode paths).

**No behavior change**: everything still runs on file/position. This is the foundation for adding GTID support behind a feature flag.

### DIFF
--- a/binlog_coordinate.go
+++ b/binlog_coordinate.go
@@ -1,0 +1,178 @@
+package ghostferry
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/go-mysql-org/go-mysql/mysql"
+)
+
+// BinlogCoordinateType identifies how a binlog position is expressed.
+//
+// This is the seam that lets Ghostferry evolve from file/position based
+// replication coordinates towards GTID based coordinates without forcing every
+// consumer to know which representation is in use. In this first iteration only
+// the file/position representation is implemented; the GTID representation is
+// reserved so that state serialized today remains forward compatible.
+type BinlogCoordinateType string
+
+const (
+	// BinlogCoordinateFilePosition is the classic (file, position) coordinate.
+	BinlogCoordinateFilePosition BinlogCoordinateType = "file_position"
+
+	// BinlogCoordinateGTID is reserved for the future GTID based coordinate.
+	// It is intentionally defined now so that the serialization format and the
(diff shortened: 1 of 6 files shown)


binlog_coordinate.go
```diff
@@ -0,0 +1,178 @@
...
+	return json.Marshal(out)
+}
+
+// UnmarshalJSON implements json.Unmarshaler.
+//
+// It accepts both the new self-describing form and, for defensive
+// compatibility, a bare mysql.Position object of the form {"Name":...,"Pos":...}.
+func (c *BinlogCoordinate) UnmarshalJSON(data []byte) error {
+	var typed serializedBinlogCoordinate
+	if err := json.Unmarshal(data, &typed); err == nil && typed.Type != "" {
```
mtaner, line 155: So this will still fallback to old format if the data is new format but just invalid, like position is not a number? But we assume it must be old format and try to unmarshal it as old format

austenLacy (approved):

PR #446, opened 2026-07-21
GTID 2/5: Add coordinate mode (representation + config + reads)

driv3r:
## Stacked PR 2 of 5 — GTID support

Base: #445. Review/merge in stack order.

### What this PR does
Extends `BinlogCoordinate` with a GTID representation and adds the `BinlogCoordinateMode` feature flag, **without switching binlog streaming yet**.

- `BinlogCoordinate` gains a GTID form (`NewGTIDCoordinate`, `IsGTID`, `ParsedGTIDSet`, GTID-aware `IsZero`/`String`, JSON). GTID coordinates use set containment (`Contains`); `Compare` panics for GTID since GTID sets have no total order.
- Config gains `BinlogCoordinateMode` (`file_position` default, `gtid` experimental) with validation.
- DB helpers: `ReadExecutedGTIDSet`, `ReadCurrentGTIDCoordinate`, `CheckServerGTIDModeEnabled` (`@@GLOBAL.GTID_EXECUTED` / `gtid_mode`).
- Unit tests for GTID coordinate semantics, JSON round-trip, and mode validation.

Streaming still runs on file/position; this only adds the GTID representation, feature flag, and validation groundwork.

PR #447, opened 2026-07-21
GTID 3/5: Binlog streaming start/stop (behind mode flag)

driv3r:
## Stacked PR 3 of 5 — GTID support

Base: #446. Review/merge in stack order.

### What this PR does
Wires the GTID coordinate mode into `BinlogStreamer` so a fresh run can stream from and stop at a GTID set, while file/position stays the default and unchanged.

- `BinlogStreamer` gains `BinlogCoordinateMode` plus GTID tracking of the streamed, resumable, and stop GTID sets.
- `ConnectBinlogStreamerToMysqlWithCoordinate` / `...FromCoordinate` dispatch to `StartSync` (file/pos) or `StartSyncGTID` (gtid), with coordinate-type validation.
- Tracks committed GTID via `XIDEvent.GSet`; records the pre-transaction set as the resumable point at `GTIDEvent` (clone to avoid aliasing go-mysql's mutable set).
- Loop stop condition replaced with `shouldContinueStreaming` (position compare for file/pos, executed-set containment for GTID). `FlushAndStop` records the GTID stop set.
- Ferry passes the mode to its streamers and uses coordinate-based connect/state updates. GTID resume-from-state is explicitly rejected until GTID state persistence lands (PR 4).
- Internal unit tests for mode dispatch, stop logic, and validation.

File/position behavior is unchanged; GTID streaming works for fresh runs.

### DIFF
--- a/binlog_streamer.go
+++ b/binlog_streamer.go
@@ -45,9 +45,23 @@ type BinlogStreamer struct {
 	DatabaseRewrites map[string]string
 	TableRewrites    map[string]string
 
+	// BinlogCoordinateMode selects whether this streamer tracks file/position
+	// or GTID coordinates. An empty value is treated as file/position for
+	// backwards compatibility.
+	BinlogCoordinateMode BinlogCoordinateType
+
 	lastStreamedBinlogPosition  mysql.Position
 	lastResumableBinlogPosition mysql.Position
 	stopAtBinlogPosition        mysql.Position
+
+	// GTID tracking, only maintained when BinlogCoordinateMode is
+	// BinlogCoordinateGTID. lastStreamedGTIDSet is the committed GTID set seen
+	// so far. lastResumableGTIDSet is the committed GTID set at the last
+	// transaction boundary (a safe resume point). stopAtGTIDSet is the target
+	// executed set to stop at during cutover.
+	lastStreamedGTIDSet  mysql.GTIDSet
+	lastResumableGTIDSet mysql.GTIDSet
+	stopAtGTIDSet        mysql.GTIDSet
 
 	lastProcessedEventTime   time.Time
 	lastLagMetricEmittedTime time.Time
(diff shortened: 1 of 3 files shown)


binlog_streamer.go
```diff
@@ -45,14 +53,37 @@ type BinlogStreamer struct {
...
+	// transaction boundary (a safe resume point). stopAtGTIDSet is the target
+	// executed set to stop at during cutover.
+	//
+	// These are *mysql.MysqlGTIDSet values (Go maps under the hood), mutated on
+	// the streaming goroutine while Ferry.Progress() reads them (via the Get*
+	// coordinate accessors) on another goroutine. Reading a set's String() or
+	// Clone() iterates the map, so concurrent read+write is a data race (and a
+	// fatal "concurrent map read and map write" if a stored set is ever mutated
+	// in place). gtidMu serialises all access to these three fields; every
+	// read and write below goes through the guarded accessors.
```
austenLacy: Unnecessary comments

binlog_streamer.go
```diff
@@ -22,6 +25,11 @@ type BinlogEventState struct {
 	isEventPositionResumable bool
 	isEventPositionValid     bool
 	nextFilename             string
+	inTransaction            bool
+	// A GTID group is open from GTIDEvent through its terminating event,
+	// including when an already committed group is replayed after reconnect.
```
austenLacy: unnecessary comment

binlog_streamer.go
```diff
@@ -45,14 +53,37 @@ type BinlogStreamer struct {
 	DatabaseRewrites map[string]string
 	TableRewrites    map[string]string
 
+	// BinlogCoordinateMode selects whether this streamer tracks file/position
+	// or GTID coordinates. An empty value is treated as file/position for
+	// backwards compatibility.
+	BinlogCoordinateMode BinlogCoordinateType
```
austenLacy: unnecessary comment

binlog_streamer.go
```diff
@@ -148,6 +179,94 @@ func (s *BinlogStreamer) ConnectBinlogStreamerToMysqlFrom(startFromBinlogPositio
...
+			return BinlogCoordinate{}, err
+		}
+		return s.ConnectBinlogStreamerToMysqlSinceCoordinate(coord)
+	default:
+		pos, err := s.ConnectBinlogStreamerToMysql()
+		if err != nil {
+			return BinlogCoordinate{}, err
+		}
+		return NewFilePositionCoordinate(pos), nil
+	}
```
austenLacy: Why wouldn't we call `ConnectBinlogStreamerToMysqlSinceCoordinate` in both cases here? It seems like it handles both type of coordinates and the main difference in this function is where we get the coordinate from (gtid or position).

binlog_streamer.go
```diff
@@ -235,13 +354,54 @@ func (s *BinlogStreamer) defaultEventHandler(ev *replication.BinlogEvent, query
...
+	if !s.stopRequested.Load() {
+		return true
+	}
+
+	// Once stopRequested is set, FlushAndStop has already recorded the stop
+	// coordinate. We must NOT treat a zero/empty stop coordinate as "not yet
+	// recorded": on a fresh source the executed GTID set (or binlog position)
+	// can legitimately be empty, and an empty GTID set is a valid stop target
+	// that any streamed set already contains. Deriving presence from IsZero()
+	// here would hang cutover forever in that case.
```
austenLacy: Nit: very verbose comment just to say "we can't treat a zero coordinate as a stop target because it's a valid use case"

